PR #1, opened 2017-03-22
basic gutter with line numbers

cmyr:
Here's a little first shot at doing line numbers. This is more or less the simplest reasonable implementation I could come up with.

This PR also includes some continued work on moving interaction & general drawing logic from the `EditView` to the `EditViewController`.

Some issues that have occurred to me through this process, which I'll open separately: 

1. This would be a good time to think about support for theming, and the division of responsibilities there. 

2. I'd like to think about horizontal scrolling. When not using word-wrap it feels like horizontal scrolling could be at the discretion of the front-end; the core doesn't need to have any knowledge of the width of the view. In any case, I'd want to think about the relationship between horizontal scrolling and the gutter view. I'm thinking of Sublime's behaviour (the gutter floats in place during hscroll) as a reference.

3. I'd like to think about user preferences; again there might be some division of responsibilities between the frontend and the core here. If line-wrap is a preference then core needs to know about it, but font size maybe not so much.

Anyway that can all get fleshed out more in issues. Let me know if you have any thoughts or concerns about my approach here.

This closes https://github.com/google/xi-editor/issues/175.

### DIFF
--- a/XiEditor/EditViewContainer.swift
+++ b/XiEditor/EditViewContainer.swift
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc. All rights reserved.
+// Copyright 2017 Google Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/XiEditor/EditViewContainer.swift
+++ b/XiEditor/EditViewContainer.swift
@@ -1,0 +1,22 @@
+// Copyright 2016 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Cocoa
+
+class EditViewContainer: NSView {
+
+    override var isFlipped: Bool {
+        return true
+    }
+}
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -14,7 +14,15 @@
 
 import Cocoa
 
-class EditViewController: NSViewController {
+/// The EditViewDataSource protocol describes the properties that an editView uses to determine how to render its contents.
+protocol EditViewDataSource {
+    var lines: LineCache { get }
+    var styleMap: StyleMap { get }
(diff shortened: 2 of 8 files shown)


cmyr:
One note on this: it assigns line numbers regardless of whether or not a break is part of the text or was introduced by the linewrapper. Afaict this information isn't currently made available to the frontend. Let me know if I'm mistaken, otherwise I'll open an issue in xi-core.

XiEditor/EditViewContainer.swift
```diff
@@ -0,0 +1,22 @@
+// Copyright 2016 Google Inc. All rights reserved.
```
raphlinus: It's 2017 now :)

Not urgent, but the general policy is that for new files the copyright should be the current year, but we don't update copyright on existing files.

XiEditor/EditViewController.swift
```diff
@@ -14,32 +14,86 @@
...
+    var gutterWidth: CGFloat {
+        return gutterViewWidth.constant
+    }
+    
+    // used to calculate the gutter width. Initial -1 so that a new document
+    // still triggers update of gutter width.
+    private var lineCount: Int = -1 {
+        didSet {
+            if lineCount != oldValue {
+                let gutterColumns = "\(lineCount)".characters.count
```
raphlinus: So there's a small off-by-1 error here, because the display is 0-based; when the lineCount is 1000, then this gives a width of 4, even though the maximum that will be displayed is 999. However, the maximum line value is another thing that will need to be computed by the core and conveyed to the front-end.

XiEditor/LineCache.swift
```diff
@@ -92,7 +82,7 @@ class LineCache {
         }
         return nil
     }
-
+
```
raphlinus: I try to keep trailing space low, but XCode really seems to want to add it, doesn't it?

XiEditor/EditViewController.swift
cmyr: good catch. My thought would be that display should be 1-based? 👍

cmyr: Also isn't the maximum line value already conveyed to the front-end? we use it to determine the height of the document view.

raphlinus: No, when there's breaking, what's conveyed is the number of visual lines, not the number of lines pre-break. Subtle but important distinction :)

raphlinus:
Thanks! Now I've got some things to do on the core.

XiEditor/LineCache.swift
cmyr: Yea, Xcode seems to have a bias towards ending files with a newline. I'm confused by this diff though, since my local revision has no trailing newline and is up to date. Github has occasional bugs around rendering whitespace in diffs, not sure if this is one of those instances?

XiEditor/EditViewController.swift
cmyr: right I wasn't thinking about the line-breaking case.

PR #5, opened 2017-03-24
select all

cmyr:
depends on https://github.com/google/xi-editor/pull/179. 

Do you have any thoughts on having the core be a git submodule? Doesn't matter hugely right now, but I could imagine versioning being frustrating down the road.

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -179,6 +179,10 @@ class EditViewController: NSViewController, EditViewDataSource {
         document?.sendRpcAsync("insert", params: insertedStringToJson(insertString as! NSString))
     }
 
+    override func selectAll(_ sender: Any?) {
+        document.sendRpcAsync("select_all", params: [])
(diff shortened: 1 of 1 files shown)


raphlinus:
git submodule is a possibility. We don't use that at all inside google, but it might be the best solution to versioning.

Another thing I've thought of is having (a copy of) the xi-core binary in this repo, pulling in a versioned crate of xi-core-lib (and probably with a Cargo.lock to really nail it down). That, I think, would improve on the usability from git submodules.

cmyr:
Hmm. I'm not in thrall to the idea of regularly committing a binary. Or do you just mean a simple crate with a `main.rs` which imports the core-lib?

Another option that saves having to commit anything but a version number or a commit hash would be shelling out to `cargo install` as a build script? This can point to either crates.io or to a repo.

raphlinus:
I meant a simple crate which imports the core-lib, not the binary itself.

PR #6, opened 2017-03-24
Stop treating document as an optional

cmyr:
This is something that had bit me last week when I was playing around with some other drawing changes. By treating the document as an optional we were failing silently if we tried to send an RPC while it was `nil`. I think it's better if we crash in that situation.


cmyr:
I also added a little commit renaming the EditViewContainer to EditContainerView; the earlier name's similarity to EditViewController made using Xcode's quick-open panel frustrating.

raphlinus:
Last I played around with this, it was making the sending of scroll not update its "last" if document was nil. It would only send new scroll parameters if the were different than the "last" ones, so when document was nil it would drop it on the floor. So my only concern is, could there be a race condition where document is nil? If there's a good reason why not, I'm happy to merge this.

PR #8, opened 2017-03-24
fix drawing flicker in newly loaded documents

cmyr:
This fixes an issue where drawing would flicker in a freshly loaded window, because the frontend was not correctly reporting its size to the core.

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -78,9 +78,14 @@ class EditViewController: NSViewController, EditViewDataSource {
         editView.dataSource = self
         gutterView.dataSource = self
         scrollView.contentView.documentCursor = NSCursor.iBeam();
-        
+    }
+
+    override func viewDidAppear() {
+        super.viewDidAppear()
         NotificationCenter.default.addObserver(self, selector: #selector(EditViewController.boundsDidChangeNotification(_:)), name: NSNotification.Name.NSViewBoundsDidChange, object: scrollView.contentView)
         NotificationCenter.default.addObserver(self, selector: #selector(EditViewController.frameDidChangeNotification(_:)), name: NSNotification.Name.NSViewFrameDidChange, object: scrollView)
+        // call to set initial scroll position once we know view size
+        updateEditViewScroll()
     }
 
     // this gets called when the user changes the font with the font book, for example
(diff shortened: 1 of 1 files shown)


XiEditor/EditViewController.swift
```diff
@@ -107,7 +112,7 @@ class EditViewController: NSViewController, EditViewDataSource {
         let first = Int(floor(scrollView.contentView.bounds.origin.y / textMetrics.linespace))
         let height = Int(ceil((scrollView.contentView.bounds.size.height) / textMetrics.linespace))
         let last = first + height
-        if first != firstLine || last != lastLine && document != nil {
```
raphlinus: This change was done for a reason (firstLine, lastLine was getting set to the window size, without the update being sent to the core, because document was nil). The change would only be correct if document was guaranteed not to be nil when this is called, which I don't have a lot of confidence about. But if we're sure it's always set, I'm ok with it.

XiEditor/EditViewController.swift
```diff
@@ -78,9 +78,14 @@ class EditViewController: NSViewController, EditViewDataSource {
...
         scrollView.contentView.documentCursor = NSCursor.iBeam();
-        
+    }
+
+    override func viewDidAppear() {
+        super.viewDidAppear()
         NotificationCenter.default.addObserver(self, selector: #selector(EditViewController.boundsDidChangeNotification(_:)), name: NSNotification.Name.NSViewBoundsDidChange, object: scrollView.contentView)
         NotificationCenter.default.addObserver(self, selector: #selector(EditViewController.frameDidChangeNotification(_:)), name: NSNotification.Name.NSViewFrameDidChange, object: scrollView)
+        // call to set initial scroll position once we know view size
+        updateEditViewScroll()
```
raphlinus: I'm ok with this change, but in my testing we _were_ getting the notification from Cocoa, it's just that it wouldn't reliably be conveyed to the core because the other initialization might not have completed.

cmyr: by moving this to `viewDidAppear` the update fires, because the view's bounds have been set. There were two previous problems: the first was that sometimes the document wasn't set, but the other one was that the notification was sent with the height of the view as set in interface builder (274pt or something) and not the height of the correctly resized view.

I played around with this change and it was working consistently. If it turns out there _is_ a race condition around the setting of the document (and I don't think there should be, the call order of lifecycle methods should be deterministic?) than we will see crashes and I'll try a different approach.

XiEditor/EditViewController.swift
cmyr: Yea, my reasoning here is just to avoid silent failures. If this is getting called too early then we're doing setup at the wrong point in the view controller lifecycle, and that's the thing to change. There is no reason for us to ever send an event before the document is set.

raphlinus: Ok, then I'll merge these and if there are problems, we'll go from there. Thanks!

cmyr: Great. If It turns out I'm wrong I will happily work on some alternative solution. 👍

PR #10, opened 2017-04-13
Add menu item to go to a specified line

cbrewster:
Relies on google/xi-editor#192


raphlinus:
Looks good. I think we can move this to Edit (or have a separate Goto menu like Sublime, since there are a lot of things to go to) fairly soon, as soon as we figure out whether the modal dialog or mini-pane or whatever. But it plays pretty nice as is. Thanks!

PR #11, opened 2017-04-14
Update gutter width when font is changed

cbrewster:


cmyr:
@cbrewster there's a hacky two-liner if you want, which is changing `EditViewController.lineCount` to some other value and back again. 🤷‍♂️

cmyr:
the better solution would be to move the logic that updates that width into a separate `calculateGutterWidth` function, which you can call both from the setter and from `changeFont`.

raphlinus:
Thanks!

PR #16, opened 2017-04-16
Intercept option-arrow events

cmyr:
This fixes the issue with sending duplicate events, and introduces the _new_ issue that Xi's behaviour on macOS is no longer in line with system idioms, e.g. we don't move down by a new paragraph when we send `moveToEndOfParagraph:` and the cursor is currently at the end of a paragraph.

This maybe asks questions about prioritizing consistency of internal editor behaviour versus respect for platform idioms? I'm probably overthinking this.

I probably wouldn't merge this as-is. If we want to go this route, we should update xi-core to recreate the previous behaviour. Might be a fun one for @cbrewster?

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -141,8 +141,24 @@ class EditViewController: NSViewController, EditViewDataSource {
     }
     
     // MARK: - System Events
+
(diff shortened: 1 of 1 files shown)


cmyr:
Closing for now, requires more thought.

Issue #18, opened 2017-04-18
No IME support？


jinyuanxie:
I use XiEditor along with chinese input method, but the window of input method didn't follow the cursor.
![2017-04-18_12-38-10](https://cloud.githubusercontent.com/assets/15538025/25114942/1fb249ca-2435-11e7-81a5-ab1bd951b13a.gif)

pocket7878:
https://github.com/google/xi-editor/pull/70

pocket7878:
Hmm...
Maybe some changes affects to `func firstRect(forCharacterRange aRange: NSRange, actualRange: NSRangePointer?) -> NSRect`'s implementation

jinyuanxie:
@pocket7878 You are right, it always report like that `XiEditor[8625:49370] (0.0, 0.0, 0.0, 0.0)`, i'm trying to fix this issue.

pocket7878:
@jinyuanxie Thank you!

pocket7878:
It seems `cursorPos` didn't updated in draw method

jinyuanxie:
@pocket7878  Thank you! I will create a PR later, but there are another problem.
How should we handle the situation that have multi-cursor ?

pocket7878:
Maybe we should define things like a "active cursor".

jinyuanxie:
Maybe we should wait backend for supporting "Multiple Cursors" https://github.com/google/xi-editor/issues/188

cmyr:
this will definitely be revisited when multi-cursor lands, but I'm not sure how IME would even work if you're inserting text in different contexts simultaneously. If you have any ideas I'm very curious! :)

tsekityam:
I would like to further improve this behaviour. 

The input box should not always follows the cursor. The box should appears at the position of active cursor, then the box's position should be fixed until the input is completed.

That's how the box works in Xcode (and most other editor, if not all), 
![xcode](https://cloud.githubusercontent.com/assets/3814422/25173043/6082928e-2525-11e7-8409-2bb24642ce99.gif)

However, in Xi, the position of the box follows the cursor from time to time,
![xi](https://cloud.githubusercontent.com/assets/3814422/25173046/63bf37a4-2525-11e7-97a1-4dd3e31e8320.gif)

Current behaviour is a bit annoying because It makes me harder to search for a word in the input box.

I will make a PR later tonight.

jinyuanxie:
[...]

PR #19, opened 2017-04-18
Fixed IME support issue

jinyuanxie:
Related issue: #18 

Fixed by updating `cursorPos` in `draw` method.

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -204,6 +204,8 @@ class EditView: NSView, NSTextInputClient {
             dataSource.styleMap.applyStyles(text: s, string: &attrString, styles: line.styles)
             for c in line.cursor {
                 let cix = utf8_offset_to_utf16(s, c)
+                // TODO: How should we handle the situations that have multi-cursor?
+                self.cursorPos = (lineIx, cix)
                 if (markedRange().location != NSNotFound) {
                     let markRangeStart = cix - markedRange().length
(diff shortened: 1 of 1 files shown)


jinyuanxie:
I signed it!

XiEditor/EditView.swift
```diff
@@ -248,6 +248,15 @@ class EditView: NSView, NSTextInputClient {
                 }
             }
         }
+
+        // Update cursorPos
+        // TODO: How should we handle the situations that have multi-cursor?
+        for lineIx in first..<last {
+            guard let line = getLine(lineIx) else { continue }
```
cmyr: Thanks for this catch!

Two little things: instead of running the loop again, you can set `cursorPos` when we draw the cursor(s), around line 205 in master. That will also let you use the utf-16 offset that is calculated there (the `cix` variable), to make sure the IME window lines up correctly with the text. (Try this out and see if it isn't more correct, the window should line up better with the input.)

jinyuanxie: Hmm... I have updated the PR.

raphlinus:
Thanks for the fix! Really gratifying to see more attention to i18n issues in the project.

Multi-cursor is going to be interesting. It will require sending IME state (specifically the composing region) back and forth over the front-end <-> core channel. Fortunately, this is something I'm motivated to work on soon, as we're building IME services in Fuchsia and I'll want to bind with that.

PR #24, opened 2017-04-21
Show unsaved changed in the UI

cbrewster:


raphlinus:
Thanks! This really adds a tremendous amount of polish, and reduces the anxiety that you might accidentally lose work.

PR #26, opened 2017-04-22
Ensure that Document.close() is only called once

### LINKED ISSUE #25: Double call of close()
Repro steps: open new editor, type "abc", ⌘-Q to quit, say "don't save".

```
thread 'main' panicked at 'missing buffer id when closing view', /Users/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-mac/build/src/libcore/option.rs:715
```

Doing a little debugging, it looks like the close() method in Document.swift is being called twice. I don't know whether this is expected behavior (in which it should null out some of its own state on the first call and skip the second), or whether the NSDocument lifecycle is doing something dodgy to make this double call happen.

Should be an easy issue for someone who understands Cocoa.

cmyr:
After sleeping on it I think this is the cleanest approach. I'm thinking about how multi-view is going to work, but I don't think that's really relevant here.

closes #25

### DIFF
--- a/XiEditor/Document.swift
+++ b/XiEditor/Document.swift
@@ -150,13 +150,12 @@ class Document: NSDocument {
         //TODO: save operations should report success, and we should pass any errors to the completion handler
         completionHandler(nil)
     }
-    
+
     // Document.close() can be called multiple times (on window close and application terminate)
-    fileprivate var _isClosed = false
     override func close() {
-        if !self._isClosed {
-            self._isClosed = true
-            Events.CloseView(viewIdentifier: coreViewIdentifier!).dispatch(dispatcher!)
+        if let identifier = self.coreViewIdentifier {
+            self.coreViewIdentifier = nil
+            Events.CloseView(viewIdentifier: identifier).dispatch(dispatcher!)
             super.close()
         }
     }
--- a/XiEditor/Document.swift
+++ b/XiEditor/Document.swift
@@ -151,9 +151,14 @@ class Document: NSDocument {
         completionHandler(nil)
     }
     
+    // Document.close() can be called multiple times (on window close and application terminate)
+    fileprivate var _isClosed = false
     override func close() {
-        super.close()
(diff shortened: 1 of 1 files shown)


raphlinus:
I agree with this approach (if it's clear that Cocoa can do multiple calls). Any particular reason you didn't just set coreViewIdentifier to nil instead of creating a new boolean? I think this would also be better with race conditions where the window is closed before it's fully initialized.

But there's a particular reason, then this patch is fine.

cmyr:
No _great_ reason. I found this the simplest approach to reason about. I do think there are some benefits to your approach though, will go in that direction.

raphlinus:
Great, thanks!

XiEditor/Document.swift
```diff
@@ -150,10 +150,14 @@ class Document: NSDocument {
...
-    
+
+    // Document.close() can be called multiple times (on window close and application terminate)
     override func close() {
-        super.close()
-        Events.CloseView(viewIdentifier: coreViewIdentifier!).dispatch(dispatcher!)
+        if let identifier = self.coreViewIdentifier {
+            self.coreViewIdentifier = nil
+            Events.CloseView(viewIdentifier: identifier).dispatch(dispatcher!)
+            super.close()
```
cbrewster: Should the `super` call be outside this if?

cmyr: this is a very good question, to which I don't have an entirely satisfactory answer. The apple documentation does not mention calling super in your implementation *at all*, although super is called in some sample code. I think there's a reasonable argument for moving it outside. Given how little we are actually involved in the NSDocument lifecycle stuff (i.e. we handle all opening/closing etc in core) I'm not inclined to worry here? But moving it outside might be more correct.

cbrewster: Since we might be moving away from NSDocument its probably not worth worrying about.

cmyr: "we're moving away from nsdocument eventually" is my current favourite broom with which to sweep potential issues under the bed. ;)

PR #27, opened 2017-04-23
Don't delete marked text when mouse down

tsekityam:
Follows the behavior of Xcode, Chrome, Safari, etc.

I think we should follow majority in this situation, in order to fit
users’ expectation when they mouse down with marked text.


raphlinus:
Thanks!

PR #29, opened 2017-04-29
Add "gesture" for Command-click

raphlinus:
Command-click now sends a "gesture" rpc with a type of "toggle_sel",
which will toggle the selection, allowing for multiple cursors.

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -158,7 +158,15 @@ class EditViewController: NSViewController, EditViewDataSource {
         lastDragPosition = position
         let flags = theEvent.modifierFlags.rawValue >> 16
         let clickCount = theEvent.clickCount
-        document.sendRpcAsync("click", params: [position.line, position.column, flags, clickCount])
+        if theEvent.modifierFlags.contains(NSCommandKeyMask) {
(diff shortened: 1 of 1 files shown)


XiEditor/EditViewController.swift
```diff
@@ -158,7 +158,15 @@ class EditViewController: NSViewController, EditViewDataSource {
         lastDragPosition = position
         let flags = theEvent.modifierFlags.rawValue >> 16
         let clickCount = theEvent.clickCount
-        document.sendRpcAsync("click", params: [position.line, position.column, flags, clickCount])
+        if theEvent.modifierFlags.contains(NSCommandKeyMask) {
```
cmyr: this is also capturing any combination of modifier flags that _includes_ command, e.g. ctrl+shift+command. Is that desired behaviour?

raphlinus: My feeling is that it's "ok for now," ie there isn't any other gesture I want to map to those other modifier chords. The idea is that this code will evolve into a nice tree of `.contains` predicates bottoming out into gesture identifiers.

cmyr: okay, sounds good.

PR #30, opened 2017-05-04
Add handling for incoming RPC requests

rkusa:
Instead of handling every incoming JSON RPC call as either a notification
or a response, the frontend is now also able to properly handle requests.

### DIFF
--- a/XiEditor/CoreConnection.swift
+++ b/XiEditor/CoreConnection.swift
@@ -96,7 +96,7 @@ class CoreConnection {
     func handleRpc(_ json: Any) {
         if let obj = json as? [String: Any], let index = obj["id"] as? Int {
             if let result = obj["result"] { // is response
-                var callback: ((Any?) -> ())? = nil
+                var callback: ((Any?) -> ())?
                 queue.sync {
                     callback = self.pending.removeValue(forKey: index)
                 }
--- a/XiEditor/CoreConnection.swift
+++ b/XiEditor/CoreConnection.swift
@@ -18,14 +18,14 @@ class CoreConnection {
 
     var inHandle: FileHandle  // stdin of core process
     var recvBuf: Data
-    var callback: (AnyObject) -> ()
+    var callback: (Any) -> Any?
 
     // RPC state
     var queue = DispatchQueue(label: "com.levien.xi.CoreConnection", attributes: [])
     var rpcIndex = 0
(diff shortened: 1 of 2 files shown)


XiEditor/CoreConnection.swift
```diff
@@ -84,14 +84,34 @@ class CoreConnection {
...
+                var callback: ((Any?) -> ())? = nil
+                queue.sync {
+                    callback = self.pending.removeValue(forKey: index)
+                }
+                callback?(result)
+            } else { // is request
                 DispatchQueue.main.async {
-                    self.callback(json as AnyObject)
+                    let result = self.callback(json as AnyObject)
+                    let resp = ["id": index, "result": result] as [String : Any?]
```
rkusa: Since we are not 100% JSON RPC spec complaint anyway, I have omitted the `"error"` property, because it is not used anywhere in the frontend-core communication, isn't it? However, I am of course fine with adding it.

XiEditor/CoreConnection.swift
```diff
@@ -18,14 +18,14 @@ class CoreConnection {
 
     var inHandle: FileHandle  // stdin of core process
     var recvBuf: Data
-    var callback: (AnyObject) -> ()
+    var callback: (Any) -> Any?
```
rkusa: Changed`AnyObject` to `Any` for consistency reasons (also, `init` method below expects `Any` instead of `AnyObject`)

XiEditor/CoreConnection.swift
```diff
@@ -84,14 +84,34 @@ class CoreConnection {
...
+            print("json error \(error.localizedDescription)")
+        }
+    }
+
+    /// handle a JSON RPC call. Determines whether it is a request, response or notifcation
+    /// and executes/responds accordingly
+    func handleRpc(_ json: Any) {
+        if let obj = json as? [String: Any], let index = obj["id"] as? Int {
+            if let result = obj["result"] { // is response
+                var callback: ((Any?) -> ())? = nil
```
raphlinus: The `= nil` isn't necessary here, and doesn't seem to add to clarity, because it will always be assigned in the sync block below.

rkusa:
I've addressed the comment and also learned about the YAGNI principle. Thanks for the review!

raphlinus:
Thanks!

Issue #23, opened 2017-04-21
move line length calculations to use CTLineEnumerateCaretOffsets


cmyr:
According to Apple, this is the suggested way of doing line measurement, and should have linear performance. This is related to #20.

cmyr:
As per @rkusa's demonstration in #31, this is not a suitable solution.

PR #31, opened 2017-05-05
Use CTLineEnumerateCaretOffsets instead of CTLineGetOffsetForStringIndex

### LINKED ISSUE #20: Choking on Long Lines
Using xi-mac, I opened a relatively large XML file (2.1MB) and was quickly met with the spinny beachball of death.

Given the rope data structure design decision, I thought I'd use this as a sort of ground-truth test for large file handling.

Let me know if this ticket isn't appropriate, or if it isn't in the appropriate place.

Including file for independent verification.

Relates to xi-core #210
[Jobe_142274871.txt](https://github.com/google/xi-mac/files/930815/Jobe_142274871.txt)

rkusa:
This was fun. No real documentation and no examples for `CTLineEnumerateCaretOffsets`. Knowing the purpose of the arguments of the closure where a funny guessing game 😄

Implements #23
However, does not seem to fix #20

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -185,9 +185,8 @@ class EditView: NSView, NSTextInputClient {
             let y = dataSource.textMetrics.linespace * CGFloat(lineIx + 1)
             context.setFillColor(textSelectionColor.cgColor)
             for selection in selections {
-                let selStart = CTLineGetOffsetForStringIndex(ctline, selection.range.location, nil)
-                let selEnd = CTLineGetOffsetForStringIndex(ctline, selection.range.location + selection.range.length, nil)
-                context.fill(CGRect.init(x: x0 + selStart, y: y - dataSource.textMetrics.ascent, width: selEnd - selStart, height: dataSource.textMetrics.linespace))
+                let sel = caretOffsets(ctline, at: [selection.range.location, selection.range.location + selection.range.length])
+                context.fill(CGRect.init(x: x0 + sel[0], y: y - dataSource.textMetrics.ascent, width: sel[1] - sel[0], height: dataSource.textMetrics.linespace))
             }
             
         }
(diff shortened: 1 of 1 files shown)


rkusa:
So I finally did @cmyr's benchmark for this change.

Before, with `CTLineGetOffsetForStringIndex`.

```
24 bytes took 0.00631099939346313
16384 bytes took 0.138375043869019
32768 bytes took 0.537413954734802
65536 bytes took 2.2434070110321
131072 bytes took 8.7914400100708
262144 bytes took 39.7263059616089
```

With `CTLineEnumerateCaretOffsets`:

```
24 bytes took 0.000867009162902832
16384 bytes took 0.157842993736267
32768 bytes took 0.564494967460632
65536 bytes took 2.19510501623154
131072 bytes took 9.65227794647217
262144 bytes took 40.974162042141
```

So, no improvement at all.

cmyr:
also apparently there's a typo in my benchmark: the first test should be 1024 bytes 😕

jjgod:
It's likely that your bottleneck is elsewhere. Would it be possible to create a standalone test case for the benchmark with this?

raphlinus:
@jjgod Did you look at the code posted in the radar attached to #20?

jjgod:
I did. I believe the suggestion meant using `CTLineEnumerateCaretOffsets` to save all the caret offsets instead of using it as a drop-in replacement of `CTLineGetOffsetForStringIndex`. Which means it's not going to be significantly faster until you change your algorithm.

cmyr:
@jjgod this issue is definitely something we'd still be interested in exploring, but it's way out of my mind at the moment. I think it should be easy to come up with some test cases if you want to play around though?

jjgod:
First of all there are some performance improvements for `CTLineGetOffsetForStringIndex()` since the radar (31689317) was filed, so you might want to try it again in current macOS. On the machine I've tested with the attached text file from #20 it's no where as slow as reported.

cmyr:
@jjgod that's very interesting, it's totally reasonable that they fixed that issue; iirc they acknowledge in private that it was a bug.

Issue #21, opened 2017-04-21
UI line number


NateLing:
![image](https://cloud.githubusercontent.com/assets/9606668/25259282/d3249384-2677-11e7-9763-04204638dc36.png)

like this picture, the line number error

cbrewster:
This happens when using a non-monospaced font.

cmyr:
yea we don't know how to measure 'character width' in a non-monospaced font. That's a holdover from older code, though, we could definitely be doing better here, by actually measuring the string in `updateGutterWidth()`

NateLing:
@cbrewster  @cmyr   i am a web developer. in CSS compute fonts width throw font-size*characters. i dont know any help for you.

rkusa:
I am wondering if the naive solution of `gutterColumns * max(stringWidth("0"), stringWidth("1"), stringWidth("2"), ..., stringWidth("9"))` (i.e. find out which of the numbers from 0-9 is the widest und use that as a replacement for the current `textMetrics.fontWidth`) is good enough?

This does not take kerning and ligatures into account, but since they usually tighten the spacing instead of widening it (this assumption could be wrong), they might not be an issue.

Otherwise, we could calculate the actual stringWidth of the numbers and maintain a cache for the widest number per column count.

cmyr:
I don't particularly dislike this approach? Seems like a reasonable hack. Another option to consider is just baking in Inconsolata, and not letting the gutter font be changed... 🤔

raphlinus:
A simpler approach would just be to measure 9's. It seems unlikely this
would be seriously wrong.

cmyr:
Another option: when we're drawing each line number, we calculate expected drawing bounds. We could check this against the current gutter width, and if the width is inadequate we could fire a notification or otherwise indicate that the gutterview needs to be resized.

We could also just stash the max of this in the gutterview, and invalidate if the line count decreases?

cbrewster:
Take the following example:
There is a file with 100+ lines, but we are only viewing the top 30 lines. The gutter will be the max width for 2 characters, but then as you scroll with lines 100+, the gutter would then resize to 3 characters wide. It would seem odd for the gutter to change width while scrolling.

cmyr:
@cbrewster yea I was thinking about the corollary of this problem during the initial implementation: in a file with 10m lines, it looks weird to have a really wide gutter with a single number.

[...]

PR #32, opened 2017-05-05
Use font based on italic and font weight settings

rkusa:
This PR is not ready, yet, but I'd like to have some feedback before continuing it (if continuing at all).

I've started working on the following code comments

```
// TODO: apply actual numeric weight
// TODO: use true italic in font if available
```

... and implemented the changes while keeping the future required string width calculation in mind.

The main concern is maybe that I made the `Style` struct font-aware. While this does work well for style-spans, it does not work well with the current `TextDrawingMetrics` that lives in the `EditViewController`. My feeling is that the `TextDrawingMetrics` could be replaced with a default Style (maybe Style 0 as the default style, and 1 the selection).

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -14,16 +14,16 @@
 
 import Cocoa
 
-//TODO: preferred font should be a user preference
-let defaultFont = CTFontCreateWithName("InconsolataGo" as CFString?, 14, nil)
-
 @NSApplicationMain
 class AppDelegate: NSObject, NSApplicationDelegate {
 
     var dispatcher: Dispatcher?
 
-    var textMetrics = TextDrawingMetrics(font: defaultFont)
-    var styleMap: StyleMap = StyleMap(font: defaultFont)
+    //TODO: preferred font should be a user preference
+    let defaultFont = CTFontCreateWithName("InconsolataGo" as CFString?, 14, nil)
+
+    lazy var textMetrics: TextDrawingMetrics = TextDrawingMetrics(font: self.defaultFont)
+    lazy var styleMap: StyleMap = StyleMap(font: self.defaultFont)
 
     func applicationWillFinishLaunching(_ aNotification: Notification) {
 
@@ -105,6 +105,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             guard let doc = doc as? Document else { continue }
             doc.editViewController?.updateGutterWidth()
             doc.editViewController?.editView.needsDisplay = true
+            doc.editViewController?.updateEditViewScroll()
         }
     }
 
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -14,12 +14,16 @@
 
 import Cocoa
 
+//TODO: preferred font should be a user preference
+let defaultFont = CTFontCreateWithName("InconsolataGo" as CFString?, 14, nil)
+
 @NSApplicationMain
 class AppDelegate: NSObject, NSApplicationDelegate {
 
     var dispatcher: Dispatcher?
-    var styleMap: StyleMap = StyleMap()
 
+    var textMetrics = TextDrawingMetrics(font: defaultFont)
+    var styleMap: StyleMap = StyleMap(font: defaultFont)
 
     func applicationWillFinishLaunching(_ aNotification: Notification) {
 
@@ -88,6 +92,20 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         }
 
         return nil
+    }
+
+    /// Passed an NSFontManager instance (as on a user-initiated font change)
+    /// computes the next set of drawing metrics and updates cached styles.
+    func handleFontChange(fontManager: NSFontManager) {
+        let newFont = fontManager.convert(textMetrics.font)
+        textMetrics = TextDrawingMetrics(font: newFont)
+        styleMap.updateFont(to: newFont)
+
+        for doc in NSApplication.shared().orderedDocuments {
+            guard let doc = doc as? Document else { continue }
+            doc.editViewController?.updateGutterWidth()
+            doc.editViewController?.editView.needsDisplay = true
+        }
     }
 
     func applicationWillTerminate(_ aNotification: Notification) {
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -42,19 +42,13 @@ struct TextDrawingMetrics {
     
     init(font: NSFont) {
         self.font = font
-        ascent = CTFontGetAscent(font)
-        descent = CTFontGetDescent(font)
-        leading = CTFontGetLeading(font)
+        ascent = font.ascender
+        descent = abs(font.descender)
+        leading = font.leading
         linespace = ceil(ascent + descent + leading)
         baseline = ceil(ascent)
         fontWidth = font.characterWidth()
-        attributes[String(kCTFontAttributeName)] = font
-    }
-    
-    /// Passed an NSFontManager instance (as on a user-initiated font change) computes the next set of drawing metrics.
-    func newMetricsForFontChange(fontManager: NSFontManager) -> TextDrawingMetrics {
-        let newFont = fontManager.convert(font)
-        return TextDrawingMetrics(font: newFont)
+        attributes[NSFontAttributeName] = font
     }
 }
 
@@ -97,7 +91,7 @@ class EditView: NSView, NSTextInputClient {
         if self.isFrontmostView {
             return NSColor.selectedTextBackgroundColor
         } else {
-        return NSColor(colorLiteralRed: 0.8, green: 0.8, blue: 0.8, alpha: 1.0)
+            return NSColor(colorLiteralRed: 0.8, green: 0.8, blue: 0.8, alpha: 1.0)
         }
     }
 
@@ -149,10 +143,6 @@ class EditView: NSView, NSTextInputClient {
     }
 
     let x0: CGFloat = 2;
-
-    let font_style_bold: Int = 1;
-    let font_style_underline: Int = 2;
-    let font_style_italic: Int = 4;
 
     override func draw(_ dirtyRect: NSRect) {
         if dataSource.document.coreViewIdentifier == nil { return }
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -43,7 +43,7 @@ struct TextDrawingMetrics {
     init(font: NSFont) {
         self.font = font
         ascent = font.ascender
-        descent = abs(font.descender)
+        descent = -font.descender // descender is returned as a negative number
         leading = font.leading
         linespace = ceil(ascent + descent + leading)
         baseline = ceil(ascent)
--- a/XiEditor/StyleMap.swift
+++ b/XiEditor/StyleMap.swift
@@ -51,7 +51,7 @@ struct Style {
         }
 
         if font == nil && weight != nil {
-            font = closestMatch(of: fromFont, traits: fm.traits(of: fromFont), weight: weight!) ?? fromFont
+            font = closestMatch(of: fromFont, traits: fm.traits(of: fromFont), weight: weight!)
         }
 
         if let font = font {
@@ -104,7 +104,7 @@ func utf8_offset_to_utf16(_ s: String, _ ix: Int) -> Int {
 /// A store of text styles, indexable by id.
 class StyleMap {
     private var font: NSFont
-    private var map: [Style?] = []
+    private var styles: [Style?] = []
 
     init(font: NSFont) {
         self.font = font
@@ -120,25 +120,26 @@ class StyleMap {
         var weight = json["weight"] as? Int
         if let w = weight {
             // convert to NSFont weight: (100-500 -> 2-6 (5 normal weight), 600-800 -> 8-10, 900 -> 12
+            // see https://github.com/google/xi-mac/pull/32#discussion_r115114037
             weight = Int(floor(1 + Float(w) * (0.01 + 3e-6 * Float(w))))
         }
         
         let style = Style(font: font, fgColor: fgColor, bgColor: bgColor,
                           underline: underline, italic: italic, weight: weight)
-        while map.count < styleID {
-            map.append(nil)
+        while styles.count < styleID {
+            styles.append(nil)
         }
-        if map.count == styleID {
-            map.append(style)
+        if styles.count == styleID {
+            styles.append(style)
         } else {
-            map[styleID] = style
+            styles[styleID] = style
         }
     }
 
     /// applies a given style to the AttributedString
     private func applyStyle(string: NSMutableAttributedString, id: Int, range: NSRange) {
-        if id >= map.count { return }
-        guard let style = map[id] else { return }
+        if id >= styles.count { return }
+        guard let style = styles[id] else { return }
 
         string.addAttributes(style.attributes, range: range)
     }
@@ -153,7 +154,7 @@ class StyleMap {
 
     func updateFont(to font: NSFont) {
         self.font = font
-        map = map.map { $0.map {
+        styles = styles.map { $0.map {
             Style(font: font, fgColor: $0.fgColor, bgColor: $0.bgColor,
                   underline: $0.underline, italic: $0.italic, weight: $0.weight)
         } }
--- a/XiEditor/StyleMap.swift
+++ b/XiEditor/StyleMap.swift
@@ -16,11 +16,55 @@ import Cocoa
 
 /// A represents a given text style.
 struct Style {
-    var fgColor: NSColor;
-    var bgColor: NSColor;
-    var weight: UInt16;
-    var underline: Bool;
-    var italic: Bool;
+    var font: NSFont?
+    var fgColor: NSColor?
+    var bgColor: NSColor?
+    var underline: Bool
+    var italic: Bool
+    var weight: Int?
+    var attributes: [String: Any] = [:]
+
+    init(font fromFont: NSFont, fgColor: NSColor?, bgColor: NSColor?, underline: Bool, italic: Bool, weight: Int?) {
+        if let fgColor = fgColor {
+            attributes[NSForegroundColorAttributeName] = fgColor
+        }
+
+        if let bgColor = bgColor, bgColor.alphaComponent != 0.0 {
+            attributes[NSBackgroundColorAttributeName] = bgColor
+        }
+
+        if underline {
+            attributes[NSUnderlineStyleAttributeName] = NSUnderlineStyle.styleSingle.rawValue
+        }
+
+        let fm = NSFontManager.shared()
+        var font: NSFont?
+
+        if italic {
+            var traits = fm.traits(of: fromFont)
+            traits.insert(NSFontTraitMask.italicFontMask)
+            if let f = closestMatch(of: fromFont, traits: traits, weight: weight ?? fm.weight(of: fromFont)) {
+                font = f
+            } else {
+                attributes[NSObliquenessAttributeName] = 0.2
+            }
+        }
+
+        if font == nil && weight != nil {
+            font = closestMatch(of: fromFont, traits: fm.traits(of: fromFont), weight: weight!) ?? fromFont
+        }
+
+        if let font = font {
+            attributes[NSFontAttributeName] = font
+        }
+
+        self.font = font
+        self.fgColor = fgColor
+        self.bgColor = bgColor
+        self.underline = underline
+        self.italic = italic
+        self.weight = weight
+    }
 }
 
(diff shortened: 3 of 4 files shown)


XiEditor/AppDelegate.swift
```diff
@@ -74,7 +74,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             }
         case "def_style":
             if let obj = params as? [String : AnyObject] {
-                styleMap.defStyle(json: obj)
+                let font = CTFontCreateWithName("InconsolataGo" as CFString?, 14, nil)
```
rkusa: This is temporary. But if we keep the Style objects font-aware, we might re-consider whether each buffer needs its own styleMap instance instead of using a global one.

XiEditor/StyleMap.swift
```diff
@@ -16,11 +16,7 @@ import Cocoa
 
 /// A represents a given text style.
 struct Style {
-    var fgColor: NSColor;
-    var bgColor: NSColor;
-    var weight: UInt16;
-    var underline: Bool;
-    var italic: Bool;
+    var attributes: [String: Any];
```
rkusa: We could also keep all the properties, but instead of creating the attributes array on every call, it is now just created once.

XiEditor/StyleMap.swift
```diff
@@ -108,4 +122,25 @@ class StyleMap {
...
+    var weight = weight
+    let fromWeight = fm.weight(of: font)
+    let direction = fromWeight > weight ? 1 : -1
+    while weight != fromWeight {
+        if let f = fm.font(withFamily: font.familyName ?? font.fontName, traits: traits, weight: weight, size: font.pointSize) {
+            return f
+        }
+        weight += direction
+    }
+    return nil
```
rkusa: This one looks for the font that best matches the targeted weight settings. (I should have added a doc comment before the function).

cmyr: you might be able to clean this up with [availableMembers(ofFontFamily:)](https://developer.apple.com/reference/appkit/nsfontmanager/1462316-availablemembers).

XiEditor/StyleMap.swift
```diff
@@ -61,15 +57,46 @@ func utf8_offset_to_utf16(_ s: String, _ ix: Int) -> Int {
...
-    func defStyle(json: [String: AnyObject]) {
+    func defStyle(for fromFont: NSFont, json: [String: AnyObject]) {
         guard let styleID = json["id"] as? Int else { return }
-        let fgColor = json["fg_color"] as? UInt32 ?? 0xFF000000
-        let bgColor = json["bg_color"] as? UInt32 ?? 0
-        let weight = json["weight"] as? UInt16 ?? 400
-        let underline = json["underline"] as? Bool ?? false
+
+        var attrs: [String: Any] = [
+            NSForegroundColorAttributeName: json["fg_color"] as? NSColor ?? NSColor.red,
```
cmyr: default color should be black.

XiEditor/StyleMap.swift
```diff
@@ -61,15 +57,46 @@ func utf8_offset_to_utf16(_ s: String, _ ix: Int) -> Int {
...
+
+        let fm = NSFontManager.shared()
+        var font: NSFont?
+
         let italic = json["italic"] as? Bool ?? false
+        // convert to NSFont weight, which is a value between 1-15 with 5 being the normal weight,
+        // from CSS-like values that are between 100-900
+        let weight = (json["weight"] as? Int ?? fm.weight(of: fromFont)) / 100 + 1
+
+        if italic {
```
cmyr: bikeshed question: if we have a true italic and a true bold, how do we prefer to render a bold italic? 🤔

rkusa: This case is handled. Since the font in this of statement takes the weight into account. The second closest match is only called ich this here did not found any match. :)

XiEditor/StyleMap.swift
rkusa: I did an implementation with `availableMembers` first, but ended up with a less understandable code. If you like, I can show of the `availableMembers` variant of this implementation? (I also did a `convertWeight` implementation,  but this API did some really strange things).

XiEditor/StyleMap.swift
raphlinus: I think the `fm.weight(of:)` method requires a slightly different conversion. https://bugs.webkit.org/show_bug.cgi?id=6484 has the info but I'll give the table below:

    100 -> 2
    200 -> 3
    300 -> 4
    400 -> 5
    500 -> 6
    600 -> 8
    700 -> 9
    800 -> 10
    900 -> 12

I heard somewhere (not sure where) that 1000 is or will be a valid CSS value. I think it should map to 15.

XiEditor/StyleMap.swift
rkusa: Oh right, the first `FF` of the previous color was the alpha part. Will be updated

cmyr: yea I made this mistake too when I was first playing around with styles for the plugin lib. 😎

XiEditor/StyleMap.swift
cmyr: i'd be curious, if you have it saved? Just throw it in a gist somewhere? But no worries if you didn't stash it.

XiEditor/StyleMap.swift
rkusa: Raph, will be updated, thanks for investigating. @cmyr I think I understood your question wrong. Currently, if there is an italic font, it will be chosen in the best matching weight, regardless of whether there is a better matching weight as non-italic font. We could also choose the better matching weight and apply the Obliqueness attribute instead. I am leaning toward using the true italic font over the better matching weight

cmyr: yea, that was my question. I prefer the current approach of using a true italic if available.

XiEditor/StyleMap.swift
rkusa: https://gist.github.com/rkusa/4266960dc20085074a90dad36fd6d2a5

A final version could use some cleanup and also some logic to keep original traits (like Condensed). But probably enough to get an idea.

XiEditor/AppDelegate.swift
raphlinus: That is open to reconsideration, yes.

XiEditor/StyleMap.swift
```diff
@@ -108,4 +122,25 @@ class StyleMap {
                 applyStyle(string: string, id: styleSpan.style, range: styleSpan.range)
         }
     }
+
+    func stringWidth(str: NSString, style id: Int) -> CGFloat? {
```
raphlinus: This is intended for consumption by line wrapping?

XiEditor/StyleMap.swift
```diff
@@ -61,15 +57,46 @@ func utf8_offset_to_utf16(_ s: String, _ ix: Int) -> Int {
...
+            attrs[NSUnderlineStyleAttributeName] = NSUnderlineStyle.styleSingle.rawValue
+        }
+
+        let fm = NSFontManager.shared()
+        var font: NSFont?
+
         let italic = json["italic"] as? Bool ?? false
+        // convert to NSFont weight: (100-500 -> 2-6 (5 normal weight), 600-800 -> 8-10, 900 -> 12
+        let w = Float(json["weight"] as? Int ?? fm.weight(of: fromFont))
+        let weight = Int(floor(1 + w * (0.01 + 3e-6 * w)))
```
rkusa: @raphlinus Updated to use the suggested formula. Hower, it currently maps `1000` to `14` (and not `15`).

If this is a issue, I could also update the logic to a select, e.g.

```swift
func fontWeight(_ input: Int) -> Int {
    switch input {
    case 100...500:
        return input / 100 + 1
    case 600...800:
        return input / 100 + 2
    case 900:
        return 12
    case 1000:
        return 15
    default:
        return 5
    }
}
```

XiEditor/StyleMap.swift
rkusa: I had this in mind at first, however, using the `.size().width` method of an `NSAttributedString` may be more appropriate to also handle the case where parts of a word have different fonts.

However, it could be useful to fix #21.

But it could probably be be removed from this PR and re-add it once it is actually used.

rkusa:
Thanks for the reviews! I've pushed a commit addressing the comments.

The remaining open questions is:

Should all usages of `TextDrawingMetrics` in `EditViewController` use a style instead? (requires some more reserved styles like `0 - default`, `1 - selection`, `2- gutter`).
If it helps, I could prepare a PR showing exactly what I have in mind.

XiEditor/StyleMap.swift
cmyr: if you only need an exception for that case you could also just do `let weight = w >= 1000 ? 15 : Int(floor(1 + w * (0.01 + 3e-6 * w)))`

rkusa: Yes, that would be less code and I guess the comment above is enough to not require a reader to quickly understand the formula. One (not clearly mentioned) part of my question was, if we want to add this case at all, since it is somehow unclear whether `1000` is already a valid CSS font weight. :)

cmyr: I like having the code self-document but you probably need a comment for the rationale anyway. We aren't _actually_ using css weights, so we can use 1000 if we want to, and it makes sense to have the full system-expressible range available?

I do appreciate that in general any approach I suggest is something you've already tried and decided against for good reasons, but I've got to keep suggesting them. 😉

raphlinus:
I think 14 is actually reasonable, after all 100 maps to 2, not 1.

My gut feeling is that all usages should use a style, and I can see that slightly expanding the pre-defined styles.

XiEditor/StyleMap.swift
rkusa: Yes, please keep giving suggestions! 😊

XiEditor/EditView.swift
```diff
@@ -228,7 +183,7 @@ class EditView: NSView, NSTextInputClient {
             // probably want to move to using CTLineDraw instead of drawing the attributed string,
             // but that means drawing the selection highlight ourselves (which has other benefits).
             //attrString.drawAtPoint(NSPoint(x: x0, y: y - 13))
-            let y = dataSource.textMetrics.linespace * CGFloat(lineIx + 1);
+            let y = defaultStyle.linespace * CGFloat(lineIx + 1);
```
rkusa: I am not sure, but would consider it possible that different font weights have a different `linespace`, which is not taken into account here. I hope we can ignore it and just keep using the default style `linespace` 🤔

cmyr: This sounds like a question @raphlinus is very equipped to answer. 😁

XiEditor/StyleMap.swift
```diff
@@ -14,13 +14,87 @@
...
 struct Style {
-    var fgColor: NSColor;
-    var bgColor: NSColor;
-    var weight: UInt16;
-    var underline: Bool;
-    var italic: Bool;
+    var font: NSFont
+    var fgColor: NSColor
+    var bgColor: NSColor
+    var setTextBgColor: Bool
```
cmyr: `wantsTextBgColor`? 

But I'm not sure I totally understand: text background color should be a distinct property from global background color. The concept of a global background color should exist outside of `Style`, which should more accurately be thought of as `TextStyle` (This renaming might be helpful?). How we end up doing non-text UI styling is something to discuss.

XiEditor/EditView.swift
```diff
@@ -166,9 +114,15 @@ class EditView: NSView, NSTextInputClient {
         path2.fill()
         */
 
+        let defaultStyle = dataSource.styleMap.defaultStyle
+
+        // set view background
+        self.wantsLayer = defaultStyle.bgColor.alphaComponent != 0.0
+        self.layer?.backgroundColor = defaultStyle.bgColor.cgColor
```
cmyr: In any case, I don't think we should be modifying our layer directly here:

[wantsLayer](https://developer.apple.com/reference/appkit/nsview/1483695-wantslayer#):
>In a layer-backed view, any drawing done by the view is cached to the underlying layer object. This cached content can then be manipulated in ways that are more performant than redrawing the view contents explicitly. AppKit automatically creates the underlying layer object (using the makeBackingLayer() method) and handles the caching of the view’s content. If the wantsUpdateLayer method returns false, you should not interact with the underlying layer object directly. Instead, use the methods of this class to make any changes to the view and its layer. If wantsUpdateLayer returns true, it is acceptable (and appropriate) to modify the layer in the view’s updateLayer() method.

At some point in the future I expect us to go down to the CALayer level for all of our drawing, but for now I think we should stick with the higher level drawing API, and just use `NSRectFill`.

Keep the bg stuff in here, I think; when the style changes fire `editView.needsDisplay = True` to invalidate the view's contents, and then it will use the new defaultStyle on the next drawing cycle.

cmyr: Also see my comment about `setTextBgColor`: the `bgColor` field of a style should not be interpreted as the view's background color, as fun as having that working is. 😢

rkusa: `editView.needsDisplay` sounds good. I'll have a look how to set the background with `NSRectFill` instead of using the layer.

XiEditor/StyleMap.swift
rkusa: Maybe my thinking is somehow influenced, because I've created a Sublime color scheme recently (which is not to be confused with a sublime ui theme). There you have a default foreground and background color, but setting the default text background color gives you

screen shot 2017-05-09 at 07 12 22

and not

screen shot 2017-05-09 at 07 01 51


The same applies to the gutter foreground and background color (which are part of the color scheme in Sublime).

screen shot 2017-05-09 at 07 15 15

vs

screen shot 2017-05-09 at 07 15 24

Which I think is the expected behaviour when creating a color scheme and I wouldn't consider both examples a part of a distinct UI-theming.

This example is also where the `setTextBgColor` comes from. Because when I am setting the "global" textview/gutter background, it is not necessary to also set the text background (even though you wouldn't notice because they are of the same color - but it felt wrong).

Again, I am open to the change these things, just want to give my reasoning first.

XiEditor/EditView.swift
cmyr: yes, all drawing has to stay in the draw call.

rkusa:
I've updated the implementation. I am not using `Style`s for editor and gutter theming anymore. However, I also haven't created a `ColorTheme` struct as discussed in IRC yesterday. Since there is no coupling between `Style` and `ColorTheme` anymore, since the `ColorTheme` wouldn't be customisable yet (i.e. no extra functionality), and since it would therefore just bloat the PR, I decided to leave it out for now (but I have it in the stash for someday in the future). I also kept the `TextDrawingMetrics` for now (I postponed my thoughts of how to refactor them, mostly to reduce the amount of changes and also since I've not included the `ColorTheme` changes).

I've also already squashed the commits (sorry for that, but having the last changes as a single commit wouldn't help review, because it reverted a lot of single lines).

Ready again for review.

XiEditor/StyleMap.swift
```diff
@@ -16,11 +16,55 @@ import Cocoa
...
+            traits.insert(NSFontTraitMask.italicFontMask)
+            if let f = closestMatch(of: fromFont, traits: traits, weight: weight ?? fm.weight(of: fromFont)) {
+                font = f
+            } else {
+                attributes[NSObliquenessAttributeName] = 0.2
+            }
+        }
+
+        if font == nil && weight != nil {
+            font = closestMatch(of: fromFont, traits: fm.traits(of: fromFont), weight: weight!) ?? fromFont
```
cmyr: closest match could just return the input value itself if there's no match. Not sure if I like this better, just musing.

XiEditor/StyleMap.swift
```diff
@@ -59,17 +103,28 @@ func utf8_offset_to_utf16(_ s: String, _ ix: Int) -> Int {
...
-        let bgColor = json["bg_color"] as? UInt32 ?? 0
-        let weight = json["weight"] as? UInt16 ?? 400
+
+        let fgColor = colorFromArgb(json["fg_color"] as? UInt32 ?? 0xFF000000)
+        let bgColor = colorFromArgb(json["bg_color"] as? UInt32 ?? 0)
         let underline = json["underline"] as? Bool ?? false
         let italic = json["italic"] as? Bool ?? false
+        var weight = json["weight"] as? Int
+        if let w = weight {
+            // convert to NSFont weight: (100-500 -> 2-6 (5 normal weight), 600-800 -> 8-10, 900 -> 12
```
cmyr: i'd be tempted to include a link to the discussion issue inline, for future archaeologists

XiEditor/StyleMap.swift
```diff
@@ -108,4 +150,26 @@ class StyleMap {
                 applyStyle(string: string, id: styleSpan.style, range: styleSpan.range)
         }
     }
+
+    func updateFont(to font: NSFont) {
+        self.font = font
+        map = map.map { $0.map {
```
cmyr: this line suggests strongly to me that we should rename `map`.

XiEditor/AppDelegate.swift
```diff
@@ -14,12 +14,16 @@
 
 import Cocoa
 
+//TODO: preferred font should be a user preference
+let defaultFont = CTFontCreateWithName("InconsolataGo" as CFString?, 14, nil)
```
cmyr: Two options:

1. if this is going to be a global, go `FULL_GLOBAL_STYLE`. Otherwise I think it could live comfortably in the app delegate itself, or as a `static` (class) property on `TextDrawingMetrics` (or somewhere else for that matter).

XiEditor/EditView.swift
```diff
@@ -42,19 +42,13 @@ struct TextDrawingMetrics {
     
     init(font: NSFont) {
         self.font = font
-        ascent = CTFontGetAscent(font)
-        descent = CTFontGetDescent(font)
-        leading = CTFontGetLeading(font)
+        ascent = font.ascender
+        descent = abs(font.descender)
```
cmyr: Any reason for choosing `abs` here? I suspect there are situations where this could cause `linespace` to be incorrect.

XiEditor/StyleMap.swift
rkusa: Actually, the `?? fromFont` wasn't supposed to be there anymore, since I prefer to not set the `NSFontAttributeName` if the default font is used. I'll remove it

XiEditor/StyleMap.swift
rkusa: you mean this one https://bugs.webkit.org/show_bug.cgi?id=6484 ?

rkusa:
@cmyr Thanks for the review! I've added a commit addressing your comments. If a change is not what you may have anticipated, feel free to ask for a change again! 😊

XiEditor/StyleMap.swift
cmyr: I was thinking of the comment from raph explaining the weighting and offered the little formula actually.

XiEditor/AppDelegate.swift
```diff
@@ -14,14 +14,14 @@
...
-//TODO: preferred font should be a user preference
-let defaultFont = CTFontCreateWithName("InconsolataGo" as CFString?, 14, nil)
-
 @NSApplicationMain
 class AppDelegate: NSObject, NSApplicationDelegate {
 
     var dispatcher: Dispatcher?
 
+    //TODO: preferred font should be a user preference
+    static let defaultFont = CTFontCreateWithName("InconsolataGo" as CFString?, 14, nil)
```
cmyr: heh if we're going to put this in app delegate then it should be an instance variable, because app delegate is already a singleton. Sorry if I was unclear 😂

XiEditor/AppDelegate.swift
```diff
@@ -90,6 +94,20 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+    /// computes the next set of drawing metrics and updates cached styles.
+    func handleFontChange(fontManager: NSFontManager) {
+        let newFont = fontManager.convert(textMetrics.font)
+        textMetrics = TextDrawingMetrics(font: newFont)
+        styleMap.updateFont(to: newFont)
+
+        for doc in NSApplication.shared().orderedDocuments {
+            guard let doc = doc as? Document else { continue }
+            doc.editViewController?.updateGutterWidth()
+            doc.editViewController?.editView.needsDisplay = true
```
cmyr: oh, one last thing: we might want, on font change, to re send the `scroll` notification? Otherwise there may be visual jank when changing font sizes.

cmyr: yea so if you change from a font that fits the document in the current window to a font which needs scroll, scrolling isn't available until you next send an update (like by moving the cursor)

cmyr:
Okay, looks good to me!

raphlinus:
Thanks!

PR #33, opened 2017-05-07
multi-plugin support

cmyr:
This branch adds support for the multi-plugin stuff in https://github.com/google/xi-editor/pull/255.

Plugins are listed in a dropdown in the Debug menu:

screen shot 2017-05-07 at 1 24 56 pm

And running multiple plugins is now fairly easy (here is syntect + the very simple bracket closing sample plugin)

multi_plug

**note**: to run the python plugins, you will need to install the plugin lib: from `xi-editor/python/xi_plugins` run `pip3 install .`.

**also**: you may run into problems with Xcode resetting environment variables. To set environment variables in Xcode, option-click on the Run button to open the scheme editor, and set variables there:

screen shot 2017-05-07 at 1 33 25 pm


trishume:
Not related to this PR but the gif you posted:

Something is terribly wrong. In the gif there are times when the character you type isn't highlighted instantly (when you are typing the comment). There's no way that should ever happen. Anyone have any idea why highlighting isn't completing properly within the space of one frame? Because that's like worse-than-Atom level slowness.

raphlinus:
A couple of observations from playing with it. First, it doesn't work with master xi-mac, it looks like the protocol has changed incompatibly.

Second, the syntect plugin looks like it needs an update to track the plugin protocol changes:

```
plugin response json err: ErrorImpl { code: Message("data did not match any variant of untagged enum UpdateResponse"), line: 0, column: 0 }
```

This may actually have been previous plugin changes (to add an edit response), not this PR. Also, it works, just prints these errors. In any case, I definitely want to track it.

raphlinus:
Above comment was meant to be posted to xi-editor, and is now https://github.com/google/xi-editor/pull/255#issuecomment-299745751 , as these are more core than FE concerns. However, leaving it, as there is some overlap.

raphlinus:
Really cool to see this.

PR #35, opened 2017-05-08
Fix for cache de-sync

### LINKED ISSUE #271: gutter flashes when a new view is opened
There's no reason we can't draw the gutter correctly immediately; this is probably similar to #257.

raphlinus:
Re-reading the logic it seems I probably missed a case.

Fixes https://github.com/google/xi-editor/issues/271


raphlinus:
I didn't verify this carefully, but does this fix it for you?

trishume:
I tried to reproduce it with this branch and I can't seem to do so anymore, so I think this fix worked.

PR #39, opened 2017-05-15
Add experimental "vim mode"

raphlinus:
When enabled, forwards all ASCII keys (including control) to a vim_key
editor command.

Progress toward google/xi-editor#93

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -155,6 +155,14 @@ class EditViewController: NSViewController, EditViewDataSource {
     
     // MARK: - System Events
     override func keyDown(with theEvent: NSEvent) {
+        if (NSApplication.shared().delegate as! AppDelegate).vimMode {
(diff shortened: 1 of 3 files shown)


XiEditor/EditViewController.swift
```diff
@@ -155,6 +155,14 @@ class EditViewController: NSViewController, EditViewDataSource {
     
     // MARK: - System Events
     override func keyDown(with theEvent: NSEvent) {
+        if (NSApplication.shared().delegate as! AppDelegate).vimMode {
+            if let codepoint = theEvent.characters?.unicodeScalars.first?.value {
+                if codepoint > 0 && codepoint < 128 {
```
vlovich: Is this just checking for 7-bit ascii?  If so, is there any link to documentation of where these magic values come from?  Also, would be helpful to understand why only the first key is being checked.

jansol:
This can probably be closed as obsolete?

PR #40, opened 2017-05-17
Optionally log client RPC traffic

cmyr:
If a filepath is supplied through the $XI_CLIENT_RPC_LOG environment
variable, all outgoing RPC traffic will be logged to that path. This can
then be used for testing and profiling.

### DIFF
--- a/XiEditor/CoreConnection.swift
+++ b/XiEditor/CoreConnection.swift
@@ -14,18 +14,62 @@
 
 import Foundation
 
+/// Env var used to specify a path for logging RPC messages.
+/// These logs can be used for profiling & debugging.
+let XI_RPC_LOG = "XI_CLIENT_RPC_LOG"
+
+/// Error tolerant wrapper for append-writing to a file.
+struct FileWriter {
+    let path: URL
+    let handle: FileHandle
+    
+    init?(path: String) {
+        let path = NSString(string: path).expandingTildeInPath
+        if FileManager.default.fileExists(atPath: path) {
+            print("file exists at \(path), will not overwrite")
+            return nil
+        }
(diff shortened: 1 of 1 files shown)


trishume:
For it to be useful for profiling, shouldn't it also log a high precision timestamp with each message?

cmyr:
I was thinking about this. The core has to handle all rpc messages sequentially, so shouldn't this be acceptable for profiling core performance?

trishume:
I don't see how this log would show any difference between one core and another that was uniformly 10x faster, the order would be the same, it doesn't tell you how low latency things are.

But it's likely that I just don't understand this since I've never looked at the frontend interface before.

cmyr:
@trishume I think I understand our confusion: the purpose of this is to make a client's commands replayable to xi-core, via something like `cat`, not to be itself a useful record of performance information.

trishume:
Oh okay that makes more sense. If that's the case though, why do you log the commands in parens instead of just raw? Doesn't that stop them from working with `cat`?

cmyr:
each line is the raw json as it would go over the wire, unless I've screwed something up. (Very possible)

trishume:
nvm that was just me not remembering the Swift string interpolation syntax.

XiEditor/CoreConnection.swift
```diff
@@ -14,18 +14,62 @@
...
+        
+        do {
+            try self.handle = FileHandle(forWritingTo: self.path)
+        } catch let err as NSError {
+            print("error opening log file \(err)")
+            return nil
+        }
+    }
+
+    func write(string: String) {
```
raphlinus: looks like dead code

PR #46, opened 2017-05-28
Support adding next/prev line to selection

cmyr:
This adds support for https://github.com/google/xi-editor/pull/320


raphlinus:
Thanks! Here's a question about the keybinding though: on my system (Sierra with very afaik very minimal customization), ⌃⇧↑ (and ↓) are bound to exposé. This is something users will have to manually reconfigure if they want access?

cmyr:
yes, I just duplicated the sublime keybindings. you would have to remove the exposé keybindings to... expose these ones.

PR #47, opened 2017-05-29
Implement find UI

rkusa:
screen shot 2017-05-29 at 18 53 38

screen shot 2017-05-29 at 18 54 18

A simple UI for google/xi-editor#286

Edit: I could also add the find view at the bottom of the window, if this is preferred

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -103,7 +103,8 @@ class EditView: NSView, NSTextInputClient {
     var cursorPos: (Int, Int)?
     fileprivate var _selectedRange: NSRange
     fileprivate var _markedRange: NSRange
-    
+
+    var isFirstResponder = false
     var isFrontmostView = false {
         didSet {
             //TODO: blinking should one day be a user preference
@@ -257,9 +258,16 @@ class EditView: NSView, NSTextInputClient {
         return true;
     }
 
-    override func mouseDown(with event: NSEvent) {
-        super.mouseDown(with: event)
+    override func becomeFirstResponder() -> Bool {
         isFrontmostView = true
+        isFirstResponder = true
+        return true
+    }
+
+    override func resignFirstResponder() -> Bool {
+        isFrontmostView = false
+        isFirstResponder = false
+        return true
     }
 
     // we use a flipped coordinate system primarily to get better alignment when scrolling
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -257,6 +257,11 @@ class EditView: NSView, NSTextInputClient {
         return true;
     }
 
+    override func mouseDown(with event: NSEvent) {
+        super.mouseDown(with: event)
+        isFrontmostView = true
+    }
+
     // we use a flipped coordinate system primarily to get better alignment when scrolling
     override var isFlipped: Bool {
         return true;
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -94,6 +94,7 @@ class EditView: NSView, NSTextInputClient {
             return NSColor(colorLiteralRed: 0.8, green: 0.8, blue: 0.8, alpha: 1.0)
         }
     }
+    var textHighlightColor: NSColor = NSColor(deviceWhite: 0.8, alpha: 0.4)
 
     var lastDragLineCol: (Int, Int)?
     var timer: Timer?
@@ -166,20 +167,30 @@ class EditView: NSView, NSTextInputClient {
             dataSource.document.sendRpcAsync("request_lines", params: [f, l])
         }
 
-        // first pass, for drawing background selections
+        // first pass, for drawing background selections and search highlights
         for lineIx in first..<last {
-            guard let line = getLine(lineIx), line.containsSelection == true else { continue }
-            let selections = line.styles.filter { $0.style == 0 }
+            guard let line = getLine(lineIx), line.containsReservedStyle == true else { continue }
             let attrString = NSMutableAttributedString(string: line.text, attributes: dataSource.textMetrics.attributes)
             let ctline = CTLineCreateWithAttributedString(attrString)
             let y = dataSource.textMetrics.linespace * CGFloat(lineIx + 1)
+
             context.setFillColor(textSelectionColor.cgColor)
+            let selections = line.styles.filter { $0.style == 0 }
             for selection in selections {
                 let selStart = CTLineGetOffsetForStringIndex(ctline, selection.range.location, nil)
                 let selEnd = CTLineGetOffsetForStringIndex(ctline, selection.range.location + selection.range.length, nil)
-                context.fill(CGRect.init(x: x0 + selStart, y: y - dataSource.textMetrics.ascent, width: selEnd - selStart, height: dataSource.textMetrics.linespace))
+                context.fill(CGRect(x: x0 + selStart, y: y - dataSource.textMetrics.ascent,
+                                    width: selEnd - selStart, height: dataSource.textMetrics.linespace))
             }
-            
+
+            context.setFillColor(textHighlightColor.cgColor)
+            let highlights = line.styles.filter { $0.style == 1 }
+            for highlight in highlights {
+                let selStart = CTLineGetOffsetForStringIndex(ctline, highlight.range.location, nil)
+                let selEnd = CTLineGetOffsetForStringIndex(ctline, highlight.range.location + highlight.range.length, nil)
+                context.fill(CGRect(x: x0 + selStart, y: y - dataSource.textMetrics.ascent,
+                                    width: selEnd - selStart, height: dataSource.textMetrics.linespace))
+            }
         }
         // second pass, for actually rendering text.
         for lineIx in first..<last {
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -32,9 +32,20 @@ class EditViewController: NSViewController, EditViewDataSource {
     @IBOutlet weak var editContainerView: EditContainerView!
     @IBOutlet var editView: EditView!
     @IBOutlet weak var gutterView: GutterView!
+    @IBOutlet weak var stackView: NSStackView!
     
     @IBOutlet weak var gutterViewWidth: NSLayoutConstraint!
     @IBOutlet weak var editViewHeight: NSLayoutConstraint!
+
+    lazy var findViewController: FindViewController! = {
+        let storyboard = NSStoryboard(name: "Main", bundle: nil)
+        let controller = storyboard.instantiateController(withIdentifier: "Find View Controller") as! FindViewController
+        controller.editViewController = self
+
+        self.stackView.insertView(controller.view, at: 0, in: .top)
+        
+        return controller
+    }()
     
     var document: Document!
     
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -52,6 +52,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate {
         controller.view.topAnchor.constraint(equalTo: self.view.topAnchor).isActive = true
         controller.view.leadingAnchor.constraint(equalTo: self.view.leadingAnchor).isActive = true
         controller.view.trailingAnchor.constraint(equalTo: self.view.trailingAnchor).isActive = true
+        controller.view.isHidden = true
         return controller
     }()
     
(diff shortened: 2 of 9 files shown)


cmyr:
Very excited to see this! 

Haven't read the code but I've played around with the branch, looks very solid. I have some UI nits / thoughts, but overall this is great. :)

nits: 

1. I don't think the cursor should blink in the `EditView` while the search view is active.
2. I think `<return>` in the search bar should either move focus to the view, or else advance to the next selection (Xcode's behaviour).
3. The search bar should handle `<esc>` on the client side in order to dismiss itself. There's an example of this in that old statusbar branch I had up, as a reference.
4. I would disable the focus ring on the search bar (the blue outline; it's a checkbox in IB).
5. ship it

I also have a bunch of error mesages in the logs like this:

```
id 40 not found in pending
unknown json from core: {
    error = "return value missing";
    id = 41;
}```

and I'm getting a crash that I haven't been able to pin down. It's an out of bounds array access in xi-core, I'll save a backtrace if it comes up again.

**Comment by rkusa** [comment]
Thanks for the feedback!

> 1. I don't think the cursor should blink in the EditView while the search view is active.

tbh, I don't know how to best implement this. Seems like there are now life cycle events when switching focus between views of the same window?

> 2. I think <return> in the search bar should either move focus to the view, or else advance to the next selection (Xcode's behaviour).

Should advance the to the next occurrence now (with the latest xi-editor#feature/search branch).

> 3. The search bar should handle <esc> on the client side in order to dismiss itself. There's an example of this in that old statusbar branch I had up, as a reference.

Implemented. (but only for the case where the search bar is focused)

> 4. I would disable the focus ring on the search bar (the blue outline; it's a checkbox in IB).

Sounds good. Done

> I also have a bunch of error mesages in the logs like this:

Should be fixed in the latest xi-editor#feature/search branch

> and I'm getting a crash that I haven't been able to pin down. It's an out of bounds array access in xi-core, I'll save a backtrace if it comes up again.

In the latest xi-editor#feature/search branch, I have replaced some code that does direct index access. But maybe the error was somewhere else.

**Comment by trishume** [comment]
The search bar should also dismiss itself when the search bar isn't focused, which is more difficult but at least in Sublime is something I use super often.

File: XiEditor/EditView.swift [hunk 1]
```diff
@@ -94,6 +94,7 @@ class EditView: NSView, NSTextInputClient {
             return NSColor(colorLiteralRed: 0.8, green: 0.8, blue: 0.8, alpha: 1.0)
         }
     }
+    var textHighlightColor: NSColor = NSColor(deviceWhite: 0.8, alpha: 0.4)
```
cmyr: I guess this will come from themes soon? (We will be able to do themes very soon)

XiEditor/EditView.swift
```diff
@@ -166,20 +167,30 @@ class EditView: NSView, NSTextInputClient {
...
+                                    width: selEnd - selStart, height: dataSource.textMetrics.linespace))
+            }
+
+            context.setFillColor(textHighlightColor.cgColor)
+            let highlights = line.styles.filter { $0.style == 1 }
+            for highlight in highlights {
+                let selStart = CTLineGetOffsetForStringIndex(ctline, highlight.range.location, nil)
+                let selEnd = CTLineGetOffsetForStringIndex(ctline, highlight.range.location + highlight.range.length, nil)
+                context.fill(CGRect(x: x0 + selStart, y: y - dataSource.textMetrics.ascent,
+                                    width: selEnd - selStart, height: dataSource.textMetrics.linespace))
```
cmyr: Would be nice to not repeat ourselves here, since the only thing that's really different is the color and the set of lines. Totally happy with it like this though, still planning on redoing all the drawing code in not-too-long.

XiEditor/EditViewController.swift
```diff
@@ -32,9 +32,20 @@ class EditViewController: NSViewController, EditViewDataSource {
     @IBOutlet weak var editContainerView: EditContainerView!
     @IBOutlet var editView: EditView!
     @IBOutlet weak var gutterView: GutterView!
+    @IBOutlet weak var stackView: NSStackView!
```
cmyr: I'm raising an eyebrow at this choice, seems like not a great fit for this problem? But 🤷‍♂️.

XiEditor/Search.swift
```diff
@@ -0,0 +1,208 @@
...
+        let recentClear = NSMenuItem(title: "Clear Recent Searches", action: nil, keyEquivalent: "")
+        recentClear.tag = Int(NSSearchFieldClearRecentsMenuItemTag)
+        menu.addItem(recentClear)
+    }
+
+    override func validateMenuItem(_ menuItem: NSMenuItem) -> Bool {
+        return true
+    }
+
+    @IBAction func selectIgnoreCaseMenuItem(_ sender: Any) {
```
cmyr: you can specify a particular sender in the method signature, e.g. `sender: NSMenuItem?`.

XiEditor/Search.swift
```diff
@@ -0,0 +1,208 @@
...
+
+class FindViewController: NSViewController, NSTextFieldDelegate {
+
+    @IBOutlet weak var searchField: NSSearchField!
+    @IBOutlet weak var navigationButtons: NSSegmentedControl!
+    @IBOutlet weak var doneButton: NSButton!
+    @IBOutlet weak var ignoreCaseMenuItem: NSMenuItem!
+    @IBOutlet weak var wrapAroundMenuItem: NSMenuItem!
+
+    var optionIgnoreCase: Bool = true
```
cmyr: `var optionIgnoreCase = true` will do type inference.

XiEditor/Search.swift
```diff
@@ -0,0 +1,208 @@
...
+    var optionWrapAround: Bool = true
```
cmyr: nit: and I would leave out the `option` part of both of these names.

XiEditor/Search.swift
```diff
@@ -0,0 +1,208 @@
...
+        if let menuItem = sender as? NSMenuItem {
+            optionIgnoreCase = !optionIgnoreCase
+            if optionIgnoreCase {
+                menuItem.state = NSOnState
+            } else {
+                menuItem.state = NSOffState
+            }
+        }
+
+        editViewController.find(searchField.stringValue,
```
cmyr: I like the thinking, but it's debatable to me whether toggling the option should actually send the find action. For instance: if there's no visible result on the screen, I think it's weird if the `EditView` scrolls when we toggle this menu item.

XiEditor/Search.swift
```diff
@@ -0,0 +1,208 @@
...
+                                case_sensitive: !optionIgnoreCase,
+                                wrap_around: optionWrapAround)
+    }
+    
+    @IBAction func selectWrapAroundMenuItem(_ sender: Any) {
```
cmyr: nit: I would replace the `Item` suffix in these names with `Action`.

XiEditor/Search.swift
```diff
@@ -0,0 +1,208 @@
...
+            optionWrapAround = !optionWrapAround
+            if optionWrapAround {
+                menuItem.state = NSOnState
+            } else {
+                menuItem.state = NSOffState
+            }
+        }
+    }
+    
+    @IBAction func clickSegmentControl(_ sender: NSSegmentedControl) {
```
cmyr: nit: I would call this `segmentControlAction`, or `findButtonsAction` or `findControlAction`.

XiEditor/Search.swift
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    override func cancelOperation(_ sender: Any?) {
+        self.close()
+    }
+
+    func open() {
+        self.view.isHidden = false
+        searchField.becomeFirstResponder()
```
cmyr: `editViewController.editView.isFrontmostView = false`, and then change it back in `close()`. You will also want to set it to true if there's a mouse click in the editVIew. This isn't very elegant, just the only way I can think of at the moment. 😒

XiEditor/Search.swift
```diff
@@ -0,0 +1,208 @@
...
+    @IBOutlet weak var searchField: NSSearchField!
+    @IBOutlet weak var navigationButtons: NSSegmentedControl!
+    @IBOutlet weak var doneButton: NSButton!
+    @IBOutlet weak var ignoreCaseMenuItem: NSMenuItem!
+    @IBOutlet weak var wrapAroundMenuItem: NSMenuItem!
+
+    var optionIgnoreCase: Bool = true
+    var optionWrapAround: Bool = true
+
+    var editViewController: EditViewController!
```
cmyr: this should be a `weak var`, or we have a reference cycle. (Not the end of the world, in this case).

XiEditor/EditView.swift
rkusa: Yes! :)

XiEditor/EditView.swift
rkusa: Agreed! I probably had the information that you are planing to redo it anyway in the back of my mind, which may have made me lazy here

XiEditor/EditViewController.swift
rkusa: I have to play around with the alternative, but I guess the alternative is changing the constraints of the EditView programatically according to whether the FindBar is visible or not? If so, isn't the StackView exactly build for this use case?

XiEditor/Search.swift
rkusa: I've tested both VSC and Sublime and both re-do the current search when the case-matching is changed. However the EditView should not scoll, if there is no search result. So if this was the case, it is a bug.

A note about this function. I was surprised that I hat to implement these methods at all, because I setup a data binding between the menu items and the `optionXX` properties. However this seems to only effect the initial state of the menu items or I did something wrong.

XiEditor/EditViewController.swift
cmyr: On further review, I can offer more concrete criticism: by using a stack view, showing the find window causes the `EditView` to move, which moves the text. I would prefer (ultimately/one day) that the find view be drawn _over_ the `EditView`. But no need to do this now.

XiEditor/Search.swift
rkusa: Good to know! I did not find any good resource about naming conventions

XiEditor/Search.swift
```diff
@@ -121,7 +121,6 @@ extension EditViewController {
     func find(_ term: String?, case_sensitive: Bool, wrap_around: Bool) {
```
rkusa: Yes, good find. Will be updated with the commit that address all other comments. Was a quick commit to provide compatibility for the xi-editor PR

rkusa:
@cmyr I've addressed your comments (except the redundant code when drawing the highlights)

I also found two additional mis-behaviours:

- [ ] The find after changing the case matching is still broken
- [ ] When re-opening the Findbar, the previous search will be automatically executed and view will jump to the first occurrence

rkusa:
Thanks for the additional feedback!

> I can still get two blinking cursors by doing a search, clicking the edit view, and then clicking back in the search text field. This problem would be simpler if EditView's cursor blinking was based on its being firstReceiver, instead of having its own flag, but I'm okay with just overriding becomeFirstResponder/resignFirstResponder in the search viewcontroller and using that to disable the cursor in the edit view.

I also found some more cases where I've ended up with two active blinking cursors, like switching between windows. I've now implemented `becomeFirstResponder`/`resignFirstResponder`, but on the `EditView` and also had to keep track whether the `EditView` is first responder, to handle the `windowDidBecomeKey` correctly. Still feels somehow like a workaround, but I think improving handling of `isFrontmostView` and `showBlinkingCursor` could be something for a separate PR.

> this is related to another little inconsistency: if I do a case-insensitive search while I'm in this previous state, and then when the active find item is not a case-sensitive match, switching to case-sensitive search does not remove that selection.
> 
> edit: Is this what you meant about find after changing case being broken? 😁

Yes, and it should now be fixed. Unfortunately, I had to extend the `find_next` RPC call by an additional parameter to handle this case 😕

> is Jump To Selection supposed to do anything?

Not yet. I've wired it up wrong, which is now fixed. It should now be grayed out (correct behaviour for now). To implement this menu item we would have to implement the `centerSelectionInVisibleArea` action and probably add an RPC method that scrolls to the current selection (whatever a current selection is in our multi-select world).

> Future polish: Show the number of results somewhere? Either like Xcode does it, or else we can put it in the status bar (I'm sort of looking for an excuse to add a status bar).

My current implementation does not search the whole file, but only the parts that are visible or necessary to find the next occurrence (which could be the whole file eventually). So, we don't have such a number in the current approach (however, the current approach is still up for review/feedback, so could change).

> Future polish: should dismiss the search bar if it's active, even if it doesn't have focus.

I think this should be part of https://github.com/google/xi-editor/issues/324?

> Future polish: image assets for the prev/next icons, instead of '</>'.

I've added some integrated images. Still not perfect. It would be nice, if we could make the SegmentControl taller. However, this is not possible (at least I was not able to find a way of doing so).

screen shot 2017-06-01 at 08 13 56

raphlinus:
Thanks! Very nice to have this.

PR #49, opened 2017-06-06
Scroll past end of file

rkusa:
Allow scrolling past the of a file. The behaviour is like Sublime's, where you can scroll down until on remaining line is still visible.

I am not sure if this is something that should be enabled by default, if not, I am fine with closing this and wait until with have user preferences.

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -147,7 +147,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate {
         }
         shadowView?.updateScroll(scrollView.contentView.bounds, scrollView.documentView!.bounds)
         // if the window is resized, update the editViewHeight so we don't show scrollers unnecessarily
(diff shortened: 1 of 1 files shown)


trishume:
Xi currently seems to act the exact same as my Sublime Text does, can you describe a file/window-size/scenario where I can replicate the difference between Sublime and Xi before this patch?

raphlinus:
This is a preference in Sublime, and defaults to "off" on mac but "on" for I think most other OS's.

I personally prefer the "off" behavior. I'm ok with this going in controlled by a bool in code for now, and it should be easy to wire up when we get prefs (I think fairly soon).

rkusa:
@trishume lowest scroll position
Before PR:
screen shot 2017-06-07 at 08 25 20

After PR:
screen shot 2017-06-07 at 08 25 38

cmyr:
uh oh, I guess we _are_ going to have to do preferences soon. 😎

PR #51, opened 2017-06-20
Themes

cmyr:
With this PR, users can now select a theme from the debug menu.

This depends on https://github.com/google/xi-editor/pull/345.

screen recording 2017-06-20 at 04 07 pm

### DIFF
--- a/XiEditor/Theme.swift
+++ b/XiEditor/Theme.swift
@@ -1,0 +1,125 @@
+// Copyright 2017 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Cocoa
+
+
+/// Theme information for styling an editor instance.
+///
+/// Note: this is copied verbatim from syntect (https://github.com/trishume/syntect/blob/master/src/highlighting/theme.rs)
+/// and that is derived from TextMate's .tmTheme format. Many fields may not be used.
+struct Theme {
+    /// Text color for the view.
+    let foreground: NSColor
+    /// Backgound color of the view.
+    let background: NSColor
+    /// Color of the caret.
+    let caret: NSColor
+    /// Color of the line the caret is in.
+    /// Only used when the `higlight_line` setting is set to `true`.
+    let lineHighlight: NSColor?
+
+
+    /// Background color of regions matching the current search.
+    let findHighlight: NSColor
+    /// Background color of regions matching the current search.
+    let findHighlightForeground: NSColor?
+
+    /// Background color of the gutter.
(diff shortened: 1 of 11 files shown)


XiEditor/Theme.swift
```diff
@@ -0,0 +1,125 @@
...
+        let gutter_foreground = NSColor(jsonRgbaColor: dict["gutter_foreground"] as? [String: AnyObject] ?? [:])
+
+        let selection = NSColor(jsonRgbaColor: dict["selection"] as? [String: AnyObject] ?? [:])
+        let selection_foreground = NSColor(jsonRgbaColor: dict["selection_foreground"] as? [String: AnyObject] ?? [:])
+        let selection_border = NSColor(jsonRgbaColor: dict["selection_border"] as? [String: AnyObject] ?? [:])
+        let inactive_selection = NSColor(jsonRgbaColor: dict["inactive_selection"] as? [String: AnyObject] ?? [:])
+        let inactive_selection_foreground = NSColor(jsonRgbaColor: dict["inactive_selection_foreground"] as? [String: AnyObject] ?? [:])
+        let shadow = NSColor(jsonRgbaColor: dict["shadow"] as? [String: AnyObject] ?? [:])
+
+        let defaults = Theme.defaultTheme()
```
eyelash: I wonder whether defaults should be handled by the core and only complete themes sent to the front-end. This code here seems very likely to create unpleasant color combinations. (Yes I am somewhat obsessed with avoiding unpleasant color combinations).

trishume:
Just a note: a cool feature Sublime has is that the highlighting theme is actually per-view (or maybe per-buffer?). This flexibility is used very infrequently but it does enable nice things like https://github.com/aziz/PlainTasks and https://github.com/SublimeText-Markdown/MarkdownEditing

It seems that this code currently treats the theme as a property of the application and not the view.

XiEditor/Theme.swift
cmyr: Yea, there are a few things here that are a bit fudgy, because the client doesn't know the themes on first launch. The default values are currently (mostly) just copied from the `InspiredGithub` theme, for no good reason. I'm not sure how defaults should actually work, e.g. for the gutter view, which isn't specified in a lot of theme files but can be pretty jarring if it's a mismatch.

I'm very happy to have people being nitty about this stuff, basically. 😎

cmyr:
@trishume couldn't that be achieved for those plugins by using custom (more specific) scopes and then providing colours in a companion theme? They would then only be applied in  those buffers?

This assumes we implement theme stacking at some point, but we're sort of headed down that path already...

XiEditor/Theme.swift
eyelash: Why not just send a `theme_changed` event from core to front-end before the first `update`?

trishume:
Actually yes that would also work. It's actually even better than the Sublime way since you can use things from the users theme if you want to, for example your fancy paper-like markdown theme could use the user theme for fenced code blocks. So yah per-app themes are totally fine then, and at some point we'll implement theme stacking.

XiEditor/Theme.swift
cmyr: @eyelash um yes that sounds at first blush like a very reasonable approach.

cmyr: @eyelash After a bit more thought: I think in general we need some `init` RPC when the client first launches, for various sorts of synchronization tasks, but I think that's a separate PR. In the meantime I suggest having the client persist the theme setting and sending the `set_theme` RPC on launch.

raphlinus:
Thanks!

PR #55, opened 2017-07-11
Preliminary support for plugin commands

cmyr:
This provides very rough support for plugin commands. There's a placeholder UI for accepting user input, and plugin menus get populated with available commands.

Annoyingly, this UI isn't really testable because there aren't any available plugins that provide commands, which is maybe something I should fix. 🤔

This requires https://github.com/google/xi-editor/pull/366.

screen shot 2017-07-11 at 12 52 59 pm
screen shot 2017-07-11 at 12 53 11 pm

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -96,8 +96,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         case "plugin_started":
             guard let obj = params as? [String : AnyObject] else { print("bad params \(params)"); return nil }
                   let view_id = obj["view_id"] as! String
-//                  let document = documentForViewIdentifier(viewIdentifier: view_id)
-                  let plugin = obj["plugin"] as! String //else { print("missing plugin field in \(params)"); return nil }
+                  let plugin = obj["plugin"] as! String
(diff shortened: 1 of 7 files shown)


XiEditor/AppDelegate.swift
```diff
@@ -121,6 +121,18 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
             }
+        case "update_cmds":
+            guard let obj = params as? [String : AnyObject],
+                let view_id = obj["view_id"] as? String,
+                let document = documentForViewIdentifier(viewIdentifier: view_id),
+                let cmds = obj["cmds"] as? [[String: AnyObject]],
+                let plugin = obj["plugin"] as? String else { print("missing plugin field in \(params)"); return nil }
+            let parsedCommands = cmds.map { Command(jsonObject: $0) }
+                .filter { $0 != nil }
+                .map { $0! }
```
trishume: the `filter` and `map` after the first `map` can be eliminated by turning the first `map` into a `flatMap`.

cmyr: Ah cool, that would feel much nicer. And yes, It was pretty frustrating writing all the json wrangling stuff knowing I'm going to tear it all out in a month or two. 😒

PR #59, opened 2017-08-18
Build failed because `String does not conform to AnyObject`

joemalle:
I tried to build this repo and got an error:

    xi-mac/XiEditor/Document.swift:216:34: error: value of type 'String' does not conform to expected dictionary value type 'AnyObject'

I fixed the error by changing `AnyObject` to `Any` on line 222

Now the build passes!

Note: I am using Swift version 3.0.2 (swiftlang-800.0.63 clang-800.0.42.1)

P.S. This is my first PR so if I am making any etiquette mistakes please let me know!

### DIFF
--- a/XiEditor/Document.swift
+++ b/XiEditor/Document.swift
@@ -208,7 +208,7 @@ class Document: NSDocument {
                       "rpc": [
                         "rpc_type": "notification",
                         "method": method,
-                        "params": innerParams]] as [String: AnyObject]
+                        "params": innerParams]] as [String: Any]
 
         dispatcher.coreConnection.sendRpcAsync("plugin", params: params)
     }


raphlinus:
I think most of us are on Swift 3.1 now, but if this fixes the build for 3.0.2, I'll be happy to merge as soon as you get the email address thing for the CLA bot figured out. (if it's a big hassle, I'll accept you saying that both email addresses belong to you)

joemalle:
Ok -- yes, this does fix the build for 3.0, but I've updated to the newest Swift to be consistent.

As for the emails, I don't think I can fix it without making a new PR.  They are both my emails, but my google account and commit info were different.  Fixed that on my end.

raphlinus:
Thanks!

PR #61, opened 2017-08-27
Better PATH handling in build script

### LINKED ISSUE #58: Build does not work with zsh-only features in zshrc
Because bash is specified at the top of `build-rust-xcode.sh`, any time a user has something not supported by bash in their zshrc, the script will error and the build will hang. Below is some sample output while using my [zshrc](https://github.com/belak/dotfiles/blob/master/zshrc).

```
    export variant=normal
    /bin/bash /Users/k.elwert/Code/xi-mac/build-rust-xcode.sh
/Users/k.elwert/.zshrc: line 4: typeset: -U: invalid option
typeset: usage: typeset [-afFirtx] [-p] name[=value] ...
/Users/k.elwert/.zshrc: line 13: zstyle: command not found
/Users/k.elwert/.zshrc: line 14: zstyle: command not found
/Users/k.elwert/.zshrc: line 15: zstyle: command not found
/Users/k.elwert/.zshrc: line 16: zstyle: command not found
/Users/k.elwert/.zshrc: line 17: zstyle: command not found
/Users/k.elwert/.zshrc: line 18: zstyle: command not found
/Users/k.elwert/.zshrc: line 19: zstyle: command not found
/Users/k.elwert/.zshrc: line 20: zstyle: command not found
/Users/k.elwert/.zshrc: line 21: zstyle: command not found
/Users/k.elwert/.antigen/antigen.zsh: line 2: /bin/antigen.zsh: No such file or directory
/Users/k.elwert/.zshrc: line 40: antigen: command not found
/Users/k.elwert/.zshrc: line 43: antigen: command not found
/Users/k.elwert/.zshrc: line 44: antigen: command not found
/Users/k.elwert/.zshrc: line 45: antigen: command not found
/Users/k.elwert/.zshrc: line 47: antigen: command not found
/Users/k.elwert/.zshrc: line 50: unsetopt: command not found
/Users/k.elwert/.zshrc: line 55: chruby_auto: command not found
^C** BUILD INTERRUPTED **
```

I believe this was caused by #56. Although it may fall victim to the same pitfalls, using something like `~/.zshenv` or `~/.zprofile` would be a little less risky.

cmyr:
This seems like a reasonable approach. It just works if people have done a default rustup install, and if someone needs to point us to a custom location that's relatively painless.

Fixes #58

### DIFF
--- a/build-rust-xcode.sh
+++ b/build-rust-xcode.sh
@@ -7,11 +7,14 @@ export PATH="$PATH:${HOME}/.cargo/bin"
 # Users can optionally set cargo path in xi-mac/.env
 if [[ -f "${SRCROOT}/.env" ]]; then
     source "${SRCROOT}/.env"
-    export PATH="$PATH:$CARGO_PATH"
+    if ! [[ -z "$CARGO_PATH" ]]; then
+        export PATH="$CARGO_PATH:$PATH"
+    else
+        echo "warning: ${SRCROOT}/.env file found, but CARGO_PATH not set."
+    fi
 fi
 
-
-if ! [ -x "$(command -v cargo)" ]; then
+if ! [[ -x "$(command -v cargo)" ]]; then
     echo 'error: Unable to find cargo command. If cargo is not installed visit rustup.rs, otherwise set CARGO_PATH in xi-mac/.env' >&2
     exit 127
 fi
--- a/build-rust-xcode.sh
+++ b/build-rust-xcode.sh
@@ -1,11 +1,19 @@
 #!/bin/bash
 
-if [[ -f "${HOME}/.bash_profile" ]]; then
-    source "${HOME}/.bash_profile"
+# When building from Xcode we want to ensure that `cargo` is in PATH.
+# as a convenience, add the default cargo install location
+export PATH="$PATH:${HOME}/.cargo/bin"
+
+# Users can optionally set cargo path in xi-mac/.env
+if [[ -f "${SRCROOT}/.env" ]]; then
+    source "${SRCROOT}/.env"
+    export PATH="$PATH:$CARGO_PATH"
 fi
 
-if [[ -f "${HOME}/.zshrc" ]]; then
-source "${HOME}/.zshrc"
+
+if ! [ -x "$(command -v cargo)" ]; then
+    echo 'error: Unable to find cargo command. If cargo is not installed visit rustup.rs, otherwise set CARGO_PATH in xi-mac/.env' >&2
+    exit 127
 fi
 
 set -e
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 xcuserdata/
 .DS_Store
 build/
+.env


build-rust-xcode.sh
```diff
@@ -1,11 +1,19 @@
...
-if [[ -f "${HOME}/.bash_profile" ]]; then
-    source "${HOME}/.bash_profile"
+# When building from Xcode we want to ensure that `cargo` is in PATH.
+# as a convenience, add the default cargo install location
+export PATH="$PATH:${HOME}/.cargo/bin"
+
+# Users can optionally set cargo path in xi-mac/.env
+if [[ -f "${SRCROOT}/.env" ]]; then
+    source "${SRCROOT}/.env"
+    export PATH="$PATH:$CARGO_PATH"
```
belak: Perhaps we should only make this change if CARGO_PATH is set? Additionally, it may be better to add it to the start of the path so it's resolved properly if there's another cargo binary in the path.

build-rust-xcode.sh
```diff
@@ -1,11 +1,19 @@
...
 fi
 
-if [[ -f "${HOME}/.zshrc" ]]; then
-source "${HOME}/.zshrc"
+
+if ! [ -x "$(command -v cargo)" ]; then
```
belak: I think this could be simplified to if ! which cargo &>/dev/null

Or something similar to that

build-rust-xcode.sh
cmyr: Sounds good. :)

build-rust-xcode.sh
cmyr: I avoid `which` for the reasons outlined in [this SO answer](https://stackoverflow.com/a/677212/1529922), although it's overkill here.

jansol:
LGTM

build-rust-xcode.sh
```diff
@@ -1,11 +1,22 @@
...
-if [[ -f "${HOME}/.bash_profile" ]]; then
-    source "${HOME}/.bash_profile"
+# When building from Xcode we want to ensure that `cargo` is in PATH.
+# as a convenience, add the default cargo install location
+export PATH="$PATH:${HOME}/.cargo/bin"
+
+# Users can optionally set cargo path in xi-mac/.env
+if [[ -f "${SRCROOT}/.env" ]]; then
+    source "${SRCROOT}/.env"
+    if ! [ -z "$CARGO_PATH" ]; then
```
belak: This one isn't as important and I think this code is fine, but it's usually less error prone to use [[ when possible (since there's already a requirement on bash)

build-rust-xcode.sh
```diff
@@ -1,11 +1,22 @@
...
+        export PATH="$CARGO_PATH:$PATH"
+    else
+        echo "warning: ${SRCROOT}/.env file found, but CARGO_PATH not set."
+    fi
 fi
 
-if [[ -f "${HOME}/.zshrc" ]]; then
-source "${HOME}/.zshrc"
+if ! [ -x "$(command -v cargo)" ]; then
```
belak: Same here.

dotZak:
This fix worked for me. Thanks!

cmyr:
@raphlinus this is gtg

raphlinus:
Thanks!

PR #63, opened 2017-09-22
Swift 4 migration

cmyr:
This was comparatively painless. Only bit of funny business was that I had to rejiggle how we get bytes out of the read buffer very slightly.

Oh, also: the method we were using to do the uft8 to utf16 conversions was deprecated. I think I've figured out the intended replacement, but I'm not 100% certain. I'll highlight the line in question below.


XiEditor/StyleMap.swift
```diff
@@ -100,7 +100,8 @@ struct StyleSpan {
 }
 
 func utf8_offset_to_utf16(_ s: String, _ ix: Int) -> Int {
-    return s.utf8.index(s.utf8.startIndex, offsetBy: ix).samePosition(in: s.utf16)!._offset
+    //TODO: (swift 4 conversion: verify that this works as expected)
```
cmyr: @raphlinus This is the one bit I'm not totally sure about.

raphlinus: Looks good to me, and I tested the utf8/utf16 conversion with emoji. I was thrown off slightly by the fact that gestures were broken, but that was unrelated to this change (google/xi-editor#404).

cmyr:
@raphlinus okay, rebased to remove that comment. You're ready for this to go in?

raphlinus:
Go for it!

PR #65, opened 2017-09-27
Whitelist selector commands

raphlinus:
After the RPC changes, sending an unknown command terminates the core
process, which can cause loss of work. This patch whitelists the
selectors, and also moves them to a map, so the command needn't be the
camel-case version of the selector (see google/xi-editor#159).

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -71,46 +71,6 @@ func colorFromArgb(_ argb: UInt32) -> NSColor {
         alpha: CGFloat((argb >> 24) & 0xff) * 1.0/255)
 }
 
-// Mapping of selectors to simple no-parameter commands.
-let selectorToCommand = [
-    "deleteBackward:": "delete_backward",
-    "deleteForward:": "delete_forward",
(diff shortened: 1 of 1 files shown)


raphlinus:
Note on this: it's missing moveWord{Left,Right} and the AndModifySelection variants. Have that in my local branch, will fold into the next update.

XiEditor/EditView.swift
```diff
@@ -71,22 +71,45 @@ func colorFromArgb(_ argb: UInt32) -> NSColor {
...
-        var upperCaseFragement: NSString?
-        if (scanner.scanUpToCharacters(from: notUpperCase, into: &upperCaseFragement)) {
-            underscored.append("_");
-            let downcasedFragment = upperCaseFragement!.lowercased;
-            underscored.append(downcasedFragment);
-        }
-    }
-    return underscored;
-}
+// Mapping of selectors to simple no-parameter commands.
```
cmyr: Nit: I might make this a static var in `EditView`, but that might be a hangover from a time when swiftc had performance issues around globals. If a global, I would give it an uppercase name?

PR #66, opened 2017-09-27
Add explicit @objc

raphlinus:
Swift is throwing warnings for implicit @objc, this fixes those by
adding explicit @objc annotations.


cmyr:
(famous last words)

PR #69, opened 2017-10-19
Add "moveToEndOfParagraph" to selector table

cmyr:
This is the selector fired by ctrl-e, apparently


raphlinus:
Good catch. I probably should have done a systematic review.

PR #72, opened 2017-10-23
Fix minor typos

tjrana:
Update Mac to macOS
Related to #41


cmyr:
Thanks for this, but I think we will leave it as is: 'Mac' here is referring to the platform ('the Mac') and not the operating system 'macOS'. If we were saying 'OS X' it would be another matter. 😄

PR #87, opened 2017-12-06
Use invalidation rectangle

raphlinus:
Do partial invalidation of EditView rather than redrawing the whole
screen. Should improve latency and power usage.

Also changes blinking cursor to not draw when off, rather than draw
background color.

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -114,7 +114,7 @@ class EditView: NSView, NSTextInputClient {
     private var cursorColor: NSColor {
         // using foreground instead of caret because caret looks weird in the default
         // theme, and seems to be ignored by sublime text anyway?
-        return _cursorStateOn ? dataSource.theme.foreground : dataSource.theme.background
+        return dataSource.theme.foreground
     }
 
     required init?(coder: NSCoder) {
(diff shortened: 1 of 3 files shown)


XiEditor/EditView.swift
```diff
@@ -436,7 +444,7 @@ class EditView: NSView, NSTextInputClient {
     /// timer callback to toggle the blink state
     @objc func _blinkInsertionPoint() {
         _cursorStateOn = !_cursorStateOn
-        needsDisplay = true
+        partialInvalidate(invalid: dataSource.lines.cursorInval)
```
cmyr: A minor improvement possible here: `EditView` could just have a `invalidationSetForCurrentCursor` property that is set when the update arrives, which would save recomputing on each blink.

Not a blocker, more an observation.

XiEditor/LineCache.swift
```diff
@@ -178,4 +199,32 @@ class LineCache {
...
+                inval.addRange(start: i + nInvalidBefore, n: 1)
+            }
+        }
+        return inval
+    }
+}
+
+/// A set of line numbers to be invalidated, in run-length representation
+class InvalSet {
+    var ranges: [(Int, Int)] = []
```
cmyr: Non-block: I'd be inclined to use either the `Range` or `ClosedRange` foundation type here.

XiEditor/EditView.swift
raphlinus: I agree this is a good optimization, but I expect the impact to be tiny.

raphlinus:
I also changed it so InvalSet is ranges. Agree this is nicer, a pair always has ambiguity whether it's (start, count) or (start, end).

XiEditor/EditView.swift
cmyr: agreed, just an observation. :)

PR #91, opened 2017-12-18
Use lock instead of queue for LineCache synchronization (speculative) 

cmyr:
This replaces the dispatch queue with a lock. The resulting code is perhaps easier to reason about and is slightly faster (though this is unlikely to impact user experience) at the cost of being less idiomatic. I did this mostly as an exercise.

@raphlinus if you prefer this then let's go with it, otherwise let's not.

### DIFF
--- a/XiEditor/LineCache.swift
+++ b/XiEditor/LineCache.swift
@@ -208,10 +208,10 @@ class LineCache {
 
     /// A semaphore we use to wake up the main thread if it is blocking missing lines
     fileprivate let waitingForLines = DispatchSemaphore(value: 0)
-    /// A dispatch queue used to synchronize access to the underlying state
-    fileprivate let queue = DispatchQueue(label: "io.xi-editor.cache-sync-queue")
-    /// A marker used to indicate if the main thread is blocked.
-    fileprivate let syncLock = Mutex()
+    /// A lock for synchronizing semaphore waits
+    fileprivate let syncLock = UnfairLock()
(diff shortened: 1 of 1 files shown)


raphlinus:
I'm torn on this; it is indeed less idiomatic, and the performance impact should be unmeasurably small. It's also very likely that a future version of Swift will have actual concurrency primitives and whatever we do should be migrated to that.

That said, I'm considering this approach because I'd like the render method in the textplane branch to be able to lock the line cache for mutable access, so it can store `TextPlane` objects in it. The easiest thing to do is to explicitly lock and unlock, which then becomes a correctness requirement for the client to respect. Another thing I'm considering is having an object analogous to `MutexGuard` in Rust, that holds the mutex and unlocks it on `deinit`. Both of these approaches work better by decoupling the lock from being able to run a block.

Long story short, I'll _probably_ merge this, then do my mutable-access changes on top of it, but I'll wait til then to decide for sure.

cmyr:
@raphlinus there's some good discussion about idiomatic locks/queues in swift at [Friday Q&A](https://www.mikeash.com/pyblog/friday-qa-2017-10-27-locks-thread-safety-and-swift-2017-edition.html) that might be worth looking at. Agree that this will have to be revisted once swift has a coherent concurrency story, but that looks like it's a ways away yet.

raphlinus:
Merging this, as we discussed on IRC. I'll update #97 on top of this.

PR #94, opened 2017-12-27
Add XiClient protocol, parse RPCs in read thread

cmyr:
A little bit of follow-up from the feedback to #89. This makes the client protocol an explicit protocol, and moves basic RPC parsing out of the AppDelegate.

A future addition might be to move more of the client logic out of the AppDelegate and into some new class; and perhaps also a complimentary `XiServer` protocol that the core connection can implement, which would let us get rid of the strange-stepchild `Dispatcher` class.


cmyr:
@raphlinus this is updated to accommodate #99, ready when you are.

XiEditor/CoreConnection.swift
```diff
@@ -150,29 +146,82 @@ class CoreConnection {
...
+            self.handleNotification(json: obj)
+        }
+    }
+
+    func handleRequest(json: [String: AnyObject]) {
+        // there are currently no core -> client requests in the protocol
+        print("Unexpected RPC Request: \(json)")
+    }
+
+    func handleNotification(json: [String: AnyObject]) {
```
raphlinus: I think at some point I'd like to see the low-level json-rpc stuff split out from the specific methods handled, but for now that's very much yagni, as there's only one interface the client needs to implement directly.

XiEditor/AppDelegate.swift
```diff
@@ -262,6 +203,21 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         }
     }
 
+    //MARK: helpers
+
+    /// returns the NSDocument corresponding to the given viewIdentifier
+    private func documentForViewIdentifier(viewIdentifier: ViewIdentifier) -> Document? {
+        //TODO: move this to use a hashmap; we will have to do some manual bookkeeping,
+        // but this is a noticeable (small) bottleneck in some scenarios
```
raphlinus: I wouldn't consider it small necessarily, but no need to quibble. On the other hand, I think it _is_ good to point out that this method needs to be thread-safe (as it's called on the update method from the read thread), even if the current impl isn't doing anything special to respect that.

XiEditor/AppDelegate.swift
```diff
@@ -107,147 +99,96 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
-    /// returns the NSDocument corresponding to the given viewIdentifier
-    private func documentForViewIdentifier(viewIdentifier: ViewIdentifier) -> Document? {
-        for doc in NSApplication.shared.orderedDocuments {
-            guard let doc = doc as? Document else { continue }
-            if doc.coreViewIdentifier == viewIdentifier {
-                return doc
-            }
-        }
-        return nil
+    // MARK: XiClient protocol
```
raphlinus: We've been using the `// MARK: -` syntax, which I think is appropriate here, but that's the tiniest of nits.

cmyr: Good nit.

An annoyance of mine is that `"//MARK: - "` as a normal comment adds an underline + a heading to the (little-used I suppose) file navigator in Xcode; but in doc comments for special sections you need the dash first, e.g. `"/// - Note: This function blah blah blah"`, in order for the inline documentation to render properly.

(after some playing around) actually after playing around: apparently you can have the dash before _or_ after the mark, and get a line break in the navigator:

screen shot 2018-01-04 at 6 21 47 pm

(apologies if everyone knows this but me)

XiEditor/CoreConnection.swift
cmyr: I.e. you mean that the parsing of _specific_ methods should live outside of core connection? I agree that would make sense.

raphlinus: Yes, that's what I meant.

PR #97, opened 2017-12-28
Change locking strategy for LineCache

raphlinus:
We want to allow the LineCache to be locked mutably for an extended
period so we can store associated data (TextLine entries and probably
caret positions). This patch introduces a MutexGuard very similar to
Rust's, and also a LineCacheLocked that is an instance, so mutations
to the state are safely guarded by the mutex.

### DIFF
--- a/XiEditor/LineCache.swift
+++ b/XiEditor/LineCache.swift
@@ -59,7 +59,7 @@ struct Line {
 }
 
 /// The underlying state of the cache, with methods for applying update deltas.
-class LineCacheState: UnfairLock {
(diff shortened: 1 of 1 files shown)


raphlinus:
This feels like a nice local change to the new locking strategy. I'll implement the TextLine caching on top of this, which will also have the effect of migrating method calls currently on LineCache to calls on LineCacheLocked (which will also decrease the amount of forwarding bureaucracy).

Note I got rid of the separate MutexGuard class. This also makes the permissions much cleaner, as LineCacheState is now fileprivate.

PR #98, opened 2017-12-28
Reenable live resize

raphlinus:
Also fixes scroll coordinates on redrawEverything()

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -134,6 +134,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     
     @objc func frameDidChangeNotification(_ notification: Notification) {
         updateEditViewHeight()
+        willScroll(to: scrollView.contentView.bounds.origin)
     }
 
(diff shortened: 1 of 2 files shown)


XiEditor/EditViewController.swift
```diff
@@ -156,7 +156,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     func redrawEverything() {
         updateGutterWidth()
         updateEditViewHeight()
-        willScroll(to: scrollView.contentView.bounds.origin)
```
cmyr: slightly confused, these should be equivalent?

raphlinus: I think you're right, I made this change in response to other problems I was seeing. I have a feeling there are still problems where the core's viewport (as notified by scroll) is out of sync with the client's, but this isn't it...

cmyr: hmm. I noticed that how we calculate viewport size in the view controller differs from how we do it in the draw call (the view controller doesn't take account of certain offsets?) could be related.

PR #100, opened 2017-12-29
Store TextLine objects in line cache

raphlinus:
This patch retains `TextLine` objects and stores them in the line cache, avoiding the need to rebuild them continually. With this change, we see consistent 120fps and pleasantly low CPU usage on scrolling. In addition, it should reduce the latency for operations such as inserting or deleting a line (most of the surrounding content is retained).

Note that the text for the line number in the gutter is also cached.

The associated data in the line cache is generic. This isn't strictly needed as there's only one instantiation, but perhaps helps highlight the separation of concerns. It's possible we'll use associated data for caret offsets (although that could also be moved into TextLine).

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -404,7 +404,7 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
     }
 
     func getLine(_ lineNum: Int) -> Line<LineAssoc>? {
-        return dataSource.lines.get(lineNum)
+        return dataSource.lines.locked().get(lineNum)
     }
 
     func partialInvalidate(invalid: InvalSet) {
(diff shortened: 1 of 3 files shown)


cmyr:
An observation on this: I suspect we're taking a performance _hit_ in the case where highlighting is in progress. I notice more misses in a debug build when first loading a file here than I do without the cache. Does that seem credible?

I'd like to profile more but I'm running out of time for the day. May have more this evening, else the weekend...

raphlinus:
It seems credible. My hunch is that this patch really is improving performance, but that may be putting more pressure on other parts of the pipeline. I think it's important to understand what's going on, so will be putting more than a little focus into tracing infra, especially generating traces in [chrome tracing format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview).

PR #101, opened 2017-12-29
Improve semaphore dynamics

raphlinus:
This patch handles races between semaphore timeout and the read thread better, and also is more thorough about waiting for all lines to be present when there's a pipeline of updates.


raphlinus:
Ok reverted the second commit.

PR #103, opened 2018-01-02
Generate simple tracing data

raphlinus:
Adds a "write trace" to the debug menu, which dumps a trace file in
Chrome trace (catapult) format to /tmp. Right now we trace the basics
of sending and receiving RPC's, as well as the render method in
EditView (this is basically the critical path for editing and
scrolling).

### DIFF
--- a/XiEditor/Trace.swift
+++ b/XiEditor/Trace.swift
@@ -17,13 +17,15 @@
 
 import Foundation
 
-var globalTrace = Trace()
-
+/// Collect trace events so they can be output in Chrome tracing format.
 class Trace {
     let mutex = UnfairLock()
     let BUF_SIZE = 100_000
     var buf: [TraceEntry]
     var n_entries = 0
+
+    /// Shared instance, most uses should call this.
+    static var shared = Trace()
 
     init() {
         buf = [TraceEntry](repeating: TraceEntry(), count: BUF_SIZE)
@@ -45,7 +47,7 @@ class Trace {
     func write() {
         let pid = getpid()
         let path = "/tmp/xi-trace-\(pid)"
-        if !FileManager().createFile(atPath: path, contents: nil, attributes: nil) {
+        if !FileManager.default.createFile(atPath: path, contents: nil, attributes: nil) {
             print("error creating trace file")
             return
         }
--- a/XiEditor/Trace.swift
+++ b/XiEditor/Trace.swift
@@ -1,0 +1,96 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// A simple mechanism for logging trace events and outputting a file in
+// Chrome tracing format
+
+import Foundation
+
+var globalTrace = Trace()
+
+class Trace {
+    let mutex = UnfairLock()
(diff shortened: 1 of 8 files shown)


XiEditor/Trace.swift
```diff
@@ -0,0 +1,96 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// A simple mechanism for logging trace events and outputting a file in
+// Chrome tracing format
+
+import Foundation
+
+var globalTrace = Trace()
```
cmyr: Swift style[1] would be to the global be a static type property, ala `NSApplication.shared` or `FileManager.default`. So perhaps `Trace.default`?

A partial solution to your TODO, below, would be for `Trace` to be init'd with a path (maybe just a directory) where it will be saved. 

1. [Adopting Cocoa Design Patterns](https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html) (search 'singleton')

XiEditor/Trace.swift
```diff
@@ -0,0 +1,96 @@
...
+
+class Trace {
+    let mutex = UnfairLock()
+    let BUF_SIZE = 100_000
+    var buf: [TraceEntry]
```
cmyr: This might make more sense as a `TraceEntry?`, which would mean we wouldn't need to have the empty init method on `TraceEntry` that just creates a dummy entry; and it would also clean up the logic in `write` where you're doing bounds checking against the number of entries.

XiEditor/Trace.swift
```diff
@@ -0,0 +1,96 @@
...
+    var n_entries = 0
+
+    init() {
+        buf = [TraceEntry](repeating: TraceEntry(), count: BUF_SIZE)
```
cmyr: Musing: This could be set at declare time if we're willing to make it lazy, or if we make BUF_SIZE a type property; then we wouldn't need the init. (Although we might want one for other reasons... ) 🤷‍♂️

XiEditor/Trace.swift
```diff
@@ -0,0 +1,96 @@
...
+        pthread_threadid_np(nil, &buf[i].tid)
+        n_entries += 1
+        mutex.unlock()
+    }
+
+    // TODO: more control over where this gets saved
+    func write() {
+        let pid = getpid()
+        let path = "/tmp/xi-trace-\(pid)"
+        if !FileManager().createFile(atPath: path, contents: nil, attributes: nil) {
```
cmyr: Normally there would be no need to init a new FileManager here; instead you would use the shared `FileManager.default` instance.

XiEditor/Trace.swift
```diff
@@ -0,0 +1,96 @@
...
+    let mutex = UnfairLock()
+    let BUF_SIZE = 100_000
+    var buf: [TraceEntry]
+    var n_entries = 0
+
+    init() {
+        buf = [TraceEntry](repeating: TraceEntry(), count: BUF_SIZE)
+    }
+
+    func trace(_ name: String, _ cat: TraceCategory, _ ph: TracePhase) {
```
cmyr: Oddly, one of my personal gripes with rust is the _lack_ of named arguments. 😕

XiEditor/Trace.swift
raphlinus: I went with `shared`, seems appropriate.

XiEditor/Trace.swift
raphlinus: It doesn't clean up `write`, as that's doing circular buffer logic and not simply bounds-checking.

XiEditor/Trace.swift
raphlinus: You can see I used named arguments extensively in the textplane work. Here I decided against because (a) I want conciseness and low noise, and (b) the fact that the arguments have bespoke types reduces potential confusion.

raphlinus:
I just measured the overhead at 13ns per trace point (in a tight loop), which is small enough we do not need to worry about it. Thanks for the review!

PR #104, opened 2018-01-05
Use config for scrollPastEnd.

cmyr:
requires https://github.com/google/xi-editor/pull/465.


XiEditor/EditViewController.swift
```diff
@@ -109,8 +109,13 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     // visible scroll region
     var visibleLines: LineRange = 0..<0
 
-    // TODO: should be an option in the user preferences
-    var scrollPastEnd = false
+    var scrollPastEnd = false {
+        didSet {
+            if scrollPastEnd != oldValue {
+                redrawEverything()
```
raphlinus: I think this is fine, but technically it probably only requires `updateEditViewHeight`

PR #105, opened 2018-01-06
Improve new window placement

cmyr:
This is a little improvement to how new windows are positioned:

1. For each application run, we keep track of the last new window position/size; new windows are positioned offset from this, so they cascade. They wrap to the top or side of the screen if their new position would clip.

2. When the user manually moves or positions a view at any point, we save the position of that view to be the 'new view position' on the next launch.

This was based on me copying the behaviour of Terminal.app. I'm not sure I find this behaviour fully satisfying, (I could imagine new windows being created offset from the current window, for instance) but I think it's a clear improvement over the current mess.

### DIFF
--- a/XiEditor/Document.swift
+++ b/XiEditor/Document.swift
@@ -37,6 +37,15 @@ class Document: NSDocument {
     /// if set, should be used as the tabbingIdentifier of new documents' windows.
     static var preferredTabbingIdentifier: String?
 
+    /// Used to determine initial locations for new windows.
+    fileprivate static var _lastWindowFrame: NSRect = {
+        if let saved = UserDefaults.standard.string(forKey: USER_DEFAULTS_NEW_WINDOW_FRAME) {
+            return NSRectFromString(saved)
+        } else {
+            return NSRect(x: 200, y: 200, width: 600, height: 600)
+        }
+    }()
+    
     var dispatcher: Dispatcher!
     /// coreViewIdentifier is the name used to identify this document when communicating with the Core.
     var coreViewIdentifier: ViewIdentifier? {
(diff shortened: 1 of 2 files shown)


XiEditor/Document.swift
```diff
@@ -224,4 +240,33 @@ class Document: NSDocument {
...
+    func frameForNewWindow() -> NSRect {
+        let offsetSize: CGFloat = 22
+        let screenBounds = NSScreen.main!.visibleFrame
+        var nextFrame = NSOffsetRect(Document._lastWindowFrame, offsetSize, -offsetSize)
+        if nextFrame.maxX > screenBounds.maxX {
+            nextFrame.origin.x = screenBounds.minX + offsetSize
+        }
+        if nextFrame.minY <= 0 {
+            nextFrame.origin.y = screenBounds.maxY - nextFrame.height
+        }
```
jansol: Might want to ensure that the new frame is not offscreen in the other direction either.

cmyr: A user can't drag a window above the menu bar unless they're dragging it to a separate view. Since we only move down and to the right, and wrap around to an edge when necessary, it shouldn't be possible for us to position a view out of bounds; the only exception would be if the user manually positions a view out so that it is clipped by the left edge of the screen before closing the app. In this case new windows will still progressively be moving into view.

So yea, maybe a manual check on first load that makes sure the left edge of the window is in bounds would make sense. :)

XiEditor/Document.swift
```diff
@@ -224,4 +240,33 @@ class Document: NSDocument {
             editVC.updateAsync(update: update)
         }
     }
+    
+    /// Returns the frame to be used for the next new window.
+    /// - Note: This attempts to replicate the behaviour of native macOS applications.
+    /// On launch, an initial location is chosen, generally based on the position of
+    /// the last manually moved view during the application's last run; subsequent
+    /// windows are offset from this by approximately thie width of the title bar.
+    /// If a window clips the screen, the position starts again from the beggining
```
raphlinus: tiniest spelling nits: "the", "beginning"

PR #106, opened 2018-01-09
Faster view resolution

cmyr:
We were going to eventually need a custom NSDocumentController subclass for things like multiview, so this seems like as good a time as any.

This isn't extensively tested, but I think the approach is solid, and I'll be using this build; will see if anything shakes out.

### DIFF
--- a/XiEditor/XiDocumentController.swift
+++ b/XiEditor/XiDocumentController.swift
@@ -1,0 +1,46 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


XiEditor/XiDocumentController.swift
```diff
@@ -0,0 +1,51 @@
...
+    func documentForViewIdentifier(_ viewIdentifier: ViewIdentifier) -> Document? {
+        lock.lock()
+        defer { lock.unlock() }
+        return openViews[viewIdentifier]
+    }
+    
+    /// Associates the given view with the given identifier. Should only be called when
+    /// the `ViewIdentifier` is first set.
+    func setIdentifier(_ viewIdentifier: ViewIdentifier, forDocument document: Document) {
+        lock.lock()
```
raphlinus: Any special reason you're doing explicit lock/unlock here but used defer above?

cmyr: I use defer above because it lets me return without using a temporary value.

PR #109, opened 2018-01-10
Refactor document controller

cmyr:
Another little batch of cleanup, moving things into the new document controller that probably belong there.


cmyr:
Noted. If at some point it makes sense to clean that up and go 10.12+ I'll go ahead.

PR #114, opened 2018-01-13
Use mach_timebase_info to convert time units

raphlinus:
The previous code assumed that mach_absolute_time returned ns, which
is not very solid. This patch corrects using mach_timebase_info.

### DIFF
--- a/XiEditor/Trace.swift
+++ b/XiEditor/Trace.swift
@@ -33,7 +33,6 @@ class Trace {
         buf = [TraceEntry](repeating: TraceEntry(), count: BUF_SIZE)
         var info = mach_timebase_info(numer: 0, denom: 0)
         mach_timebase_info(&info)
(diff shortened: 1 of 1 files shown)


XiEditor/Trace.swift
```diff
@@ -23,12 +23,20 @@ class Trace {
...
+    let mach_time_denom: UInt64
 
     /// Shared instance, most uses should call this.
     static var shared = Trace()
 
     init() {
         buf = [TraceEntry](repeating: TraceEntry(), count: BUF_SIZE)
+        var info = mach_timebase_info(numer: 0, denom: 0)
+        mach_timebase_info(&info)
+        print(info.numer, info.denom)
```
vlovich: left-over debugging print?

raphlinus: good catch, thanks

PR #118, opened 2018-01-23
Add bounce to edit view

### LINKED ISSUE #107: Scrolling beyond the ends of a file is differen from TextEdit (trackpad)
In TextEdit, scrolling with the trackpad is possible even when the content fits entirely in the window or is already scrolled all the way to either end. It will overscroll and then bounce back.

Xi-mac only overscrolls when hitting the start or end of the file while scrolling. When all content fits in the window or is already scrolled to the start/end it will not scroll past the end at all. Continuing scrolling at slow speed also stops moving the view when hitting the edge of the view contents.

vlovich:
This maintains consistency with the behaviour of other text editing applications on OX.
Fixes #107

### DIFF
--- a/XiEditor/Main.storyboard
+++ b/XiEditor/Main.storyboard
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="13529" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
+<document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="13771" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="13529"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="13771"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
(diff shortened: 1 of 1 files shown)


raphlinus:
It addresses the issue, but I'm not yet 100% convinced this should be the behavior. Sublime and VS Code and Atom all agree on the existing behavior, while TextEdit and Xcode have this behavior.

The Linux way of course would be to make it an option (and not just a bool but an enum with several more choices, including enabling it only when the document is larger than the viewport). I'm interested to hear from people what they prefer.

vlovich:
My personal preference is to make it feel like a native app.  I believe the other apps are all cross-platform which is why they struggle to do this there's an open ask for each:

https://github.com/atom/atom/issues/345
https://github.com/Microsoft/vscode/issues/14823
https://forum.sublimetext.com/t/elastic-scrolling/4504

All the explanations given are that it's hard to get the native effect for them & there doesn't seem to be any comments saying it's undesirable, although it's quite possible the silent majority is happy without it/don't care enough to file a bug.

As for preferences, I agree 100% we should make things configurable (at least via the current TOML mechanism - UI preferences should probably hide crazy stuff by default to be user friendly) and I'm happy to add that here if you wish.  The only if larger than viewport option *may* be more difficult to implement, so I'd punt on that initially until someone asks for it.

The one question I would have about preferences is whether or not they're actually hooked up.  I see places where it looks like they might be but they look to be all over the place & it's unclear if they actually change behaviour.

raphlinus:
Yeah, that's a convincing argument, and it's definitely consistent with the mission to feel native rather than cross-platform.

Let's give it a few days and see if we hear strong preferences against, and if not we can just merge this.

Thanks!

cmyr:
I haven't built this patch, but I'm firmly on board with replicating the native behaviour. That sublime and vscode don't speaks more to those not being native cocoa applications than anything else.

I would be curious to hear an 'anti' position, but I am strongly inclined to merge anything that makes us behave more like a normal scrollview.

PR #119, opened 2018-01-23
Remove gutter from line cache

### LINKED ISSUE #115: empty newlines can screw up gutter line numbers
Entering newlines fails to update the line numbers of subsequent lines.

screen shot 2018-01-18 at 3 16 35 pm

vlovich:
The gutter invalidation logic is different from regular lines.  It
should only be invalidated when a newline is inserted & that newline
insertion invalidates all following lines.  This is in contrast to other
things using the line cache which only invalidate on line changes &
never newlines.  In fact, the gutter is invalidated whenever the cursor
changes lines too.

This fixes the bug where adding a newline in the middle of a document
has an off-by-1 error in the gutter line count.

Fixes #115

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -472,12 +472,7 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
                 styleMap.applyStyles(builder: builder, styles: line.styles)
                 textLine = builder.build(fontCache: renderer.fontCache)
 
-                let gutterText = "\(lineIx + 1)"
-                let gBuilder = TextLineBuilder(gutterText, font: font)
-                gBuilder.setFgColor(argb: line.containsCursor ? foregroundArgb: gutterArgb)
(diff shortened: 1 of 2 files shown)


raphlinus:
Riiight. Now that you point it out, it's clear that just being associated data cannot work, as the validity is not maintained by insert/delete operations. I also see that this patch fixes it, but at the same time I'm concerned about the performance hit; as I understand it, we're going to be doing the TextLine.build operations for every line while scrolling.

I just measured this, and it seems to take a hot-cache paint from about 0.9ms to 1.3ms. So noticeable regression but not a killer.

vlovich:
Right.  I've adopted your scatter gather suggestion from IRC in #121 which builds on top of this CL to add back a cache to the gutter that's more correct in that it's still recomputing the gutter numbers live just without the expensive part.  In my benchmarks it's only ~15-30% slower.  On my testing it went from 0.682ms -> ~1.638 w/o cache -> 0.785ms w/ scatter gather.

PR #121, opened 2018-01-27
Optimize gutter cache

### LINKED ISSUE #115: empty newlines can screw up gutter line numbers

vlovich:
Build the gutter digits for foreground & background.  Then use scatter-gather on the appropriate cached TextLine to quickly create a TextLine for the necessary line number.  This gets us back within range of the previous cached gutter performance with the same correctness as without the cache.

Fixes #115

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -39,7 +39,7 @@ struct TextDrawingMetrics {
     var baseline: CGFloat
     var linespace: CGFloat
     var fontWidth: CGFloat
-    
+
     init(font: NSFont, textColor: NSColor) {
         self.font = font
         ascent = font.ascender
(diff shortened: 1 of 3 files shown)


XiEditor/XiTextPlane/TextLine.swift
```diff
@@ -168,6 +168,24 @@ struct TextLine {
...
+            -> TextLine where C.Element == Int, C.IndexDistance == Int {
+        var newGlyphs = Array<GlyphInstance>()
+        newGlyphs.reserveCapacity(glyphIndices.count)
+
+        var glyphPosition = 0
+        for i in glyphIndices {
+            newGlyphs.append(self.glyphs[i])
+            newGlyphs[glyphPosition].x = self.glyphs[glyphPosition].x
+            newGlyphs[glyphPosition
+                ].y = self.glyphs[glyphPosition].y
```
cmyr: Not sure this line break was intentional?

XiEditor/EditView.swift
```diff
@@ -56,6 +56,45 @@ struct TextDrawingMetrics {
...
+/// The pre-renderered cache of the digits 0-9 for display in the gutter.
+/// These can be scatter-gathered to quickly construct any sequence of digits.
+struct GutterCache {
+    private let font: CTFont
+    // Scatter-gather from here for the gutter for lines not containing a cursor.
+    private let noCursorCache: TextLine
+    // Scatter-gather from here for the gutter for lines containing a cursor.
+    private let cursorCache: TextLine
+
+    // Initializes the cache with the given font & the requested colors for when
```
cmyr: `///` here would make this a doc-comment, rendered on hover in Xcode. I regularly find myself checking this reference: http://nshipster.com/swift-documentation/

XiEditor/XiTextPlane/TextLine.swift
```diff
@@ -168,6 +168,24 @@ struct TextLine {
...
+            glyphPosition += 1
+        }
+        let newWidth = self.width * (Double(glyphIndices.count)) / (Double(self.glyphs.count))
+        let newLine = CTLineCreateTruncatedLine(ctLine, newWidth, CTLineTruncationType.start, nil)!
```
cmyr: This unwrap is potentially trouble; there are legitimate instances where this function may return nil. As an example,  this will currently crash if the font is "Hoefler Text". (found by absolute accident)

This might be revealing some of the assumptions we're making around monospaced type; I do think it's reasonable that the gutter if visible always uses a mono typeface, but this isn't enforced anywhere. (Also this doesn't fail on _all_ fonts with variable-width numerals).

XiEditor/EditView.swift
vlovich: Fixed.

XiEditor/XiTextPlane/TextLine.swift
```diff
@@ -168,6 +168,32 @@ struct TextLine {
     var width: Double {
         return CTLineGetTypographicBounds(ctLine, nil, nil, nil)
     }
+
+    /// Fast approach to build a new string out of an atlast of glyphs.  This
```
cmyr: s/atlast/atlas

vlovich: Done

PR #129, opened 2018-02-01
Capture FPS statistics

vlovich:
Measure how frequently we're rendering.  Do to the way CoreAnimation
works we'll only be able to capture FPS up to VSync since we'll never
get called faster than that to draw (the draw itself doesn't actually
capture render time since that's deferred).

Also remove the blocking lines print spew in release builds & switch to 
DispatchTime instead of mach_absolute_time for timing that part of the code.

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -43,6 +43,39 @@ extension NSValueTransformerName {
     static let boolToControlStateValueTransformerName = NSValueTransformerName(rawValue: "BoolToControlStateValueTransformer")
 }
 
+class ScrollTester {
+    private let timer : Timer?
+
+    init(_ document: Document) {
+        if #available(OSX 10.12, *) {
+            // 200 hz timer
+            var line = Int(0)
+            var direction = 1
+            let scrollSize = 50
+
+            timer = Timer.scheduledTimer(withTimeInterval: 0.005, repeats: true, block: { (_) in
+                let height = document.editViewController?.lines.height ?? 0
+
+                line += scrollSize * direction
+                if line >= height {
+                    line = height
+                    direction = -1
+                } else if line < 0 {
+                    line = 0
+                    direction = 1
+                }
+                document.editViewController?.scrollTo(line, 0)
+            });
+        } else {
+            timer = nil
+        }
+    }
+
+    deinit {
+        self.timer?.invalidate()
+    }
+}
+
 @NSApplicationMain
 class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
 
@@ -305,6 +338,22 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         testWindow = NSWindow(contentRect: frame, styleMask: [.titled, .closable, .resizable, .miniaturizable], backing: .buffered, defer: false)
         testWindow?.makeKeyAndOrderFront(self)
         testWindow?.contentView = TextPlaneDemo(frame: frame)
+    }
+
+    var scrollTesters = [Document: ScrollTester]()
+    @IBAction func toggleScrollBenchmark(_ sender: AnyObject) {
+        var maybeTopmostDocument : Document?
+        for doc in NSApplication.shared.orderedDocuments {
+            guard let doc = doc as? Document else { continue }
+            maybeTopmostDocument = doc
+        }
+        guard let topmostDocument = maybeTopmostDocument else { return }
+
+        if scrollTesters.keys.contains(topmostDocument) {
+            scrollTesters.removeValue(forKey: topmostDocument)
+        } else {
+            scrollTesters.updateValue(ScrollTester(topmostDocument), forKey: topmostDocument)
+        }
     }
     
     func updateRpcTracingConfig(_ enabled: Bool) {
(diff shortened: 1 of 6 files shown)


XiEditor/AppDelegate.swift
```diff
@@ -250,6 +283,22 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         testWindow?.contentView = TextPlaneDemo(frame: frame)
     }
 
+    var scrollTesters = [Document: ScrollTester]()
+    @IBAction func toggleScrollBenchmark(_ sender: AnyObject) {
```
cmyr: If this method lived in `EditViewController` you wouldn't have to do the dance of finding the correct document. You also wouldn't need the map, you could just have a `ScrollTester?` member variable in the view controller. (I'm also wary that using `Document`s as keys here could cause them to be retained if they're closed while this is running, which isn't a huge concern but could _possibly_ cause some debugging headache down the road)

XiEditor/LineCache.swift
```diff
@@ -286,19 +286,21 @@ class LineCacheLocked<T> {
...
         if !missingLines.isEmpty {
             // TODO: should we send request to core?
+#if DEBUG
             print("waiting for lines: (\(missingLines.first!), \(missingLines.last!))")
+#endif
             //TODO: this timing + printing code can come out
             // when we're comfortable with the performance and
             // the timeout duration
-            let blockTime = mach_absolute_time()
+            let blockTime = DispatchTime.now()
```
cmyr: Should I be preferring this to `mach_absolute_time()` in general?

XiEditor/Fps.swift
```diff
@@ -0,0 +1,179 @@
...
+    public func meanFps() -> Double {
+        let totalNanoseconds = samples.reduce(UInt64(0)) { (elapsedNanosTotal, measurement) -> UInt64 in
+            return elapsedNanosTotal + measurement.elapsedNanoseconds()
+        }
+        let meanNanosecondsPerFrame = totalNanoseconds / UInt64(samples.count)
+        return FpsMeasurement.fps(nanosecondsPerFrame: meanNanosecondsPerFrame)
+    }
+}
+
+public protocol FpsObserver : class {
```
cmyr: Do we need both of these? I saw only one was actually implemented. Or maybe the fps could be an optional field that we only include when there's a tick?

XiEditor/AppDelegate.swift
cmyr: This would also potentially fix an issue I was having where the menu item wouldn't seem to do anything, or would start scrolling a background window.

XiEditor/XiTextPlane/TextPlane.swift
```diff
@@ -116,10 +118,29 @@ class TextPlaneLayer : NSOpenGLLayer {
...
         return NSOpenGLPixelFormat(attributes: attr)!.cglPixelFormatObj!
-        
+    }
+
+    func changed(fps: Double) {
+        // print("Fps \(fps), ms/frame = \(1000.0 / fps)")
+    }
+
+    func changed(fpsStats stats: FpsSnapshot) {
+        print("Fps mean: \(stats.meanFps()), 99%: \(stats.fps(percentile: 0.01)), min: \(stats.minFps()), max: \(stats.maxFps())")
```
cmyr: maybe we also want to put this behind an ifdef?

XiEditor/LineCache.swift
vlovich: Yes.

XiEditor/Fps.swift
vlovich: I wanted to do it as an overlay view but I'm a bit lazy to hook that part up.    You did get that the changed with the raw frame rate happens every tick but the FpsSnapshot is a longer time period (i.e. the past 1 second).

XiEditor/XiTextPlane/TextPlane.swift
vlovich: I'll hide it behind a debug "show fps" option.

XiEditor/AppDelegate.swift
vlovich: if I make the IBAction in EditViewController would Cocoa document magic automatically know the right document to talk to?

XiEditor/XiTextPlane/TextPlane.swift
cmyr: Elsewhere you'd used an `#if DEBUG` to gate a print statement, I thought that was a good idea, and seems equally suited here. Your idea makes sense too, but it might be confusing to actual users, since it won't show anything unless you run from xcode (at least until we wire up an overlay)

XiEditor/AppDelegate.swift
cmyr: ah I hadn't seen this. Yes: the foremost controller handles the selector, and it has a reference to the document.

XiEditor/XiTextPlane/TextPlane.swift
vlovich: I think #if DEBUG is too annoying since most developers won't care about FPS most of the time & it's pretty spammy.  Besides a UI label is nicer than console prints most of the time.

raphlinus:
Needs some merging, but otherwise looks good. Thanks!

PR #131, opened 2018-02-01
More rational view reuse, fewer redundant redraws

### LINKED ISSUE #130: When opening preferences via `CMD+,`, the entire page doesn't load at first.
Just exploring this repo and xi-editor, going to get my feet wet soon. I got it open and figured I'd try out the preferences panel, and this is what happened.

xi-mac bug

Only the first few lines are displayed, and then the rest is filled in as I scroll.

If, however, I type into the document and then hit `CMD+,` it opens it in a new frame which works just fine. So this is only when it opts to open it in the existing frame.

cmyr:
There was a small bug where we were incorrectly comparing new font names
to old ones, which caused us to unnecessarily redraw all views whenever
a new view was opened or a config changed.

We were also apparently relying on this behaviour to redraw the content
of redraw views, which we now do explicitly.

- fixes #130

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -225,7 +225,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     }
 
     func handleFontChange(fontName: String?, fontSize: CGFloat?) {
-        guard textMetrics.font.fontName != fontName || textMetrics.font.pointSize != fontSize else { return }
+        guard (textMetrics.font.fontName != fontName && textMetrics.font.familyName != fontName)
+            || textMetrics.font.pointSize != fontSize else { return }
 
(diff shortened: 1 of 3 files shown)


vlovich:
Will this cause it to fail to update the font if it's changed from the postscript name to a family name?  It seems like changing from Helvetica-Regular -> Helvetica would bypass the redraw when it shouldn't since the fontFamily stays Helvetica...

cmyr:
@vlovich I don't believe 'Helvetica-Regular' is a valid postscript name. The postscript name of 'Helvetica Regular' (just looking in fontbook) is 'Helvetica'.

Changing from 'Helvetica-Light' to 'Helvetica-Oblique' or to 'Helvetica' causes a redraw. I had to double check the logic again, but I think it's correct?

PR #132, opened 2018-02-02
gitignoring the xi-editor folder

zachdaniel:
Seems like it will be pretty common to be managing a xi-editor folder inside of xi-mac, so gitignoring it should be useful.


zachdaniel:
I signed it!

raphlinus:
Thanks! This does show up in my `git status` so will be an improvement.

PR #137, opened 2018-02-06
Fix crash on certain fonts

vlovich:
This is kind of a hack that ensures that any font will still render.  My guess is that it's  a bug in CTLineCreateTruncatedLine where the width maybe gets converted to single precision floating point?.

Fixes #135

Issue #135, opened 2018-02-05
Instant crash when using Fira Mono


jansol:
Setting the font to Fira Mono causes xi-mac to crash instantly upon saving the config file, as well as on subsequent launches.

Stacktrace:
```
Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.levien.XiEditor           	0x000000010cc82705 specialized TextLine.scatterGather<A>(indices:font:) + 853 (TextLine.swift:194)
1   com.levien.XiEditor           	0x000000010ccb368c GutterCache.lookupLineNumber(lineIdx:hasCursor:) + 396 (EditView.swift:94)
2   com.levien.XiEditor           	0x000000010ccc0af4 specialized EditView.render(_:dirtyRect:) + 15876 (EditView.swift:599)
3   com.levien.XiEditor           	0x000000010ccb72c4 protocol witness for TextPlaneDelegate.render(_:dirtyRect:) in conformance EditView + 68
4   com.levien.XiEditor           	0x000000010cc8b513 specialized TextPlaneLayer.draw(in:pixelFormat:forLayerTime:displayTime:) + 531
5   com.levien.XiEditor           	0x000000010cc8a591 @objc TextPlaneLayer.draw(in:pixelFormat:forLayerTime:displayTime:) + 49
6   com.apple.QuartzCore          	0x00007fff5a20f96e CAOpenGLLayerDraw(CAOpenGLLayer*, double, CVTimeStamp const*, unsigned int) + 1959
7   com.apple.QuartzCore          	0x00007fff5a20efcf -[CAOpenGLLayer _display] + 565
8   com.apple.AppKit              	0x00007fff4c9189be -[NSOpenGLLayer display] + 240
9   com.apple.QuartzCore          	0x00007fff5a1b0d3b CA::Layer::display_if_needed(CA::Transaction*) + 633
10  com.apple.QuartzCore          	0x00007fff5a1b07f9 CA::Layer::layout_and_display_if_needed(CA::Transaction*) + 35
11  com.apple.QuartzCore          	0x00007fff5a1af894 CA::Context::commit_transaction(CA::Transaction*) + 326
12  com.apple.QuartzCore          	0x00007fff5a1af43d CA::Transaction::commit() + 487
[...]

PR #140, opened 2018-02-13
Use calculated line widths for gutter text

cmyr:
A possible alternative to #137?


cmyr:
Okay I played around with also caching the calculated width on the first call, but with swift value semantics you can't modify `self` in a getter.

Also tested a bunch of fonts, and everything but Zapfino renders correctly, and even that isn't too too bad, so I'm going to merge this as is.

closes #137 
closes #135

PR #141, opened 2018-02-14
Ignore fullscreen window when positioning new one

cmyr:
This requires us to do some bookkeeping ourselves to track whether a
given window is fullscreen or not.

Edit:

I've added another commit to this (in response to https://github.com/google/xi-mac/issues/128#issuecomment-365868531), which enforces a minimum size for new windows.

### DIFF
--- a/XiEditor/XiDocumentController.swift
+++ b/XiEditor/XiDocumentController.swift
@@ -60,7 +60,7 @@ class XiDocumentController: NSDocumentController {
 
     /// Updates the location used for creating new windows on launch
     @objc func windowChangedNotification(_ notification: Notification) {
-        if let window = notification.object as? XiWindow, window.isFullscreen == false {
+        if let window = notification.object as? XiWindow, !window.isFullscreen {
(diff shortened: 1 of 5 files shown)


vlovich:
Does it make sense to restore the windows as full-screen instead of just ignoring them?

XiEditor/XiDocumentController.swift
```diff
@@ -46,23 +46,45 @@ class XiDocumentController: NSDocumentController {
...
+            self,
+            selector: #selector(XiDocumentController.didLeaveFullscreen(_:)),
+            name: NSWindow.didExitFullScreenNotification, object: nil)
     }
-    
+
     /// Updates the location used for creating new windows on launch
     @objc func windowChangedNotification(_ notification: Notification) {
-        if let window = notification.object as? NSWindow {
+        if let window = notification.object as? XiWindow, window.isFullscreen == false {
```
raphlinus: Generally I prefer `!window.isFullscreen` but that's a style issue.

PR #144, opened 2018-02-16
Add make uppercase/lowercase

scholtzan:
This adds functionality to make text uppercase and lowercase.
This relies on https://github.com/google/xi-editor/pull/523

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -389,8 +389,8 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         "transpose:": "transpose",
         "yank:": "yank",
         "cancelOperation:": "cancel_operation",
-        "uppercaseWord": "uppercase",
-        "lowercaseWord": "lowercase",
+        "uppercaseWord:": "uppercase",
+        "lowercaseWord:": "lowercase",
     ]
 
     override func doCommand(by aSelector: Selector) {
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -389,6 +389,8 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         "transpose:": "transpose",
         "yank:": "yank",
         "cancelOperation:": "cancel_operation",
+        "uppercaseWord": "uppercase",
+        "lowercaseWord": "lowercase",
     ]
 
     override func doCommand(by aSelector: Selector) {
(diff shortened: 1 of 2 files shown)


XiEditor/EditViewController.swift
```diff
@@ -259,6 +259,14 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         editView.inputContext?.discardMarkedText()
         document.sendRpcAsync("select_all", params: [])
     }
+    
+    override func uppercaseWord(_ sender: Any?) {
+        document.sendRpcAsync("uppercase", params: [])
```
cmyr: The way we handle system selectors is that we actually catch them in the `EditView` itself, in the `doCommand` method; there's a map there that translates from Cocoa commands to rpcs. As is neither of these will currently get called.

scholtzan:
Thanks! I updated it. Before I could only make the text uppercase/lowercase by clicking on the menu entries. Now it also works using custom key bindings.

cmyr:
Ah interesting, I didn't even know those menu items were there, and didn't know that they worked as implemented; hence my baroque workaround. Cool.

XiEditor/EditViewController.swift
cmyr: This is... ambiguous. The `Events` enum was a drive-by addition, and hasn't really been used consistently. I think the longer-term goal would be to have a `XiCore` protocol that is implemented by `CoreConnection` (or a thin wrapper) which exposes the API. In any case I don't mind this as is, I don't love the `Events` API particularly, and this is clear.

XiEditor/EditView.swift
```diff
@@ -389,6 +389,8 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         "transpose:": "transpose",
         "yank:": "yank",
         "cancelOperation:": "cancel_operation",
+        "uppercaseWord": "uppercase",
```
raphlinus: These need to have trailing colons (confirm that it works when changed, doesn't work for me as-is).

cmyr:
@scholtzan something weird happened to this PR, you've dragged in some stale changes, not totally sure what's up.

scholtzan:
Ooh, right! I was syncing my fork a couple of days ago to get the newest updates. 
I will try to fix it now.

scholtzan:
Fixed. And lesson learned to do my changes on a separate branch. ;)

raphlinus:
Thanks!

PR #149, opened 2018-03-06
changed beginning and end of line selectors to use existing commands

nangtrongvuon:
After making editViewController handle selectors, pressing Cmd+Left calls moveToBeginningOfLine and pressing Cmd+Right calls moveToEndOfLine, which does not have an existing command implementation in xi-core, making Cmd+Left and Cmd+Right not have its intended behaviour (to move to beginning or end of line)

These fixes points the selectors to the move_to_left_end_of_line and move_to_right_end_of_line commands that exists in xi-core respectively.


nangtrongvuon:
I have signed the CLA.

nangtrongvuon:
I added my GitHub email to my commits and my GitHub account.

nangtrongvuon:
I forgot I was committing with my company email. Is it ok if I create a new branch?

PR #150, opened 2018-03-06
changed beginning and end of line selectors to use existing commands

nangtrongvuon:
After making editViewController handle selectors, pressing Cmd+Left calls moveToBeginningOfLine and pressing Cmd+Right calls moveToEndOfLine, which does not have an existing command implementation in xi-core, making Cmd+Left and Cmd+Right not have its intended behaviour (to move to beginning or end of line)

These fixes points the selectors to the move_to_left_end_of_line and move_to_right_end_of_line commands that exists in xi-core respectively.

This is almost identical to the previous PR, except I have committed them with the email I set in the CLA.


nangtrongvuon:
> I wonder if the keybindings have changed somehow - it seems to me that "left" and "right" make more sense if they're bound to arrow keys than "beginning" and "end". We'll have to revisit this when it comes time to do bidi, as the behavior is distinguishable, but this looks fine for now.
> 
> Do the shift versions of these keybindings add the "AndModifySelection" to the selector?
> 
> One tiny nit, I think we want the selectors in alphabetical order (which doesn't always match up with logical groupings, but should make them easier to find).

The Shift version also adds that, as I've just found out. Pressing Cmd+Shift+Left or Cmd+Shift+Right also complains about unhandled selector.

Issue #136, opened 2018-02-05
capture core stderr, write on crash


cmyr:
afaict there's no way to recover a stacktrace if xi-core crashes when launched outside Xcode.

Related to this, xi-mac currently doesn't have any way of knowing whether or not xi-core has crashed.

To fix the first, we should capture the output of xi-core' stderr, and store it as a string in xi-mac; perhaps only keeping the most recent 100 lines or so. (We should probably continue printing stderr as well, at least if `DEBUG` is set).

To fix the second, `CoreConnection` should keep a handle to the process, and periodically call `process.isRunning`. If this returns false, we should dump the contents of the captured stderr to disk somewhere.

cmyr:
closed by #151

PR #151, opened 2018-03-09
Capture core stderr and writes to log on core crash

nangtrongvuon:
This PR is meant to address issue #136.

A handler is added to capture stderr output from xi-core and print it into Xcode's console. When xi-core crashes or terminates, a log of stderr output is dumped to the user's /Library/Logs/XiEditor folder with a timestamp in its filename. 

A button to open the log folder has also been added to the debug menu.

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -141,6 +141,9 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         return applicationDirectory
     }()
     
+    // The default name for XiEditor's error logs
+    let defaultCoreLogName = "xi_tmp.log"
+    
     lazy var errorLogDirectory: URL = {
         let logDirectory = FileManager.default.urls(
         for: .libraryDirectory,
@@ -217,7 +220,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     
     // Clean up temporary Xi stderr log
     func applicationWillTerminate(_ notification: Notification) {
-        let tmpErrLogFile = errorLogDirectory.appendingPathComponent("xi_tmp.log")
+        let tmpErrLogFile = errorLogDirectory.appendingPathComponent(defaultCoreLogName)
         do {
             try FileManager.default.removeItem(at: tmpErrLogFile)
         } catch let err as NSError {
@@ -416,6 +419,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
             Events.SaveTrace(destination: destination, frontendSamples: Trace.shared.snapshot()).dispatch(self.dispatcher!)
         }
     }
+    
     @IBAction func openErrorLogFolder(_ sender: Any) {
         NSWorkspace.shared.selectFile(nil, inFileViewerRootedAtPath: errorLogDirectory.path)
     }
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -83,7 +83,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     var documentController: XiDocumentController!
 
     // This is set to 'InconsolataGo' in the user preferences; this value is a fallback.
-    let fallbackFont = CTFontCreateWithName(("Menlo" as CFString?)!, 14, nil)
+    let fallbackFont = CTFontCreateWithName(("Inconsolata" as CFString?)!, 14, nil)
 
     lazy fileprivate var _textMetrics = TextDrawingMetrics(font: self.fallbackFont,
                                                            textColor: self.theme.foreground)
@@ -144,25 +144,23 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     // The default name for XiEditor's error logs
     let defaultCoreLogName = "xi_tmp.log"
     
-    lazy var errorLogDirectory: URL = {
+    lazy var errorLogDirectory: URL? = {
         let logDirectory = FileManager.default.urls(
-        for: .libraryDirectory,
-        in: .userDomainMask)
-        .first!
-        .appendingPathComponent("Logs")
-        .appendingPathComponent("XiEditor")
+            for: .libraryDirectory,
+            in: .userDomainMask)
+            .first!
+            .appendingPathComponent("Logs")
+            .appendingPathComponent("XiEditor")
         
         // create XiEditor log folder on first run
-        if !FileManager.default.fileExists(atPath: logDirectory.path) {
-            do {
-                try FileManager.default.createDirectory(at: logDirectory,
-                                                        withIntermediateDirectories: true,
-                                                        attributes: nil)
-            } catch let err as NSError {
-                print("failed to create error log directory. \(err)")
-            }
+        do {
+            try FileManager.default.createDirectory(at: logDirectory,
+                                                    withIntermediateDirectories: true,
+                                                    attributes: nil)
+        } catch let err as NSError {
+            print("failed to create error log directory. \(err)")
+            return nil
         }
-        
         return logDirectory
     }()
 
@@ -186,7 +184,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         Trace.shared.trace("appWillLaunch", .main, .begin)
 
         guard let corePath = Bundle.main.path(forResource: "xi-core", ofType: ""),
-        let bundledPluginPath = Bundle.main.path(forResource: "plugins", ofType: "")
+            let bundledPluginPath = Bundle.main.path(forResource: "plugins", ofType: "")
             else { fatalError("Xi bundle missing expected resouces") }
 
         let dispatcher: Dispatcher = {
@@ -199,7 +197,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         updateRpcTracingConfig(collectSamplesOnBoot)
 
         let params = ["client_extras_dir": bundledPluginPath,
-                           "config_dir": getUserConfigDirectory()]
+                      "config_dir": getUserConfigDirectory()]
         dispatcher.coreConnection.sendRpcAsync("client_started",
                                                params: params)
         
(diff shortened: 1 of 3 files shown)


XiEditor/CoreConnection.swift
```diff
@@ -39,24 +39,37 @@ struct FileWriter {
             return nil
         }
     }
-
```
cmyr: There's a bunch of whitespace changes in this file that should be cleaned up. Xcode likes to introduce them. 😕

XiEditor/CoreConnection.swift
```diff
@@ -39,24 +39,37 @@ struct FileWriter {
...
     func write(bytes: Data) {
         handle.write(bytes)
     }
 }
 
 class CoreConnection {
-
+    
+    let task = Process()
+    var timer: Timer?
```
cmyr: is this still used?

XiEditor/CoreConnection.swift
```diff
@@ -39,24 +39,37 @@ struct FileWriter {
...
     var inHandle: FileHandle  // stdin of core process
     var recvBuf: Data
     weak var client: XiClient?
     let rpcLogWriter: FileWriter?
+    let errLogWriter: FileWriter?
+    
+    // default log directory on MacOS is /Library/Logs
+    let logDirectory = FileManager.default.urls(
```
cmyr: we've already declared this in the app delegate; we should only declare it once. (App delegate is probably the reasonable place, and we can access the app delegate vae with `(NSApplication.shared.delegate as! AppDelegate).errorLogDirectory`

XiEditor/CoreConnection.swift
```diff
@@ -39,24 +39,37 @@ struct FileWriter {
...
+        .appendingPathComponent("Logs")
+        .appendingPathComponent("XiEditor")
     
     // RPC state
     var queue = DispatchQueue(label: "com.levien.xi.CoreConnection", attributes: [])
     var rpcIndex = 0
     var pending = Dictionary<Int, (Any?) -> ()>()
-
+    
+    var errOutput = "" // output of stderr as String
```
cmyr: I think we can delete this; see comment below

XiEditor/CoreConnection.swift
```diff
@@ -66,23 +79,72 @@ class CoreConnection {
         } else {
             self.rpcLogWriter = nil
         }
-        let task = Process()
+        
+        let tmpErrLog = logDirectory.appendingPathComponent("xi_tmp.log").path
```
cmyr: Similar to above, this string appears in a few places. I would just add a `defaultCoreLogPath` var to the app delegate, and use that anywhere we need this path.

XiEditor/CoreConnection.swift
```diff
@@ -66,23 +79,72 @@ class CoreConnection {
...
+        errPipe.fileHandleForReading.readabilityHandler = { handle in
+            let data = handle.availableData
+            var lineCount = 1
+            
+            if let errString = String(data: data, encoding: String.Encoding.utf8) {
+                print(errString, terminator: "")
+                self.errOutput += errString
+            }
+            
+            if self.errOutput.hasSuffix("\n") {
```
cmyr: Okay I get the idea here, but I think it's probably more clever than we need. Since we delete the log if we exit normally we aren't particularly worried about wasting disk space; I would just log everything as it comes in, and not worry about it.

There's also a logic bug here, in that it's possible for errString to include interior (non-trailing) newlines.

XiEditor/CoreConnection.swift
```diff
@@ -66,23 +79,72 @@ class CoreConnection {
...
+        
+        // write to log on xi-core crash
+        task.terminationHandler = { _ in
+            // get current date to use as timestamp
+            let dateFormatter = DateFormatter()
+            let currentTime = Date.init()
+            dateFormatter.dateFormat = "yyyy-MM-dd-HHMMSS"
+            let timeStamp = dateFormatter.string(from: currentTime)
+            
+            let tmpErrLog = self.logDirectory.appendingPathComponent("xi_tmp.log")
```
cmyr: see earlier comment about having this be declared only once

XiEditor/AppDelegate.swift
```diff
@@ -140,6 +140,28 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
...
+        .appendingPathComponent("XiEditor")
+        
+        // create XiEditor log folder on first run
+        if !FileManager.default.fileExists(atPath: logDirectory.path) {
+            do {
+                try FileManager.default.createDirectory(at: logDirectory,
+                                                        withIntermediateDirectories: true,
+                                                        attributes: nil)
+            } catch let err as NSError {
+                print("failed to create error log directory. \(err)")
```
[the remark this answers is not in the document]
nangtrongvuon: I kind of emulated the config directory above that code, but I don't think there will ever be any reason it'd fail. I think I'll make it a fatal error if the directory is missing (probably only possible if the user deletes it while running Xi?)

XiEditor/CoreConnection.swift
vlovich: Agreed.  Let's unify the path computation.

XiEditor/CoreConnection.swift
vlovich: +1.  Let's just dump directly to disk.  If stderr verbosity is a problem we can either reduce the verbosity or add log rollover.  This also gets rid of the in-memory buffering which means you can lose partial log data if the frontend crashes which is likely to contain the important info you want.

XiEditor/CoreConnection.swift
vlovich: +1.

XiEditor/AppDelegate.swift
```diff
@@ -83,7 +83,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     var documentController: XiDocumentController!
 
     // This is set to 'InconsolataGo' in the user preferences; this value is a fallback.
-    let fallbackFont = CTFontCreateWithName(("Menlo" as CFString?)!, 14, nil)
+    let fallbackFont = CTFontCreateWithName(("Inconsolata" as CFString?)!, 14, nil)
```
cmyr: maybe this was committed by mistake?

XiEditor/AppDelegate.swift
```diff
@@ -144,25 +144,23 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
...
+    lazy var errorLogDirectory: URL? = {
         let logDirectory = FileManager.default.urls(
-        for: .libraryDirectory,
-        in: .userDomainMask)
-        .first!
-        .appendingPathComponent("Logs")
-        .appendingPathComponent("XiEditor")
+            for: .libraryDirectory,
+            in: .userDomainMask)
+            .first!
```
cmyr: since this is now a nil, we can just use `.first?`

XiEditor/AppDelegate.swift
```diff
@@ -144,25 +144,23 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
...
-            } catch let err as NSError {
-                print("failed to create error log directory. \(err)")
-            }
+        do {
+            try FileManager.default.createDirectory(at: logDirectory,
+                                                    withIntermediateDirectories: true,
+                                                    attributes: nil)
+        } catch let err as NSError {
+            print("failed to create error log directory. \(err)")
+            return nil
```
cmyr: we don't need to handle the error if we're okay being nil; so this can just be something like `return try? FileManager...`

XiEditor/CoreConnection.swift
```diff
@@ -112,9 +112,11 @@ class CoreConnection {
             let currentTime = Date.init()
             dateFormatter.dateFormat = "yyyy-MM-dd-HHMMSS"
             let timeStamp = dateFormatter.string(from: currentTime)
-            
-            let tmpErrLog = self.logDirectory.appendingPathComponent(self.tmpLogName)
-            let timestampedLog = self.logDirectory.appendingPathComponent("XiEditor_\(timeStamp).log")
+
+            guard let tmpErrLog = self.appDelegate.errorLogDirectory?.appendingPathComponent(self.appDelegate.defaultCoreLogName)
```
cmyr: you can write multiple clauses in a single guard statement in swift, separated by commas, as in:

```swift
var maybeBool: Bool?
guard let bool = maybeBool, bool == false else {
    return
}
```

XiEditor/AppDelegate.swift
```diff
@@ -140,6 +140,25 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
...
+            in: .userDomainMask)
+            .first?
+            .appendingPathComponent("Logs")
+            .appendingPathComponent("XiEditor")
+
+        // create XiEditor log folder on first run
+        guard logDirectory != nil else { return nil }
+        try? FileManager.default.createDirectory(at: logDirectory!,
+                                                 withIntermediateDirectories: true,
+                                                 attributes: nil)
```
vlovich: Am I misreading this?  It seems like it swallows the createDirectory failure instead of resulting in nil.

XiEditor/AppDelegate.swift
```diff
@@ -140,6 +140,25 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
...
+        let logDirectory = FileManager.default.urls(
+            for: .libraryDirectory,
+            in: .userDomainMask)
+            .first?
+            .appendingPathComponent("Logs")
+            .appendingPathComponent("XiEditor")
+
+        // create XiEditor log folder on first run
+        guard logDirectory != nil else { return nil }
+        try? FileManager.default.createDirectory(at: logDirectory!,
```
vlovich: I a bit rusty with Swift but doesn't the guard above remove the need to unwrap logDirectory here?

XiEditor/AppDelegate.swift
nangtrongvuon: From what I understand:

Here, logDirectory is an URL? which will become nil if the user's /Library/ folder doesn't exist (I'm not sure exactly when, maybe when the admin disables it). `FileManager.default.createDirectory` returns void, as detailed in the documentation [here.](https://developer.apple.com/documentation/foundation/filemanager/1407884-createdirectory). I used `try?` here because we don't care if the creation fails, since as you and cmyr have mentioned, it shouldn't impede on normal operation even if it does.

XiEditor/AppDelegate.swift
nangtrongvuon: As mentioned above, logDirectory is nil when the /Library/ folder doesn't exist, so if it doesn't, we just return nil right there. Using guard doesn't actually unwrap the optional logDirectory value, though, so I still had to force unwrap it after the check, since I thought using `if let` would make it harder to read.

XiEditor/AppDelegate.swift
vlovich: Since the intermediate directories arg is true, the only way for createDirectory to fail is that the directory doesn't exist.  I think the correct behaviour is that if creation fails it should set the logDirectory to nil rather than try to use it & fail later.

XiEditor/AppDelegate.swift
vlovich: OK, as I mentioned I haven't used Swift in a while.  Thanks for explaining.

vlovich:
LGTM.

cmyr:
Okay built this one last time and looks like it's working as advertised. thanks @nangtrongvuon!

Issue #145, opened 2018-02-21
Horizontal scrolling


raphlinus:
Note: see #4 for an earlier discussion. This issue encapsulates my current thinking.

## Viewport width: longest line in document, or longest visible line

One of the more fundamental design choices is how to compute the viewport width. The two most reasonable alternatives are the longest line in the document, or the longest line in the viewport. In the latter case, the viewport width would change dynamically on scrolling. In the former case, it would be a function only of changes to the document.

This choice has profound implications for the architecture; for "longest line in document," the core needs to be involved, which also means that it has to understand "width". This is planned (for line wrapping) but workable horizontal scrolling shouldn't block on it.

Thus, the plan of record is that it is based on **the longest visible line** for now, and, when width infrastructure is in place, we can revisit it (possibly offering an option).

## Wiring up scrolling

For the most part, scrolling works now, the tough part is setting the viewport width. This basically means writing a function to compute the width (max of widths of visible lines) and adding invalidation logic to recalculate it as necessary. The visible line width is available as the `width` parameter of the TextLine, available as associated data from the line cache. We might want to make the width eagerly calculated on TextLine build, as the current implementation *might* be slow (it calls into CoreText).

Special care may be required to keep the horizontal scroll bar from flickering in and out of existence (which may cause cascading layout recalculations). One approach I'd be happy with is that it's sticky - once it's set it stays.

## Shadow

We had a "ShadowView" that drew a partially transparent gradient over the main EditText view, dependent on the scroll status. It would be nice to bring this back. The best place would be the `render` method in EditView, near the end.

We could write a new shader for it, or reuse an existing shader. The easiest would probably be to draw a fake "glyph" with the real width, but placeholder height, then use stretching to make it cover the window height.

## Optimization

We should do this in stages, first getting it working, then performance tuning.

Currently, the draw operation for TextLine iterates all the glyphs in the line. We should introduce a horizontal viewport, and do culling based on that.

The glyphs in a TextLine are almost monotonic. If they were fully monotonic, it would be straightforward to binary-search for the first glyph that fits inside the viewport. Even so, I think the approach is viable, with some care. During build, compute a "max deviation from monotonicity" (which in the monotonic case is likely zero). Binary search (even on not-quite-sorted data) for the left side of the viewport minus this slop factor. Iterate through the glyphs until the right side of the viewport plus this slop factor is reached. We might also need to compute a max glyph width (should also be straightforward on build).

[...]

PR #152, opened 2018-03-13
horizontal scroll

cmyr:
An initial no-fuss implementation of horizontal scrolling, including drawing shadows when text is occluded. 

towards #145

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -405,7 +405,12 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         }
     }
 
-    // Rendering using TextPlane
+    /// Our equivalent of drawRect:, with rendering using TextPlane.
+    ///
+    /// Note: This function has side effects in two cases: It stashes
+    /// the length of the longest line it sees, and passes this to the view controller;
+    /// if this line is longer than the previous longest, the view's width is updated.
+    /// It also updates IME state if IME is active.
     func render(_ renderer: Renderer, dirtyRect: NSRect) {
         Trace.shared.trace("EditView render", .main, .begin)
         renderer.clear(dataSource.theme.background)
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -130,7 +130,6 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
     }
     var lastRevisionRendered = 0
     var gutterXPad: CGFloat = 8
-    var gutterWidth: CGFloat = 0
     var gutterCache: GutterCache?
 
     var dataSource: EditViewDataSource!
@@ -364,7 +363,7 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
     /// Note: - The returned position is not guaruanteed to be an existing line. For instance, if a buffer does not fill the current window, a point below the last line will return a buffer position with a line number exceeding the number of lines in the file. In this case position.column will always be zero.
     func bufferPositionFromPoint(_ point: NSPoint) -> BufferPosition {
         let point = self.convert(point, from: nil)
-        let x = point.x + scrollOrigin.x - gutterWidth
+        let x = point.x + scrollOrigin.x - dataSource.gutterWidth
         let y = point.y + scrollOrigin.y
         let lineIx = yToLine(y)
         if let line = getLine(lineIx) {
@@ -414,7 +413,7 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         }
         let linespace = dataSource.textMetrics.linespace
         let topPad = linespace - dataSource.textMetrics.ascent
-        let xOff = gutterWidth + x0 - scrollOrigin.x
+        let xOff = dataSource.gutterWidth + x0 - scrollOrigin.x
         let yOff = topPad - scrollOrigin.y
         let firstVisible = max(0, Int((floor(dirtyRect.origin.y - topPad + scrollOrigin.y) / linespace)))
         let lastVisible = Int(ceil((dirtyRect.origin.y + dirtyRect.size.height - topPad + scrollOrigin.y) / linespace))
@@ -428,6 +427,7 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         let font = dataSource.textMetrics.font as CTFont
         let styleMap = dataSource.styleMap.locked()
         var textLines: [TextLine?] = []
+        var maxLineWidth: Double = 0
 
         // The actual drawing is split into passes for correct visual presentation and
         // also to improve batching of the OpenGL draw calls.
@@ -464,6 +464,7 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
                 lineCache.setAssoc(lineIx, assoc: assoc)
                 textLines.append(textLine)
             }
+            maxLineWidth = max(maxLineWidth, textLine.width)
             let y0 = yOff + linespace * CGFloat(lineIx)
             renderer.drawLineBg(line: textLine, x0: GLfloat(xOff), yRange: GLfloat(y0)..<GLfloat(y0 + linespace), selColor: selArgb)
         }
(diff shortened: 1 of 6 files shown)


cmyr:
playing around with this a bit more and I can get a crash pretty consistently when scrolling vigorously on a trackpad:

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.levien.XiEditor           	0x000000010f866015 _T08XiEditor8EditViewC6renderyAA8RendererC_SC6CGRectV9dirtyRecttFTf4gnn_n + 18501 (EditView.swift:428)
1   com.levien.XiEditor           	0x000000010f85cda4 _T08XiEditor8EditViewCAA17TextPlaneDelegateA2aDP6renderyAA8RendererC_SC6CGRectV9dirtyRecttFTW + 68
2   com.levien.XiEditor           	0x000000010f830d1f _T08XiEditor14TextPlaneLayerC4drawySo15NSOpenGLContextC2in_So0G13GLPixelFormatC05pixelK0Sd03forE4TimeSPySC11CVTimeStampVG07displayN0tFTf4ddddn_n + 639
3   com.levien.XiEditor           	0x000000010f82fb41 _T08XiEditor14TextPlaneLayerC4drawySo15NSOpenGLContextC2in_So0G13GLPixelFormatC05pixelK0Sd03forE4TimeSPySC11CVTimeStampVG07displayN0tFTo + 49
4   com.apple.QuartzCore          	0x00007fffa78ae165 CAOpenGLLayerDraw(CAOpenGLLayer*, double, CVTimeStamp const*, unsigned int) + 873
5   com.apple.QuartzCore          	0x00007fffa78adce3 -[CAOpenGLLayer _display] + 351
6   com.apple.AppKit              	0x00007fff9fa50332 -[NSOpenGLLayer display] + 241
7   com.apple.QuartzCore          	0x00007fffa78c1052 CA::Layer::display_if_needed(CA::Transaction*) + 572
8   com.apple.QuartzCore          	0x00007fffa78c117d CA::Layer::layout_and_display_if_needed(CA::Transaction*) + 35
9   com.apple.QuartzCore          	0x00007fffa78b6934 CA::Context::commit_transaction(CA::Transaction*) + 280
10  com.apple.QuartzCore          	0x00007fffa77ab7e1 CA::Transaction::commit() + 475
11  com.apple.AppKit              	0x00007fff9f9f506b -[_NSScrollingConcurrentMainThreadSynchronizer _synchronize:completionHandler:] + 380
12  com.apple.AppKit              	0x00007fff9f9f4ebd __80-[_NSScrollingConcurrentMainThreadSynchronizer initWithSharedData:constantData:]_block_invoke + 145
13  libdispatch.dylib             	0x00007fffb73ce8fc _dispatch_client_callout + 8
14  libdispatch.dylib             	0x00007fffb73e3a16 _dispatch_continuation_pop + 533
15  libdispatch.dylib             	0x00007fffb73d9a6c _dispatch_source_latch_and_call + 194
16  libdispatch.dylib             	0x00007fffb73d113b _dispatch_source_invoke + 1248
17  libdispatch.dylib             	0x00007fffb73db908 _dispatch_main_queue_callback_4CF + 505
18  com.apple.CoreFoundation      	0x00007fffa1cbcbc9 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
19  com.apple.CoreFoundation      	0x00007fffa1c7dc0d __CFRunLoopRun + 2205
20  com.apple.CoreFoundation      	0x00007fffa1c7d114 CFRunLoopRunSpecific + 420
21  com.apple.HIToolbox           	0x00007fffa11ddebc RunCurrentEventLoopInMode + 240
22  com.apple.HIToolbox           	0x00007fffa11ddcf1 ReceiveNextEventCommon + 432
23  com.apple.HIToolbox           	0x00007fffa11ddb26 _BlockUntilNextEventMatchingListInModeWithFilter + 71
24  com.apple.AppKit              	0x00007fff9f774a54 _DPSNextEvent + 1120
25  com.apple.AppKit              	0x00007fff9fef07ee -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 2796
26  com.apple.AppKit              	0x00007fff9f7693db -[NSApplication run] + 926
27  com.apple.AppKit              	0x00007fff9f733e0e NSApplicationMain + 1237
28  com.levien.XiEditor           	0x000000010f8144c9 main + 9 (AppDelegate.swift:80)
29  libdyld.dylib                 	0x00007fffb7404235 start + 1
```

There's also some occasional jank around scroller rendering, so I'm going to investigate a bit further.

XiEditor/ShadowView.swift
```diff
@@ -0,0 +1,99 @@
...
+        self.layer!.addSublayer(rightShadow)
+
+        updateVisibility()
+    }
+
+    override func updateLayer() {
+        leftShadow.frame = CGRect(x: leftShadowMinX, y: self.bounds.origin.y,
+                                  width: 4, height: self.bounds.height)
+
+        rightShadow.frame = CGRect(x: self.bounds.width - 4, y: self.bounds.origin.y,
```
nangtrongvuon: It seems to me that the right shadow currently doesn't compensate for the scroll bar? The left shadow is much more visible than the right shadow, unless that's intended.

XiEditor/ShadowView.swift
```diff
@@ -0,0 +1,99 @@
...
+        leftShadow.autoresizingMask = .layerHeightSizable
+
+        rightShadow.colors = [shadowColor.cgColor, NSColor.clear.cgColor]
+        rightShadow.transform = CATransform3DMakeRotation(CGFloat.pi / 2, 0, 0, 1)
+        rightShadow.opacity = 0.4
+        rightShadow.autoresizingMask = [.layerHeightSizable, .layerMinXMargin]
+    }
+
+    // by default hiding and revealing our shadows is animated, which is distracting
+    func action(for layer: CALayer, forKey event: String) -> CAAction? {
```
nangtrongvuon: I personally found that it looked really nice. Maybe the duration could be reduced?

cmyr: I wouldn't be opposed to a very short-duration animation, as long as there was some way to specify that the animation should begin from whatever the existing animation state is, so that we don't have animations that interrupt each other; I looked into this briefly but it seems harder to do in AppKit than it would've been in UIKit.

When animations were on, it also caused the right shadow view to animate when you first opened the window, which would have to be fixed.

I initially thought no animations would be a bit jarring, but playing around with it I think it feels okay; and looking at other editors (particularly sublime) this is their behaviour as well.

XiEditor/ShadowView.swift
cmyr: Yea, I'm still slightly confused by this, and can't tell if it's an optical illusion or something. I think the right side looks less sharp because the hard edge of the shadow is shared with the hard edge of the window.

I don't think we want the shadow to be outside the scrollbar; the scrollers on modern macOS are supposed to be semi-translucent overlays on top of the content, not individual views outside of the content.

One possible problem I see with this is drawing on dark backgrounds; in that case the shadows aren't very visible. Themes are supposed to provide a `shadowColor` for this, but not many do.

XiEditor/Main.storyboard
```diff
@@ -39,6 +39,9 @@
                             <customView wantsLayer="YES" translatesAutoresizingMaskIntoConstraints="NO" id="q1P-kd-995" customClass="EditView" customModule="XiEditor" customModuleProvider="target">
                                 <rect key="frame" x="0.0" y="0.0" width="480" height="270"/>
                             </customView>
+                            <customView translatesAutoresizingMaskIntoConstraints="NO" id="x8W-KD-KOP" customClass="ShadowView" customModule="XiEditor" customModuleProvider="target">
```
raphlinus: For me, I get a crash in ShadowView.setup because layer is nil, unless I also add `wantsLayer="YES"` here. Wonder why it seems to be working for you but not me.

XiEditor/EditView.swift
```diff
@@ -464,6 +466,7 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
                 lineCache.setAssoc(lineIx, assoc: assoc)
                 textLines.append(textLine)
             }
+            maxLineWidth = max(maxLineWidth, textLine.width)
```
raphlinus: I'm slightly uncomfortable with drawing being side-effecting. I'd prefer if the width recalculation were done separately, triggered by updates. It occurs to me that logic will need full access to the TextLine objects, which means that we'd have to move the "get cached TextLine" logic into a separate method. Among other things, I'm pretty sure this causes a double draw when the width increases.

As a side note, side effects from drawing are also used by the IME logic.

cmyr: I tried an approach similar to what you'd described, but because LineCache has a generic payload I can't even really access the `TextLine`s. (It occurred to me just now that I could write an extension for `LineCache where T == TextLine`, which might work). 

Also, when text lines arrive we tend to be blocked for drawing; we certainly don't want to introduce additional blocking gets earlier. So doing the calculation while drawing felt like an expedient compromise.

Ideally, we would have width measurement, and core would do a bunch of the work here for us; I  think that should be the final goal.

In terms of redundant draws with the current implementation, they will be happening but only in the case where a draw introduces a new 'longest line', so the amortized cost shouldn't be too bad.

Issue #134, opened 2018-02-02
Include Inconsolata (font) in the app bundle


cmyr:
This would ensure that everyone had the same experience out of the box, and also has nice evangelism benefits. ~I'm not sure if there are any legal or otherwise concerns, though. @raphlinus?~

In terms of the actual work, bundling the fonts should be trivial; what I'm not sure about is how we actually use them from the app bundle. It's certainly doable though.

raphlinus:
I think this is a good idea. There shouldn't be any legal concerns, the fonts are released under the perfectly cromulent OFL license, which allows bundling in a wide range of contexts.

There is a question of which version is best. InconsolataGo is a possibility, but it also lacks some updates and character set coverage from the latest Google Fonts [version](https://github.com/google/fonts/tree/master/ofl/inconsolata). I've heard reports of rendering artifacts with the latter; these should be investigated.

cmyr:
closed by #153

PR #153, opened 2018-03-15
Bundle Inconsolata with Xi-Mac

nangtrongvuon:
This PR addresses issue #134.

This PR simply bundles the Inconsolata version taken from Google Fonts with Xi-Mac and also includes the Open Font License that comes with it.


XiEditor/AppDelegate.swift
```diff
@@ -83,7 +83,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     var documentController: XiDocumentController!
 
     // This is set to 'InconsolataGo' in the user preferences; this value is a fallback.
```
cmyr: This comment is out of date. It's also not totally clear that we need a fallback font now, but we may be relying on it in certain places (like setting up the initial drawing metrics,  below)

cmyr: (to clarify: I think it's fine to leave the fallback)

nangtrongvuon: I made the comment change to clarify Inconsolata being bundled now.

raphlinus:
Also looks good to me. I think the top-level README can be updated. First, the "users are encouraged to try Inconsolata" can be changed to: "The default font is Inconsolata, which is bundled with the app".

Second, I'd add a licensing section, right above `## Contributions`:

```
## License

This project is licensed under the Apache 2 [license](LICENSE). The bundled fonts are under a different license, the Open Font License. See the [fonts](fonts) directory for the fonts and associated license.
```

Thanks! Neat that `ATSApplicationFontsPath` makes this so easy, without having to write code to explicitly load fonts from the bundle.

raphlinus:
@nangtrongvuon I'm curious exactly what versions these are. The ones from https://github.com/google/fonts/tree/master/ofl/inconsolata are now at version 2.001, while the ones in this PR seem to be version 2.000. I think there are some bugs in this one, and am looking at maybe updating to the latest version. I'm still digging into the question of which version is the authoritative best one...

nangtrongvuon:
I took Inconsolata directly from Google Fonts site (picked Inconsolata, chose to download it only) about nearly 2 weeks ago. I tried to put in InconsolataGo instead but I couldn’t download it from your site (either dead link or inaccessible to me.)

PR #161, opened 2018-04-03
Implement unified titlebar preference

lukakerr:
Similar to other editors (Sublime, Atom...) a unified titlebar is a nice thing to have.

This can be set in preferences with `unified_titlebar = true`, and when the theme changes, the titlebar adapts as seen in the gif below:

xi

If the current theme is dark (see the new NSColor extension), the windows appearance is set to `vibrantDark`, which also darkens other UI things such as the find input bar. I think this is a nice touch, but if its not wanted, I can remove it.

This also works for tabs:

screen shot 2018-04-03 at 8 09 28 pm

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -131,6 +131,25 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
             }
         }
     }
+  
+    var unifiedTitlebar = false {
+        didSet {
+            // Dont check if value is same as previous
+            // so when theme updates, background color still changes
+            if let window = self.view.window {
+                let color = self.theme.background
+        
+                window.titlebarAppearsTransparent = unifiedTitlebar
+                window.backgroundColor = unifiedTitlebar ? color : nil
+        
+            if color.isDark && unifiedTitlebar {
+                window.appearance = NSAppearance(named: NSAppearance.Name.vibrantDark)
+            } else {
+                window.appearance = NSAppearance(named: NSAppearance.Name.aqua)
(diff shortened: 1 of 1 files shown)


cmyr:
@lukakerr super interesting! Will give this a proper look soon. I'm surprised that this works without adding support in xi-core, although it's nice that it does.

My one hesitation here is that this is a distinctly non-macOS feel, but having it off by default makes that not feel like a big deal; it looks clean.

I'm also curious how this would interact with the 10.10+  trend towards translucent titlebars / `NSVisualEffectView`? Just thinking out loud.

Anyway, very cool, thanks for the PR!

cmyr:
@lukakerr Good points. Although I like Sublime Text, I don't really hold it up as an example of a 'native mac app'. Although BBEdit is about as Mac-y an editor as you'll find, and it's titlebars are all black by default I believe?

The system apps are interesting: instead of being coloured explicitly they're mostly just making extensive use of transparency, although the calculator is an interesting counter-example maybe?

I'm also not seeing a ton of guidance in the HID.

There's obviously infinite bike shed potential here, so I'm going to try not to over-indulge; happy to have someone interested in hacking on mac UI stuff, it hasn't been a real focus lately. :)

There's also always lots of room for future tweaking, so I'm disposed towards merging this; going to ask for a second opinion though.

cmyr:
thanks for this! and you're definitely encouraged to pick at any other warts on the mac client; for more drastic changes you might want to open a discussion issue or drop by irc.mozilla.org#xi to discuss before doing too much work, but in general PRs are very welcome.

PR #165, opened 2018-04-10
Add measure_width implementation

raphlinus:
Adds a method to measure the widths of strings, and exposes it through
RPC. Also adds a debug wrap with width. To be applied in combination
with google/xi-editor#607

Part of google/xi-editor#565

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -442,7 +442,11 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     @IBAction func debugRewrap(_ sender: AnyObject) {
         document.sendRpcAsync("debug_rewrap", params: [])
     }
-    
+
+    @IBAction func debugWrapWidth(_ sender: AnyObject) {
+        document.sendRpcAsync("debug_wrap_width", params: [])
+    }
+
     @IBAction func debugSetTheme(_ sender: NSMenuItem) {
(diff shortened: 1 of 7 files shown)


XiEditor/EditViewController.swift
```diff
@@ -442,7 +442,11 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     @IBAction func debugRewrap(_ sender: AnyObject) {
         document.sendRpcAsync("debug_rewrap", params: [])
     }
-    
+
+    @IBAction func debugWrapWidth(_ sender: AnyObject) {
+        document.sendRpcAsync("debug_wrap_width", params: [])
```
cmyr: We should start thinking about what the actual protocol changes in support of this should look like.

XiEditor/StyleMap.swift
```diff
@@ -203,6 +203,23 @@ class StyleMapState: UnfairLock {
...
     }
+
+    func measureWidth(id: Int, s: String) -> Double {
+        let builder = TextLineBuilder(s, font: self.font)
+        let range = NSMakeRange(0, s.utf16.count)
+        applyStyle(builder: builder, id: id, range: range)
+        return builder.measure()
+    }
+
+    func measureWidths(_ args: [[String: AnyObject]]) -> [[Double]] {
```
cmyr: I'd keep in mind whether in this particular case it would be worth trying to stay in `NSDictionary` land.

PR #166, opened 2018-04-10
Set minimum window size

paulsmith:
Set a minimum window size so that manually resizing a window can't collapse it all the way, ala:

xi-before

Most windowed Mac apps keep a small content area displayed at their smallest size. After the change:

xi-after

### DIFF
--- a/XiEditor/Document.swift
+++ b/XiEditor/Document.swift
@@ -89,7 +89,7 @@ class Document: NSDocument {
         }
         
         windowController.window?.setFrame(newFrame, display: true)
-        windowController.window?.minSize = NSSize(width: 200, height: 100)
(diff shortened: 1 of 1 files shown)


cmyr:
Cool, this is a good addition, thanks!

In `Document.swift:199` we declare a minimum size for _new_ windows; I think it makes sense for us to use the same size in both places, so maybe we could declare a static `minWindowSize` (or `minWindowWidth` + `minWindowHeight` in `Document` and then use that both for new windows and for this setting?

paulsmith:
Good point - I just simplified it by reusing the frame object to take its size for the min size instead.

cmyr:
Thanks again @paulsmith!

PR #175, opened 2018-05-01
Add process/thread name capturing for traces

vlovich:
Depends on https://github.com/google/xi-editor/pull/588 (which has the screenshot of what the end-to-end work looks like in Chrome trace).

### DIFF
--- a/XiEditor/Trace.swift
+++ b/XiEditor/Trace.swift
@@ -21,6 +21,22 @@ protocol Writeable {
     func write(_ data: Data)
 }
 
+func gettid() -> UInt64 {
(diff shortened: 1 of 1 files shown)


vlovich:
Yeah, I'd prefer to not store the name the way it is now.  The problem is that thread names change at arbitrary points in time & there's no way to register a listener.  You might be able to create a global structure to store it uniquely which will trade-off memory for CPU (higher synchronization cost). 

For me the big open question is the impedance mismatch between OSX programming & chrome trace event samples since the latter doesn't support the concept of thread pools.  Maybe we can file an enhancement request to the Chrome team?

Issue #155, opened 2018-03-20
scrollbars show in empty window if resizing down from original size


cmyr:
When a new window is created, resizing it horizontally to be smaller than it's original size causes scrollbars to appear.

This happens because in `EditView.redrawEverything()` we set the width constraint to be == the current width of the window. This is incorrect, but is necessary because when determining the width of the view (which increases as we encounter new lines) we don't save this width, except in the state of the constraint on the view. This means that when we're resetting, we don't know if the current constraint's value is the result of a long line or not.

I'll get around to this at some point, but if someone else wanted to take it i'd encourage them to. :)

PR #177, opened 2018-05-03
Fix scrollbars showing when text is shorter than window size

nangtrongvuon:
This fix aims to addresses issue #155.


nangtrongvuon:
Well you’re right, more work on this is needed. I’ll look into this more first.

cmyr:
@nangtrongvuon I think we should close this for now, and decide how to handle this properly. I _think_ that means tracking the width of each line in core, and keeping the client informed (somehow) about the width of the longest existing line.

PR #182, opened 2018-05-25
Modify build-rust-xcode.sh to handle syntect update

cmyr:
With https://github.com/google/xi-editor/pull/672, the syntect plugin can now include config overrides for languages. This patch includes these files when installing the plugin on xi-mac.

### DIFF
--- a/build-rust-xcode.sh
+++ b/build-rust-xcode.sh
@@ -70,13 +70,17 @@ function build_target () {
     fi
 }
 
+function install_syntect () {
+    SYNTECT_ROOT="${SRCROOT}/xi-editor/rust/syntect-plugin"
+    mkdir -p "${BUILT_PRODUCTS_DIR}/plugins/syntect/bin"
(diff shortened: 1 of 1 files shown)


eyelash:
Why not use the makefile instead of duplicating its logic here? This would also give the makefile some more real world testing.

cmyr:
@eyelash this is definitely a [Chesterton's fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence) decision for me; there's a bunch of stuff going on in this file, and I'm not totally sure if it's history or rationale, so I left as much of it in place as I could.

Make also has troubles handling paths with spaces, which are super common on the mac, and I couldn't come up with a very satisfactory solution to that. 😒

jansol:
I believe the copy step touched by this PR belongs into the xcodeproj as a custom build step. That ought to also avoid the whitespace issue.

Issue #180, opened 2018-05-17
Status Bar for Xi-Mac


nangtrongvuon:
Along with xi-editor's issue [#665](https://github.com/google/xi-editor/issues/665), this issue serves as a place to track the work in progress to implement a status bar frontend in Xi-Mac.

Part of #176 

- [x] Blank status bar anchored at bottom of each EditView
- [x] Front end support for status bar items
- [x] Wiring status bar and items to core

Issue #176, opened 2018-05-03
[GSoC] Xi-Mac UI Improvements


nangtrongvuon:
This is a list of goals that I aim to meet in order to improve Xi-Mac's UI and make it more featured and usable editor on MacOS.

This roadmap is a work in progress and subject to changes and improvements.

**Roadmap**
- [x] Status bar
- [x] Implement UI elements for LSP Support
    - [x] Autocomplete Menu
    - [x] Show Definition Menu
    - [x] Go to Definition Menu
- [ ] Implement split views
- [x] Tab/Window behavior examination

PR #183, opened 2018-05-28
Initial implementation of Xi-Mac status bar

nangtrongvuon:
This PR adds a initial status bar implementation, which supports the addition, update and deletion of status items. 

#180 
Part of #176 
To be used with xi-editor's [#677](https://github.com/google/xi-editor/pull/677)

### DIFF
--- a/XiEditor/Client.swift
+++ b/XiEditor/Client.swift
@@ -54,7 +54,9 @@ protocol XiClient: AnyObject {
     /// A notification containing an alert message to be shown the user.
     func alert(text: String);
 
-    // Status bar notifications
+    /// A list of notifications that manages status items.
+    /// Keys are unique, and alignment (left or right) cannot be
+    /// changed after creating the status item.
     func addStatusItem(viewIdentifier: String, key: String, value: String, alignment: String);
     func updateStatusItem(viewIdentifier: String, key: String, value: String);
     func removeStatusItem(viewIdentifier: String, key: String);
--- a/XiEditor/Client.swift
+++ b/XiEditor/Client.swift
@@ -54,6 +54,11 @@ protocol XiClient: AnyObject {
     /// A notification containing an alert message to be shown the user.
     func alert(text: String);
 
+    // Status bar notifications
+    func addStatusItem(viewIdentifier: String, key: String, value: String, alignment: String);
+    func updateStatusItem(viewIdentifier: String, key: String, value: String);
+    func removeStatusItem(viewIdentifier: String, key: String);
+
     /// A notification containing changes to the current config for the given view.
     /// - Note: The first time this message is sent, `changes` contains all defined
     // config keys and their values. Subsequent calls contain only those items which
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -145,10 +145,10 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         
                 if color.isDark && unifiedTitlebar {
                     window.appearance = NSAppearance(named: NSAppearance.Name.vibrantDark)
+                    statusBar.updateStatusBarColor(newBackgroundColor: theme.background, newTextColor: theme.foreground)
                 } else {
                     window.appearance = NSAppearance(named: NSAppearance.Name.aqua)
                 }
-                statusBar.updateStatusBarColor(newBackgroundColor: theme.background, newTextColor: theme.foreground)
             }
         }
     }
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -181,7 +181,11 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     }
 
     func setupStatusBar() {
-        statusBar = StatusBar(frame: .zero, backgroundColor: self.theme.background, textColor: self.theme.foreground)
+        if self.unifiedTitlebar == true {
+            statusBar = StatusBar(frame: .zero, backgroundColor: self.theme.background, textColor: self.theme.foreground)
+        } else {
+            statusBar = StatusBar(frame: .zero)
+        }
         self.view.addSubview(statusBar)
 
         NSLayoutConstraint.activate([
@@ -213,8 +217,8 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
 
     @objc func frameDidChangeNotification(_ notification: Notification) {
         updateEditViewHeight()
-        statusBar.updateItemVisibility(windowWidth: self.editViewWidth.constant)
         willScroll(to: scrollView.contentView.bounds.origin)
+        statusBar.updateItemVisibility(windowWidth: self.view.frame.width)
         updateViewportSize()
     }
 
@@ -247,7 +251,9 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         updateViewportSize()
         editView.gutterCache = nil
         shadowView.updateShadowColor(newColor: theme.shadow)
-        statusBar.updateStatusBarColor(newBackgroundColor: theme.background, newTextColor: theme.foreground)
+        if self.unifiedTitlebar == true {
+            statusBar.updateStatusBarColor(newBackgroundColor: theme.background, newTextColor: theme.foreground)
+        }
         editView.needsDisplay = true
 
     }
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -180,10 +180,16 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
 
         if #available(OSX 10.12, *) {
             let testLabelLeft = NSTextField(labelWithString: "Xi-Mac")
+            let anotherLeft = NSTextField(labelWithString: "Xi-Mac")
+
             let testLabelRight = NSTextField(labelWithString: "Status Bar")
+            let anotherRight = NSTextField(labelWithString: "Status Bar")
+
 
             statusBar.addSBItem(testLabelLeft, alignment: .left)
+            statusBar.addSBItem(anotherLeft, alignment: .left)
             statusBar.addSBItem(testLabelRight, alignment: .right)
+            statusBar.addSBItem(anotherRight, alignment: .right)
         }
 
         shadowView.setup()
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -213,6 +213,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
 
     @objc func frameDidChangeNotification(_ notification: Notification) {
         updateEditViewHeight()
+        statusBar.updateItemVisibility(windowWidth: self.editViewWidth.constant)
         willScroll(to: scrollView.contentView.bounds.origin)
         updateViewportSize()
     }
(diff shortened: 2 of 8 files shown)


cmyr:
Haven't had a chance to take a look at this yet, but it might be a good idea to open a PR in xi-editor with the changes to the API that support this work, and maybe also some changes to the sample plugin to use this api? Or a link to some plugin that does?

nangtrongvuon:
@cmyr Will do shortly.

XiEditor/Client.swift
```diff
@@ -54,6 +54,11 @@ protocol XiClient: AnyObject {
     /// A notification containing an alert message to be shown the user.
     func alert(text: String);
 
+    // Status bar notifications
```
cmyr: I would follow the documentation style of the rest of this file; three `///` (for doc comments) and a description of the specific command.

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,140 @@
+//
```
cmyr: needs a copyright notice

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,140 @@
...
+import Cocoa
+
+enum StatusItemAlignment: String {
+    case left = "left"
+    case right = "right"
+}
+
+class StatusItem: NSTextField {
+
+    var key: String = ""
```
cmyr: perhaps these should be `let` bindings? In which case you don't need to assign them initial values, if  you assign them values in `init`.

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+class StatusBar: NSView {
+
+    private let backgroundColor = NSColor(deviceWhite: 0.9, alpha: 1.0)
```
cmyr: this can probably change based on theme or other settings (you mentioned this, just noting it again)

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,140 @@
...
+    private let statusBarHeight: CGFloat = 20
```
cmyr: why 20? (just curious)

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,140 @@
...
+        self.widthAnchor.constraint(equalTo: editView.widthAnchor).isActive = true
+        self.leadingAnchor.constraint(equalTo: editView.leadingAnchor).isActive = true
+        self.trailingAnchor.constraint(equalTo: editView.trailingAnchor).isActive = true
+        self.bottomAnchor.constraint(equalTo: editView.bottomAnchor).isActive = true
+    }
+
+    // Adds a status bar item. Only appends status bar items for now.
+    func addStatusItem(_ item: StatusItem) {
+        // If item exists, update its value
+        if let existingItem = currentItems[item.key] {
```
cmyr: I think trying to add an item that already exists either replaces the exiting item or is an error. Having it just silently update instead seems like an easy way to let people write bugs.

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,140 @@
...
+            case .left:
+                leftItems.append(item)
+            case .right:
+                rightItems.append(item)
+        }
+        self.needsUpdateConstraints = true
+    }
+
+    // Update a status bar item with a new value.
+    func updateStatusItem(_ key: String, _ value: String) {
```
cmyr: similarly trying to update an item that is missing should maybe be an error? at the very least we should log it.

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,140 @@
...
+            }
+        }
+        super.updateConstraints()
+    }
+
+    override func draw(_ dirtyRect: NSRect) {
+        super.draw(dirtyRect)
+        let path = NSBezierPath()
+        backgroundColor.setFill()
+        __NSRectFill(dirtyRect)
```
cmyr: I... don't think we're supposed to use this function. I think there's just a `fill()` method on NSRect now?

XiEditor/EditViewController.swift
```diff
@@ -165,16 +167,35 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
         super.viewDidAppear()
+        setupStatusBar()
         shadowView.setup()
         NotificationCenter.default.addObserver(self, selector: #selector(EditViewController.frameDidChangeNotification(_:)), name: NSView.frameDidChangeNotification, object: scrollView)
         // call to set initial scroll position once we know view size
         redrawEverything()
     }
 
+    func setupStatusBar() {
+        if self.unifiedTitlebar == true {
```
cmyr: we shouldn't need to do styling here. The statusbar should draw itself with the 'default style' (to be determined) by default, and if we're using the unified style we should be updating the statusbar in the same place where we update other elements (which is the `unifiedTitlebar`'s `didSet` method).

To see the advantage of this approach, notice that if you change the `unified_toolbar` setting, the statusbar isn't updated.

Similarly, if you select a dark theme the statusbar becomes unreadable.

XiEditor/EditViewController.swift
```diff
@@ -214,7 +236,11 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         willScroll(to: scrollView.contentView.bounds.origin)
         editView.gutterCache = nil
         shadowView.updateShadowColor(newColor: theme.shadow)
+        if self.unifiedTitlebar == true {
```
cmyr: see above; any properties of the statusbar related to this setting should be set in along with other changes related to this setting

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,241 @@
...
+    var currentItems: [StatusItem] {
+        return (leftItems + rightItems).sorted {$0.key < $1.key}
+    }
+
+    var lastLeftItem: StatusItem?
+    var lastRightItem: StatusItem?
+
+    var backgroundColor: NSColor = NSColor.white
+    var itemTextColor: NSColor = NSColor.black
+    var statusBarPadding: CGFloat = 10
```
cmyr: This can maybe be a `let` binding?

XiEditor/StatusBar.swift
```diff
@@ -0,0 +1,241 @@
...
+    }
+}
+
+class StatusBar: NSView {
+
+    var currentKeys = [String]()
+    var hiddenItems = [StatusItem]()
+    var leftItems = [StatusItem]()
+    var rightItems = [StatusItem]()
+    var currentItems: [StatusItem] {
```
cmyr: instead of storing current keys and current items separately, can we just have a dictionary of keys: items?

cmyr:
Getting there! Still some design/layout nits:

### buggy stuff:

statusbar-weird-hide

Three things in this example:
- the 'f' item is initially shown even though it's clipped by the window
- adding the 'g' item causes the 'f' to be correctly hidden
- when the z item is added, the 'e' should probably be hidden; they're very close to one another.

Left padding: the left padding of the status bar item should be equal to the left padding of the buttons in the top menu bar.

screen shot 2018-06-08 at 10 14 29 am

typing this sequence causes the 'z' item to be inserted overlapping with the previous item:

screen shot 2018-06-08 at 10 17 59 am

but resizing the view even slightly correctly hides the previous item:

screen shot 2018-06-08 at 10 18 04 am


.. so there's some bug here where we aren't correctly hiding items on one side after adding them on another.

### general design stuff:

- When drawing not-unified, I think there should be a width-0 light gray line separating the statusbar from the main view. For comparison, here are three versions: without line, with line, and gradient with line:
screen shot 2018-06-08 at 11 34 24 am
screen shot 2018-06-08 at 11 33 59 am
screen shot 2018-06-08 at 11 33 18 am


And with the unified toolbar:

screen shot 2018-06-08 at 11 35 26 am
screen shot 2018-06-08 at 11 36 08 am


I think that the decision is less clear with the unified toolbar, but I still maybe prefer the line?


#### font size:

I think that the default font size may be too big?

12 pt (default)
screen shot 2018-06-08 at 11 47 29 am
11 pt
screen shot 2018-06-08 at 11 46 15 am
10 pt
screen shot 2018-06-08 at 11 47 02 am


I haven't had a chance to go through the code super carefully yet, will have to do that before we can merge. I'd like to get this in by early next week though so we can move on to hover definitions. :)

jansol:
The separator line looks good both on the unified and non-unified versions IMHO. For non-unified I like the 2nd one from the top of the "general design stuff" section.

As for font size, I generally like my fonts small, but 10pt might be a bit straining on the eyes in the long run. I currently have iTerm set to 11pt so if the rest of the statusbar can be scaled to that size I'd probably go for that. Unless you want to improve readability (especially on non-retina screens), in which case 12pt is the way to go.

nangtrongvuon:
I’ll iron out the remaining bugs. I do agree with the designs with the light gray line and the solid gray as I think the metallic gray should only be reserved for the title bar. It would also help when Mojave hits and people use the dark theme.

XiEditor/StatusBar.swift
```diff
@@ -112,32 +112,32 @@ class StatusBar: NSView {
         item.textColor = itemTextColor
         self.addSubview(item)
         currentItems[item.key] = item
-        checkItemsFitFor(windowWidth: self.superview!.frame.width)
         self.needsUpdateConstraints = true
+        checkItemsFitFor(windowWidth: self.superview!.frame.width)
```
cmyr: shouldn't `self.bounds.width` always suffice? our width should be the same as the superview's.

XiEditor/StatusBar.swift
```diff
@@ -82,11 +82,12 @@ class StatusBar: NSView {
...
     let minWidthDifference: CGFloat = 3
 
-    // Returns the minimum width required to display all items.
+    // Returns the minimum width required to display items without
+    // clipping in the status bar.
     var minWidth: CGFloat {
         return currentItems.values.filter { $0.isHidden == false }
             .map({$0.bounds.width})
-            .reduce(CGFloat(currentItems.count - 1) * statusBarPadding, +)
+            .reduce(CGFloat(currentItems.count - hiddenItems.count - 1) * statusBarPadding, +)
```
cmyr: I think this should have braces? Subtraction isn't associative, so `(currentItems.count - hiddenItems.count) - 1 != (currentItems.count - (hiddenItems.count - 1)`

(I'm sure the expected behaviour is defined in the swift spec, but braces make it easier for the code reviewer)

XiEditor/EditViewController.swift
```diff
@@ -165,16 +169,32 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
     }
 
+    func setupStatusBar() {
+        statusBar = StatusBar(frame: .zero)
+        statusBar.hasUnifiedTitlebar = unifiedTitlebar
+        self.view.addSubview(statusBar)
+
+        NSLayoutConstraint.activate([
+            statusBar.heightAnchor.constraint(equalToConstant: statusBar.statusBarHeight),
+            statusBar.widthAnchor.constraint(equalTo: editView.widthAnchor),
```
cmyr: just noticed this, but the width constraint shouldn't be necessary if you have a leading and trailing.

cmyr:
@lukakerr If you're interested, you might have an opinion on this?

cmyr:
Talked to raph yesterday and he has a bunch on his plate, so we decided it's better to get this in now than to wait for time for a second review.

As follow-up, though, I would like to have at least an option for the statusbar to not be shown if it doesn't contain any items. @nangtrongvuon could  you open an issue for this?

PR #193, opened 2018-06-06
Show result count at right of search bar

### LINKED ISSUE #192: Display number of matches in search bar
The core sends `find_status` notifications to the frontend containing information about the number of matches. This number should be displayed in the search bar, similar to this:
screen shot 2018-06-06 at 10 58 42

There is already a `// TODO` in the code where the number of matches are processed: https://github.com/google/xi-mac/blob/9cdd77524aadcb849f7712df13998042aca44ba1/XiEditor/Search.swift#L169

cmyr:
screen shot 2018-06-06 at 7 42 49 pm

fixes #192

### DIFF
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -23,6 +23,7 @@ class FindViewController: NSViewController, NSSearchFieldDelegate {
     @IBOutlet weak var doneButton: NSButton!
     @IBOutlet weak var viewHeight: NSLayoutConstraint!
 
+    let resultCountLabel = Label(title: "")
     @objc var ignoreCase = true
     @objc var wrapAround = true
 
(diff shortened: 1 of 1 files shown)


XiEditor/Search.swift
```diff
@@ -149,7 +149,10 @@ extension EditViewController {
         if term != nil {
             params["chars"] = term
         }
-        
+
+        if term ?? "" == "" {
```
nangtrongvuon: This seems very confusing to me. Perhaps if you wanted to check for empty strings, you could use `term.isEmpty`?

cmyr: Ah yea I'm trying to check if the string is nil _or_ the empty string. So maybe `(term ?? "").isEmpty` would be more clear.

PR #196, opened 2018-06-07
Handle errors returned from request RPCs

### LINKED ISSUE #188: Cannot open a text file, after trying to open a binary file
## Repro steps

1. Build `XiEditor.app` locally, start it.
2. Open a binary file thru the open menu. Nothing happens (assume it's not supported)
3. Try to open any text file.

## Expected

You should be able to open text file.

## Actually

Nothing happens when you try to open a text file.

cmyr:
Previously we were assuming that any response to a request was a success case.
In particular, this improves handling of errors in the case that core cannot
open a particular file.

closes #188

### DIFF
--- a/XiEditor/CoreConnection.swift
+++ b/XiEditor/CoreConnection.swift
@@ -19,6 +19,26 @@ import AppKit
 /// These logs can be used for profiling & debugging.
 let XI_RPC_LOG = "XI_CLIENT_RPC_LOG"
 
+/// An error returned from core
+struct RemoteError {
+    let code: Int
+    let message: String
+    let data: AnyObject?
+
+    init?(json: [String: AnyObject]) {
+        guard let code = json["code"] as? Int,
+        let message = json["message"] as? String else { return nil }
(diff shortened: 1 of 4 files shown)


XiEditor/CoreConnection.swift
```diff
@@ -19,6 +19,26 @@ import AppKit
...
+/// An error returned from core
+struct RemoteError {
+    let code: Int
+    let message: String
+    let data: AnyObject?
+
+    init?(json: [String: AnyObject]) {
+        guard let code = json["code"] as? Int,
+        let message = json["message"] as? String else { return nil }
+//        let data = json["data"] as? AnyObject? else { return nil }
```
raphlinus: Remove?

XiEditor/CoreConnection.swift
```diff
@@ -19,6 +19,26 @@ import AppKit
...
+
+        self.code = code
+        self.message = message
+        self.data = json["data"]
+    }
+}
+
+/// A completion handler for a synchrounous RPC
```
raphlinus: typo: synchronous

PR #202, opened 2018-06-11
Implement selection for find

scholtzan:
This adds selection for find functionality (<kbd>cmd</kbd> + <kbd>E</kbd>) and depends on https://github.com/google/xi-editor/pull/701

### DIFF
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -182,7 +182,7 @@ extension EditViewController {
         }
         document.sendRpcAsync("find", params: params)
     }
-
+    
     func clearFind() {
         document.sendRpcAsync("find", params: ["chars": "", "case_sensitive": false])
     }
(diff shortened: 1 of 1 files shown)


XiEditor/Search.swift
```diff
@@ -182,7 +182,7 @@ extension EditViewController {
         }
         document.sendRpcAsync("find", params: params)
     }
-
```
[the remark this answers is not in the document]
scholtzan: Oh, interesting

XiEditor/Search.swift
```diff
@@ -228,9 +228,8 @@ extension EditViewController {
             Swift.print("replaceAndFind not implemented")
 
         case .setSearchString:
-            openFind()
-            self.find(nil, caseSensitive: !findViewController.ignoreCase, regex: findViewController.regex)
-
+            document.sendRpcAsync("selection_for_find", params: ["case_sensitive": false])
```
scholtzan: Not sure whether it should be case sensitive (Sublime behaviour) or case insensitive (macOS). For now I set it to be case insensitive.

cmyr: I think Sublime uses whatever the previous setting in the find panel is/was? However I like adding this option, since it gives us more flexibility down the road.

PR #203, opened 2018-06-12
Add next to selection

scholtzan:
This adds support for <kbd>cmd</kbd> + <kbd>D</kbd> and depends on https://github.com/google/xi-mac/pull/202

### DIFF
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -203,7 +203,12 @@ extension EditViewController {
     
     @IBAction func addNextToSelection(_ sender: AnyObject?) {
         document.sendRpcAsync("selection_for_find", params: ["case_sensitive": false])
-        document.sendRpcAsync("find_next", params: ["allow_same": true, "add_to_selection": true])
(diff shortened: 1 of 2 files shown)


XiEditor/Search.swift
```diff
@@ -182,7 +182,7 @@ extension EditViewController {
         }
         document.sendRpcAsync("find", params: params)
     }
-
+
```
cmyr: whitespace

XiEditor/Search.swift
```diff
@@ -200,6 +200,16 @@ extension EditViewController {
             (findViewController.searchField as? FindSearchField)?.resultCount = resultCount
         }
     }
+    
+    @IBAction func addNextToSelection(_ sender: AnyObject?) {
+        document.sendRpcAsync("selection_for_find", params: ["case_sensitive": false])
```
cmyr: not a problem unique to this PR (we're doing this in lots of places) but in the future I'd like us to stop reaching into the document and sending RPCs directly; the document (or something similar) should instead expose a function that handles sending the RPC. (no need to fix this now)

PR #207, opened 2018-06-22
Replace

scholtzan:
This is a WIP for the replace interface. I am not very familiar with building interfaces, so feedback and tips are welcome.

This depends on https://github.com/google/xi-editor/pull/720 and https://github.com/google/xi-mac/pull/206

### DIFF
--- a/XiEditor/Main.storyboard
+++ b/XiEditor/Main.storyboard
@@ -190,11 +190,11 @@ Gw
                             <stackView distribution="fill" orientation="horizontal" alignment="centerY" horizontalStackHuggingPriority="249.99998474121094" verticalStackHuggingPriority="249.99998474121094" detachesHiddenViews="YES" translatesAutoresizingMaskIntoConstraints="NO" id="0fT-ae-quE">
                                 <rect key="frame" x="8" y="0.0" width="997" height="60"/>
                                 <subviews>
-                                    <stackView distribution="fill" orientation="vertical" alignment="leading" horizontalStackHuggingPriority="249.99998474121094" verticalStackHuggingPriority="249.99998474121094" detachesHiddenViews="YES" translatesAutoresizingMaskIntoConstraints="NO" id="lrE-3U-oZn">
-                                        <rect key="frame" x="0.0" y="4" width="832" height="52"/>
+                                    <stackView distribution="fill" orientation="vertical" alignment="leading" spacing="4" horizontalStackHuggingPriority="249.99998474121094" verticalStackHuggingPriority="249.99998474121094" detachesHiddenViews="YES" translatesAutoresizingMaskIntoConstraints="NO" id="lrE-3U-oZn">
+                                        <rect key="frame" x="0.0" y="6" width="832" height="48"/>
                                         <subviews>
                                             <searchField wantsLayer="YES" focusRingType="none" verticalHuggingPriority="750" textCompletion="NO" translatesAutoresizingMaskIntoConstraints="NO" id="yjF-1R-sp7" customClass="FindSearchField" customModule="XiEditor" customModuleProvider="target">
-                                                <rect key="frame" x="0.0" y="30" width="682" height="22"/>
+                                                <rect key="frame" x="0.0" y="26" width="682" height="22"/>
                                                 <searchFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" selectable="YES" editable="YES" borderStyle="bezel" focusRingType="none" bezelStyle="round" recentsAutosaveName="xi-editor-find" id="4a0-aC-Kre">
                                                     <font key="font" metaFont="system"/>
                                                     <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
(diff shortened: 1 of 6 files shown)


scholtzan:
The replace interface looks currently a bit out of place. Also, to actually set the replacement string, after typing it into the replace field, <kbd>return</kbd> needs to be pressed. I couldn't figure out how to send the updated replacement string to the core while typing :/. 
I am sure that the interface can be improved in many ways, but for now it can at least used for testing the replace functionality.

scholtzan:
Right now, hiding/showing the replace panel leads to misalignment of the search buttons. I'd be happy about any hints on how to fix that.

scholtzan:
There is still an issue with setting the top offset of the edit view after hiding/showing the replace panel. For some reason the offset is only correctly calculated after hitting <kbd>cmd</kbd><kbd>F</kbd> or <kbd>cmd</kbd><kbd>option</kbd><kbd>F</kbd> twice.

cmyr:
Cool, I think the overall look is totally reasonable right now; we may want to polish at some point but this looks nice, appreciate the aligned edges on the two rows of buttons.

Haven't dug into the code yet, but some initial things i notice playing around:

- replace all + undo leaves the highlights in an unexpected state (only the first result highlighted)
- (with a find query that has a match): sending `replace_all` twice in succession causes a panic in core
- the 'ignore case' menu item isn't working, and fires a Cocoa exception; this was working earlier, assume it's something in this PR? Actually maybe a bunch of these items aren't working. This is probably related to the weird way the `searchMenuTemplate` works? basically we don't create the actual search menu, instead we create a 'template' from which the menu is created for us, and (iirc) we can't wire up IBOutlets or IBActions directly, we need to manually set a tag for each menu item, which we check to figure out what item we're dealing with?
- <kbd>return</kbd> when the replace field is active should probably send `replace_next`?
- the text in the replace buttons looks different from the text in the other buttons? This is a weird one, it almost looks like they're being drawn with different anti-aliasing rules or something? 

Overall though I think this is pretty close!

scholtzan:
@cmyr: I resolved the bugs. 
In my opinion the replace buttons look just the same. Not sure, though. I went through the options and they looked the same for replace and find buttons.

cmyr:
I'm still seeing this one:
> replace all + undo leaves the highlights in an unexpected state (only the first result highlighted)

but the rest looks good. Will go through the code shortly.

scholtzan:
Oh, it seems that the highlights on the first line are correct but the following lines do not have highlights.

XiEditor/Main.storyboard
```diff
@@ -184,72 +184,172 @@ Gw
             <objects>
                 <viewController storyboardIdentifier="Find View Controller" id="lKE-xV-0EG" customClass="FindViewController" customModule="XiEditor" customModuleProvider="target" sceneMemberID="viewController">
                     <visualEffectView key="view" blendingMode="behindWindow" material="appearanceBased" state="inactive" id="NrM-ac-2ZW">
-                        <rect key="frame" x="0.0" y="0.0" width="480" height="38"/>
+                        <rect key="frame" x="0.0" y="0.0" width="1559" height="30"/>
```
cmyr: this is the nitiest of nits, but when I open the storyboard the find view _is_ really wide. On my laptop in particular this is kind of annoying; I can't fit the whole view on the screen.

XiEditor/Main.storyboard
```diff
@@ -184,72 +184,172 @@ Gw
...
-                                        </buttonCell>
-                                        <connections>
-                                            <action selector="performCustomFinderAction:" target="MxZ-GG-2K0" id="xqt-Pf-AM8"/>
-                                        </connections>
-                                    </button>
+                                    <stackView focusRingType="exterior" distribution="fill" orientation="vertical" alignment="leading" spacing="4" horizontalStackHuggingPriority="249.99998474121094" verticalStackHuggingPriority="249.99998474121094" detachesHiddenViews="YES" translatesAutoresizingMaskIntoConstraints="NO" id="oqn-cC-uXf">
+                                        <rect key="frame" x="0.0" y="0.0" width="1543" height="30"/>
+                                        <subviews>
+                                            <stackView distribution="fill" orientation="horizontal" alignment="top" horizontalStackHuggingPriority="249.99998474121094" verticalStackHuggingPriority="249.99998474121094" detachesHiddenViews="YES" translatesAutoresizingMaskIntoConstraints="NO" id="87Z-4W-agC">
```
cmyr: Is there a rationale for having the immediately nested stackview?

XiEditor/Search.swift
```diff
@@ -196,19 +232,55 @@ extension EditViewController {
...
+        if status["chars"] != nil && !(status["chars"] is NSNull) {
+            findViewController.replaceField.stringValue = status["chars"] as! String
+        }
+
+        // todo: preserve case
+    }
+
+    func replace(_ term: String?) {
+        var params: [String: Any] = [
+            "preserve_case": false,     // todo: impement option for preserving case
```
cmyr: Are you thinking you'll implement this in the near future? In general I'm reluctant to add unused params, but if it's next up then no problem. 👍

scholtzan:
@cmyr: I fixed the undo issue

scholtzan:
Regarding not everything found: is the Whole words option selected by any chance?

XiEditor/Search.swift
scholtzan: Preserving case is quite hard. I wasn't planning on working on it next, it would be future work. So, I guess I will remove it for now.

cmyr:
@scholtzan It isn't selected before I do <kbd>cmd</kbd><kbd>d</kbd>, but it becomes selected afterwards.

What was the intended behaviour with <kbd>cmd</kbd><kbd>d</kbd> and <kbd>cmd</kbd><kbd>e</kbd>, when the selection is a region and when it is a caret?

scholtzan:
<kbd>cmd</kbd><kbd>e</kbd>:
* caret: expand to word and set as search query
* selection: use selection as search query

<kbd>cmd</kbd><kbd>d</kbd>: <kbd>cmd</kbd><kbd>e</kbd> + find next whole word occurrence

Sublime behaviour is to only search for whole words when we initially start with a caret. I will see how to implement this behaviour in a future PR. The tricky part in the current state would be to set the `whole_words` flag in `Find` based on whether <kbd>cmd</kbd><kbd>d</kbd> starts with a caret or selection.

scholtzan:
Fixed the button alignments, the size of the find panel on the StoryBoard and removed the unused parameter.

cmyr:
Interesting, seems like the weird drawing of the button text has to do with the alignment, because I see it again now. I suspect by manually setting the width of the buttons we force the text to be aligned on a pixel boundary? In any case this isn't a huge issue, and I'm not sure what the best approach is. The 'magic' of NSStackView (not manually seeing all the constraints etc) makes this annoying to debug, I played around a bit now and couldn't come up with much.

I did notice that there are a bunch of storyboard attributes for various controls + some constraints that are a little messy, but it's hard to really critique that stuff in a review like this; but maybe at some point we could do a screen share and I can show you what I'm talking about?

PR #209, opened 2018-06-24
Add support for core->client new_view RPC

cmyr:
This lets us initiate view opening in the core, for things like go to definition.

It feels hacky; I think it's probably time to drop NSDocument again, but that will be a solid weekend's work.

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -228,12 +228,34 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         }
     }
 
+    // Called when using the 'Open Recent' menu. If we return false, the item
+    // is removed from that menu.
+    func application(_ sender: NSApplication, openFile filename: String) -> Bool {
+        if FileManager.default.fileExists(atPath: filename) {
+            documentController.openDocumentAsync(atUrl: URL(fileURLWithPath: filename))
+            return true
+        } else {
+            return false
+        }
+    }
+
     // MARK: - XiClient protocol
 
-    func update(viewIdentifier: String, update: [String: AnyObject], rev: UInt64?) {
-        let document = documentForViewIdentifier(viewIdentifier: viewIdentifier)
-        if document == nil { print("document missing for view id \(viewIdentifier)") }
-        document?.updateAsync(update: update)
+    func newView(viewIdentifier: String, path: String?) {
+        let fileUrl = path.map() { URL(fileURLWithPath: $0) }
+        documentController.addDocument(withIdentifier: viewIdentifier, url: fileUrl)
+    }
+
+    func update(viewIdentifier: String, update: [String: AnyObject], rev: UInt64?){
+        documentForViewIdentifier(viewIdentifier: viewIdentifier)?.updateAsync(update: update)
+    }
+
+    // Only used during development, but annoying to have to always rewrite
+    private func debugPrintUpdate(viewId: ViewIdentifier, update: [String: AnyObject], retry: Bool) {
+        let opString = (update["ops"] as! [[String: AnyObject]])
+            .map({"\($0["op"] as! String): \($0["n"] as! Int)"})
+            .joined(separator: "\n\t")
+        print("update \(viewId) \(retry ? "(retry)" : ""):\n\t\(opString)")
     }
 
     func scroll(viewIdentifier: String, line: Int, column: Int) {
@@ -358,17 +380,11 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     }
 
     //MARK: - top-level interactions
+
     @IBAction func openPreferences(_ sender: NSMenuItem) {
         let delegate = (NSApplication.shared.delegate as? AppDelegate)
         if let preferencesPath = delegate?.defaultConfigDirectory.appendingPathComponent(PREFERENCES_FILE_NAME) {
-            NSDocumentController.shared.openDocument(
-                withContentsOf: preferencesPath,
-                display: true,
-                completionHandler: { (document, alreadyOpen, error) in
-                    if let error = error {
-                        print("error opening preferences \(error)")
-                    }
-            });
+            documentController.openDocumentAsync(atUrl: preferencesPath)
         }
     }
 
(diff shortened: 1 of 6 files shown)


XiEditor/XiDocumentController.swift
```diff
@@ -105,83 +118,63 @@ class XiDocumentController: NSDocumentController {
...
+        if displayDocument {
+            document.makeWindowControllers()
+            document.showWindows()
+        }
+        return document
+    }
+
+    // this is called when documents are restored on startup. We still use the
+    // sync API for creating a new view because we're still using NSDocumentController.
+    override func makeDocument(for urlOrNil: URL?, withContentsOf contentsURL: URL, ofType typeName: String) throws -> NSDocument {
```
nangtrongvuon: I feel like saying something is an optional is another way of saying that it can be nil - I think "urlOrNil" is really unnecessary.

XiEditor/AppDelegate.swift
```diff
@@ -228,12 +228,34 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
...
+    }
+
     // MARK: - XiClient protocol
 
-    func update(viewIdentifier: String, update: [String: AnyObject], rev: UInt64?) {
-        let document = documentForViewIdentifier(viewIdentifier: viewIdentifier)
-        if document == nil { print("document missing for view id \(viewIdentifier)") }
-        document?.updateAsync(update: update)
+    func newView(viewIdentifier: String, path: String?) {
+        let fileUrl = path.map() { URL(fileURLWithPath: $0) }
```
nangtrongvuon: People write "URL" differently in code. I saw that in Rust, Url is preferred while URL is the standard for Swift/Objective-C. I think having consistency between both codebases is nice, but I generally highly prefer the use of "URL" in Swift code.

nangtrongvuon:
Also, I don't really get it - what's wrong with NSDocument? What would be a suitable alternative?

XiEditor/XiDocumentController.swift
cmyr: Yea, I agree with you. This method (on `NSDocumentController`) was introduced in the objc days, and objc naming conventions were more verbose. I wonder if perhaps the swift name was generated automatically by some conversion tool?

XiEditor/AppDelegate.swift
cmyr: Yea, this can be annoying. As a general rule I think it's more important to be consistent with the naming conventions of a given language over being consistent between languages in a single project. You're right though that, according to the Swift [API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#general-conventions) this should be URL:

>Acronyms and initialisms that commonly appear as all upper case in American English should be uniformly up- or down-cased according to case conventions:
>
>```swift
>var utf8Bytes: [UTF8.CodeUnit]
>var isRepresentableAsASCII = true
>ar userSMTPServer: SecureSMTPServer
>```

cmyr:
Nothing is wrong with NSDocument per-se, it just doesn't really suit our use-case. Because we handle all of our own file persistence in xi-core, we end up doing a lot of work overriding NSDocument methods to pretend like we're a more conventional document-based application, and that becomes more and more tricky as we want more and more custom behaviour.

The alternative is just writing our own code to handle this stuff; it would look quite a bit like the NSDocument model, but would be more adapted to our particular needs.

XiEditor/Document.swift
```diff
@@ -38,6 +38,8 @@ class Document: NSDocument {
     // minimum size for a new or resized window
     static var minWinSize = NSSize(width: 240, height: 160)
 
+    var lines = LineCache<LineAssoc>()
```
nangtrongvuon: Interestingly, I had better performance in Xi with this change than without it. Opening mod.rs from the Rust source code had no hitches, while master's version would leave me with blank lines on the first few lines.

XiEditor/Document.swift
```diff
@@ -176,9 +178,18 @@ class Document: NSDocument {
         self.sendRpcAsync("scroll", params: [first, last])
     }
 
+    /// handles the `update` RPC. This is called from a dedicated thread.
     func updateAsync(update: [String: AnyObject]) {
```
nangtrongvuon: I still feel like a reference to this function could still be in EditViewController, since it does modify the state of the EditView.

XiEditor/XiDocumentController.swift
```diff
@@ -105,83 +118,63 @@ class XiDocumentController: NSDocumentController {
...
                     (NSApplication.shared.delegate as! AppDelegate).alert(text: error.message)
                 }
             }
         }
+        return newDocument
+    }
+
+    // Called when we create a new view, or when a new view is created at startup.
+    override func openUntitledDocumentAndDisplay(_ displayDocument: Bool) throws -> NSDocument {
+        let document = try makeDocumentSync(atUrl: nil)
```
nangtrongvuon: When a new untitled view is opened (at launch), the view has no plugins available when it should have. You can reproduce this simply by opening a blank view and try to look at the plugins menu.

After digging into this, I realised this was caused by the lack of view identifier for this new document until the handling method for available_plugin has already been executed.

XiEditor/Document.swift
cmyr: You mean on current master/master you can get blank lines at the start of a document?

XiEditor/Document.swift
cmyr: There's a reason for this change, related to moving the `LineCache` into `Document`:

Basically with this change, we can have a situation where the first `update` arrives before the view controller is initialized. This wasn't possible before, because we blocked while waiting for the view_id to arrive, and so we could always finish setup before core sent `update`.

By moving the `LineCache` into the document, we can handle `update` even when there's no view controller; when the view controller actually gets initialized, the line cache will be up to date.

XiEditor/XiDocumentController.swift
cmyr: Okay cool, I had noticed sometimes the plugin menu was empty but hadn't tracked down a cause. Will look into that. 👍

XiEditor/Document.swift
nangtrongvuon: I thought that was the case, but opening those aforementioned files in master seemed okay. I might have mistaken it with some other branch?

cmyr: yes, I wonder. That was a behaviour I was worried about with this work; it can happen if we miss the first `update`, which includes the first lines. 😕

nangtrongvuon: It doesn't seem to happen to me on both my testing machines (a MBP 2014 and a MBP 2015), though perhaps it can show on a slower machine.

PR #213, opened 2018-07-02
Fix bug with building xi-syntect-plugin?

lord:
Kept getting

```
fatal error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: can't open input file: target/x86_64-apple-darwin/release/xi-syntect-plugin (No such file or directory)
```

errors until I made this change. Not sure if anybody else ran into this, but thought I'd just submit just in case. Rust 1.27, xcode 9.4.1.

### DIFF
--- a/build-rust-xcode.sh
+++ b/build-rust-xcode.sh
@@ -71,7 +71,7 @@ function build_target () {
 }
 
 build_target xi-core rust
-build_target xi-syntect-plugin rust/syntect-plugin
+build_target xi-syntect-plugin rust
 
 # move syntect plugin into plugins dir
 mkdir -p "${BUILT_PRODUCTS_DIR}/plugins/syntect/bin"


nangtrongvuon:
I didn't run into this at all - maybe check if you have cloned xi-core properly?

This change would make syntect not build at all, so I don't think this is the right fix (or if it's needed).

lord:
Yea, I've got the `xi-editor` folder copied in. You sure it would make it not build? This script for me creates a `xi-editor/rust/target/x86_64-apple-darwin/release/xi-syntect-plugin`, and the resulting app seems to be syntax highlighting. Without this change, the script attempts to open `xi-editor/rust/syntect-plugin/target/x86_64-apple-darwin/release/xi-syntect-plugin`, but the folder `xi-editor/rust/syntect-plugin/target` doesn't exist on my machine.

nangtrongvuon:
That’s weird. I develop xi-mac on 3 different Mac devices and they all work fine. Perhaps there is something in your configuration? 

@cmyr any thoughts?

cmyr:
This is likely a result of https://github.com/google/xi-editor/pull/730, which adds a workspace to the cargo project; I guess that changes what base directory is necessary for building the plugin?

lord:
Ah, yea, that sounds like it could be a problem? In a workspace, [all built files from subprojects are placed in `target` next to the root `Cargo.toml`](https://github.com/nox/rust-rfcs/blob/master/text/1525-cargo-workspace.md#detailed-design).

This is definitely not the correct change to fix this bug, since it'll run the builds twice, but I'll open another issue to track this problem.

PR #214, opened 2018-07-02
Performed small cleanups and refactors to silence warnings

nangtrongvuon:
The small cleanups performed in this patch were:

- Updating the project's build settings to XCode's recommended settings

Mostly to silence build warnings.

- Change objc inference to Default, which is recommended in Swift 4

More info on this can be seen at the [proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0160-objc-inference.md) for this change. For xi-mac, this would remove extraneous objc entry points, which has been seen to increase binary size and adversely impact load time. Since we already explicitly define objc entry points when interacting with UI elements, this change would not affect us too much.

- Removed check for C.IndexDistance == Int, which is deprecated in Swift
4 as IndexDistance are all now Ints

Since we are now on Swift 4, checking for IndexDistance type for Int is no longer [required](https://github.com/apple/swift-evolution/blob/master/proposals/0191-eliminate-indexdistance.md).

- Change addConstraint calls to changing isActive instead, recommended
by Apple

From the [addConstraint](https://developer.apple.com/documentation/uikit/uiview/1622523-addconstraint) docs:

> When developing for iOS 8.0 or later, set the constraint’s isActive property to true instead of calling the addConstraint(_:) method directly. The isActive property automatically adds and removes the constraint from the correct view.

Barring the iOS part, using isActive instead of addConstraint would avoid cases where the constraints are added to wrong views, which can happen often when working with subviews.


XiEditor/EditViewController.swift
```diff
@@ -445,7 +445,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     
     // Determines the gesture type based on flags and click count.
     private func clickGestureType(event: NSEvent) -> String {
-        let flags = event.modifierFlags.rawValue >> 16
+        _ = event.modifierFlags.rawValue >> 16
```
cmyr: I think this is a relic of how we used to send modifiers; we can just remove it now.

nangtrongvuon:
@raphlinus You meant not having them? :P

PR #221, opened 2018-07-13
Added frontend UI for Hover and Go To Definition

nangtrongvuon:
This PR adds frontend UI for hover and goto definition, both of which are LSP features implemented in [xi-editor #717](https://github.com/google/xi-editor/pull/717).

Currently, the UI elements are styled similar to XCode's way of showing quick help and definition.

To use **Hover**, _Option+Click_ on the symbol desired, or move the mouse to the symbol while holding Option. To use **Definition**, _Control+Click_ on the symbol desired.

Some screenshots of these elements can be found below:

### Hover
The hover screen does not use markdown yet. I still feel this is open to discussion.
screen shot 2018-07-13 at 09 24 47

### Definition
Definition is styled to be very similar to XCode's own definition view. Rows are highlighted on mouse hover, although the actual goto command is not implemented yet.

definition-orange
definition-green

**Feedback**
@raphlinus @cmyr @betterclever @scholtzan @jansol 

There is still a bit of work to do - such as cleaning up code here and there, or more documentation in general, though I think it is ready for some preliminary feedback.

Feedback, especially on how the elements should look, is very much appreciated. My personal opinion is the hover view could use a little more work to make it look nicer, and a Markdown library could be put in Xi, although I don't know how that will work, considering Xi's status as an open source Google project. That being said, I'm open to any and all suggestions.


jansol:
What's going on with the ^ in the definition screenshots?

nangtrongvuon:
It's just part of the popover, the arrow's origin is where the cursor was. MacOS's screenshot doesn't capture the cursor, so it's not very clear.

jansol:
I mean the border looks weird in the screenshots. Like the draw order is wrong? It looks right in the hover screenshot.

jansol:
Oh and please align the point of the ^ with the baseline of the text if possible, instead of centering it on the line. That way it's still clear what it points to but it doesn't obscure the word needlessly.

nangtrongvuon:
The border seems to be a screenshot thing where the popover shadow doesn’t get rendered. I might take other screenshots that aren’t window shots.

Gotcha on the line thing.

cmyr:
Very cool that this is working!

I noticed that the hover view seems to clip the definition if the definition is too long. We should probably either resize the view in this case, or maybe allow scrolling? What do you think?

cmyr:
Also I see what I think @jansol was pointing out, which is that in the `goto` view, the popup view draws a shadow over top of the arrow of the callout?

nangtrongvuon:
I’ll have to check again, but I don’t think it looks like that normally? I took the screenshot as a window specific screenshot, so maybe that could be why.

cmyr:
I was testing manually when I saw this.

nangtrongvuon:
Got it, I’ll check that too.

nangtrongvuon:
Both of the aforementioned issues were fixed. You can see now the popover views look like now in the updated main PR post.

nangtrongvuon:
I'll close this PR, since this work will be split in two parts - Hover and Definition.

Issue #194, opened 2018-06-06
find bar looks out of place when unified_toolbar = true


cmyr:
We should probably be styling the find bar as well, in this case, or else we look like this:

![screen shot 2018-06-06 at 7 44 33 pm](https://user-images.githubusercontent.com/3330916/41070670-16ec1dce-69c2-11e8-9a72-b8af1bdfb0e7.png)
![screen shot 2018-06-06 at 7 44 40 pm](https://user-images.githubusercontent.com/3330916/41070672-194e0f50-69c2-11e8-9122-f3fd186c5312.png)

jansol:
Looks like it's broken again, now the find bar appears to be transparent...

jansol:
And fixed again in 8901cdc818f9dda498a98a4b0817322bcfb8267f

PR #223, opened 2018-07-13
Match search panel color to theme

ryanbloom:
Here's my attempt at #194. As far as I can tell, there's no way to customize the background color of an `NSVisualEffectView`, so I changed it to a plain `NSView`. Now the search panel matches the theme's background color whenever the `unified_titlebar` preference is enabled.

Screenshots:
findbar_solarized_light
findbar_solarized_dark

Without `unified_titlebar`, it defaults to the standard whitish-gray color:
findbar

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -147,6 +147,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
                 window.backgroundColor = unifiedTitlebar ? color : nil
 
                 statusBar.updateStatusBarColor(newBackgroundColor: self.theme.background, newTextColor: self.theme.foreground, newUnifiedTitlebar: unifiedTitlebar)
+                findViewController.updateColor(newBackgroundColor: self.theme.background, unifiedTitlebar: unifiedTitlebar)
 
                 if color.isDark && unifiedTitlebar {
                     window.appearance = NSAppearance(named: NSAppearance.Name.vibrantDark)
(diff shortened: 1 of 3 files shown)


cmyr:
Cool!

I think it would be nice to keep the current appearance when we _aren't_ doing the unified toolbar. One approach could be to have a normal view on top of the visual effects view, with a specific color, but to only make it visible when `unified_toolbar = false`? Or the other way around?

ryanbloom:
On my computer, the `NSVisualEffectView` had a solid whitish-gray background whenever `unified_toolbar` was false. Was it supposed to look different, or are you thinking 2 views would be better implementation-wise?

cmyr:
Ah weird I thought the `NSVisualEffectView` was being used to set a `.dark` appearance when the theme was dark. Now I'm not sure what it's doing there. 😕

ryanbloom:
The `NSVisualEffectView` has a dark appearance when the theme is dark *and* `unified_titlebar` is true (see [here](https://github.com/google/xi-mac/blob/master/XiEditor/EditViewController.swift#L151)). Without `unified_titlebar`, it looks like screenshot 3 above.

If we want to adopt dark mode in macOS Mojave, I suppose the `NSVisualEffectView` could make that easier. We'd keep using the theme color when `unified_titlebar` is true, but there are a few ways to handle the non-`unified_titlebar` case:

1. Use a hardcoded color to mimic the current appearance (what I did in this PR) and later add another color for dark mode
2. Have a separate `NSVisualEffectView` that shows only when `unified_toolbar` is false (maybe easier to adapt to dark mode?)
3. Use the NSColor constant `controlBackgroundColor` for the find bar, which is white in light mode and presumably gray/black in dark mode (the `NSVisualEffectView` has a background of RGB 246,246,246)

What do you think?

PR #225, opened 2018-07-14
Scroll to the the old visible area after find bar is shown

### LINKED ISSUE #224: Enhancement request on Find and Replace
Glad to see xi-mac has implemented basic Find and Replace functionality! The general experience is good, but I still think there could be an enhancement.

That is, when I press cmd+F or option+cmd+F, I find that the Find&Replace bar covers the first and second line text as shown below:
1. original view
original view
2. after open Find bar
after open Find
3. after open Find&Replace bar
after open Find&Replace bar

Even I can scroll down two lines to view the 1st and 2nd lines, I still think that the text should scroll down automatically 1 or 2 lines when I am searching or replacing, since this is implemented in other macOS apps like TextEdit.app and Terminal.app:
1. original view
original view
2. after open Find&Replace bar
image

1. original view
original view
2. after open Find bar
after open Find bar

I would be grateful if you can make this improvement to xi-mac!

tsekityam:
This PR attempt to Fix #224 by scroll up to existing visible area after find bar is shown

### DIFF
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -183,6 +183,10 @@ extension EditViewController {
             findViewController.view.isHidden = true
             (findViewController.searchField as? FindSearchField)?.resultCount = nil
             scrollView.contentInsets = NSEdgeInsetsZero
+            
+            let offset = findViewController.view.fittingSize.height
+            let origin = scrollView.contentView.visibleRect.origin
+            scrollView.contentView.scroll(to: NSMakePoint(origin.x ,origin.y + offset))
         }
 
         editView.window?.makeFirstResponder(editView)
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -163,17 +163,18 @@ class FindViewController: NSViewController, NSSearchFieldDelegate, NSControlText
 
 extension EditViewController {
     func openFind(replaceHidden: Bool) {
+        let offset = findViewController.view.fittingSize.height
+
         if findViewController.view.isHidden {
             findViewController.view.isHidden = false
+            let origin = scrollView.contentView.visibleRect.origin
+            scrollView.contentView.scroll(to: NSMakePoint(origin.x ,origin.y - offset))
             document.sendRpcAsync("highlight_find", params: ["visible": true])
         }
 
         findViewController.replacePanel.isHidden = replaceHidden
-        let offset = findViewController.view.fittingSize.height
         scrollView.contentInsets = NSEdgeInsetsMake(offset, 0, 0, 0)
 
-        let origin = scrollView.contentView.visibleRect.origin
-        scrollView.contentView.scroll(to: NSMakePoint(origin.x ,origin.y - offset))
 
         editView.window?.makeFirstResponder(findViewController.searchField)
     }
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -175,7 +175,6 @@ extension EditViewController {
         findViewController.replacePanel.isHidden = replaceHidden
         scrollView.contentInsets = NSEdgeInsetsMake(offset, 0, 0, 0)
 
-
         editView.window?.makeFirstResponder(findViewController.searchField)
     }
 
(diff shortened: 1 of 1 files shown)


cmyr:
Sorry, lost track of this PR while discussing in the issue. Will take a look tomorrow!

XiEditor/Search.swift
```diff
@@ -171,6 +171,10 @@ extension EditViewController {
         findViewController.replacePanel.isHidden = replaceHidden
         let offset = findViewController.view.fittingSize.height
         scrollView.contentInsets = NSEdgeInsetsMake(offset, 0, 0, 0)
+
+        let origin = scrollView.visibleRect.origin
```
cmyr: `scrollView.visibleRect` is just the frame of the scrollView; you probably want `scrollView.contentView.visibleRect`?

As is this only works at the top of the file.

XiEditor/Search.swift
```diff
@@ -171,6 +171,10 @@ extension EditViewController {
...
+        scrollView.contentView.scroll(to: NSMakePoint(origin.x ,origin.y - offset))
```
cmyr: This correctly adjusts the offset when opening the find view, but it doesn't adjust it back when the find view is closed; this means repeatedly opening the find view and closing it will push the scroll further and further down. :)

XiEditor/Search.swift
tsekityam: Fixed

XiEditor/Search.swift
tsekityam: Fixed by add the `offset` back to the `origin.y` when find bar is closed

XiEditor/Search.swift
```diff
@@ -163,17 +163,18 @@ class FindViewController: NSViewController, NSSearchFieldDelegate, NSControlText
 
 extension EditViewController {
     func openFind(replaceHidden: Bool) {
+        let offset = findViewController.view.fittingSize.height
+
         if findViewController.view.isHidden {
             findViewController.view.isHidden = false
+            let origin = scrollView.contentView.visibleRect.origin
+            scrollView.contentView.scroll(to: NSMakePoint(origin.x ,origin.y - offset))
```
cmyr: I don't think I understand this change. Was there a problem you were trying to solve?

This now creates a new issue if you open the 'replace' view while the find view is open; we don't correctly adjust to the height of the replace view.

XiEditor/Search.swift
```diff
@@ -163,14 +163,26 @@ class FindViewController: NSViewController, NSSearchFieldDelegate, NSControlText
...
+        let replaceHiddenChanged = findViewController.replacePanel.isHidden != replaceHidden
+
+        if !findViewController.view.isHidden && replaceHiddenChanged {
+            let offset = findViewController.view.fittingSize.height
+            let origin = scrollView.contentView.visibleRect.origin
+            scrollView.contentView.scroll(to: NSMakePoint(origin.x ,origin.y + offset))
         }
 
         findViewController.replacePanel.isHidden = replaceHidden
         let offset = findViewController.view.fittingSize.height
```
cmyr: we declare `offset` and `origin` twice. Maybe we can just declare them once at the top of the fn?

tsekityam: Although we declare them twice, they are declared in two different functions, `openFind()` and `closeFind()`. If we want to share them between two functions, then we should extract these variables to `EditViewController` class.

Am I understand correctly?

tsekityam: Done

tsekityam: `findViewController.replacePanel.isHidden = replaceHidden` may affect the value of `findViewController.view.fittingSize.height` and `scrollView.contentView.visibleRect.origin`. If replace bar is gone, then the height of the the entire find bar will be changed; similarly, the origin will be different too. 

As a result, we have to assign these values twice.

cmyr: oookay, thank you for the explanation, I was confused. Sorry!

Let's go back to how you had it before this change, and then we can get that merged.

This logic will also need to be updated when #229 is merged.

cmyr: Ah sorry I missed this comment.

All that I meant was that my earlier advice to declare `offset` and `origin` once, at the top, was because I did not understand that those values changed at various times in the function.

Now that I understand, I think that you were correct the first time, and that my advice made the code worse. 🙇 

This is not a big problem, though, and I will merge the current version in a few days.

Sorry for the confusion, and thank you for the patch! 🙏

PR #226, opened 2018-07-15
Use 'paste' RPC

cmyr:
This adds frontend support for the paste RPC added in https://github.com/google/xi-editor/pull/747.

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -566,8 +566,8 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     // may be used for IME?
     override func insertText(_ insertString: Any) {
         let text: String
-        if let s = insertString as? NSString {
-            text = s as String
+        if insertString is NSString {
+            text = insertString as! String
         } else if let s = insertString as? NSAttributedString {
             text = s.string as String
         } else {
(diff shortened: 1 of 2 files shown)


XiEditor/EditViewController.swift
```diff
@@ -516,10 +516,19 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
     }
-    
-    // used mostly for paste
+
+    // NOTE: this was previously used for paste, and could possibly be removed, but it
+    // may be used for IME?
     override func insertText(_ insertString: Any) {
-        document.sendRpcAsync("insert", params: insertedStringToJson(insertString as! NSString))
+        let text: String
+        if let s = insertString as? NSString {
```
nangtrongvuon: You can write this check as `insertString is NSString` and it would look a bit more verbose.

XiEditor/EditViewController.swift
```diff
@@ -516,10 +516,19 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
+            text = s as String
+        } else if let s = insertString as? NSAttributedString {
```
nangtrongvuon: Ditto here.

cmyr: Hmm, I don't think it works here; I need access to the `.string` method on `NSAttributedString`, I don't think I can just cast an `NSAttributedString` to a `String`.

PR #228, opened 2018-07-18
Added hover UI for front end

nangtrongvuon:
This is the hover part of #221. This PR adds the hover UI to Xi to work with the LSP's hover feature.

In order to use hover, **Option+Click** on the desired symbol.
This PR can be tested using [my dummy plugin](https://github.com/nangtrongvuon/xi-editor/tree/dummy-hover-plugin), or with xi-editor's [#753](https://github.com/google/xi-editor/pull/753).

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -550,7 +550,6 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         if let event = hoverEvent {
             let hoverPosition = editView.bufferPositionFromPoint(event.locationInWindow)
             document.sendRpcAsync("request_hover", params: ["request_id": hoverRequestID, "position": ["line": hoverPosition.line, "column": hoverPosition.column]])
-            hoverRequestID += 1
         }
     }
 
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -461,6 +461,14 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     }
 
     //MARK: Other system events
+    override func flagsChanged(with event: NSEvent) {
+        if event.modifierFlags.contains(.option) {
+            NSCursor.crosshair.set()
+        } else {
+            NSCursor.arrow.set()
+        }
+    }
+
     override func keyDown(with theEvent: NSEvent) {
         self.editView.inputContext?.handleEvent(theEvent);
     }
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -182,9 +182,6 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     var hoverRequestID = 0
 
     override func viewDidLoad() {
-        if let path = ProcessInfo.processInfo.environment["PATH"] {
-            print(path)
-        }
         super.viewDidLoad()
         shadowView.wantsLayer = true
         editView.dataSource = self
@@ -514,9 +511,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         let gestureType = clickGestureType(event: theEvent)
 
         if gestureType == "request_hover" {
-            document.sendRpcAsync("request_hover", params: ["request_id": hoverRequestID, "position": ["type": "utf8_line_char", "line": position.line, "character": position.column]])
-            hoverEvent = theEvent
-            hoverRequestID += 1
+            sendHover()
         } else {
             document.sendRpcAsync("gesture", params: [
                 "line": position.line,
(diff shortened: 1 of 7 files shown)


cmyr:
Okay playing around this seems excellent! The only visual issue I see is that a second option+click while the hover view is visible causes a jerky animation.

betterclever:
It is a but hard to describe it here. Probably it is  best to do a screenshare where I can show you what is happening in my computer.

XiEditor/EditViewController.swift
```diff
@@ -505,19 +539,41 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
-    
+
     override func mouseUp(with theEvent: NSEvent) {
         dragTimer?.invalidate()
         dragTimer = nil
         dragEvent = nil
     }
-    
+
+    override func mouseMoved(with theEvent: NSEvent) {
```
cmyr: is this supposed to do anything? I can't seem to get it to work.

XiEditor/EditViewController.swift
```diff
@@ -505,19 +539,41 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
+        hoverEvent = theEvent
+    }
+
+    @objc func sendHover() {
+        if !infoPopover.isShown {
+            if let event = hoverEvent {
+                let hoverPosition = editView.bufferPositionFromPoint(event.locationInWindow)
+                hoverTimer?.invalidate()
+                hoverTimer = nil
+                document.sendRpcAsync("request_hover", params: ["request_id": hoverRequestID, "position": ["type": "utf8_line_char", "line": hoverPosition.line, "character": hoverPosition.column]])
```
cmyr: If we're going to use this in two places, I would add a method to `Document` that handles the implementation.

I believe that the RPC expected in core has changed as well; we don't send `type` and we send `column` instead of `character`.

XiEditor/EditViewController.swift
```diff
@@ -480,21 +505,30 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
-            "ty": clickGestureType(event: theEvent)
-        ])
-        
+
+        let gestureType = clickGestureType(event: theEvent)
+
+        if gestureType == "request_hover" {
+            document.sendRpcAsync("request_hover", params: ["request_id": hoverRequestID, "position": ["type": "utf8_line_char", "line": position.line, "character": position.column]])
+            hoverEvent = theEvent
+            hoverRequestID += 1
```
cmyr: we don't seem to actually ever use this anywhere. Perhaps it's unnecessary?

XiEditor/HoverViewController.swift
```diff
@@ -0,0 +1,120 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Cocoa
+
+class HoverView: NSTextView {
+    // Lets AppKit do all the work when setting up a new text view
+    override init(frame frameRect: NSRect, textContainer container: NSTextContainer?) {
+        let textView = NSTextView(frame: .zero)
+        super.init(frame: frameRect, textContainer: textView.textContainer)
```
cmyr: this override doesn't seem to actually do anything. What happens if we just leave it out?

XiEditor/HoverViewController.swift
```diff
@@ -0,0 +1,120 @@
...
+    lazy var scrollView: NSScrollView = {
+        let scrollView = NSScrollView(frame: NSRect(x: 0, y: 0, width: hoverPopoverWidth, height: 0))
+        scrollView.borderType = .noBorder
+        scrollView.hasVerticalScroller = true
+        scrollView.autoresizingMask = [.height]
+        return scrollView
+    }()
+    var resultContent: String
+    var hoverView: HoverView
+    let hoverPopoverWidth: CGFloat = 500 // XCode size for quick help popovers
```
cmyr: I think we should probably have a minimum and a maximum width, and then we resize ourself to the preferred width of our text, within those constraints.

XiEditor/EditViewController.swift
nangtrongvuon: It is used for when an Option+Click is detected.

XiEditor/HoverViewController.swift
nangtrongvuon: I tried this, and there is a whole host of unintended behaviors when I tried to get the view to size itself to fit all the text perfectly. Through much trial and error I and lots of reading I realized many macOS popovers (not just in XCode) works best when they have a fixed width or fixed height, because it's much easier to calculate how much space text would take in one direction. 

Also, I think it looks a bit more aesthetically more pleasing this way. :)

XiEditor/EditViewController.swift
nangtrongvuon: This is the one part that I don't think this implementation would do justice. In order to get us a real hover, some more work that hooks on text is required. For now, Option+Click is the most reliable option when trying to get a hover to work.

XiEditor/EditViewController.swift
```diff
@@ -505,19 +530,29 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
         dragTimer?.invalidate()
         dragTimer = nil
         dragEvent = nil
     }
-    
+
+    @objc func sendHover() {
+        if let event = hoverEvent {
+            let hoverPosition = editView.bufferPositionFromPoint(event.locationInWindow)
+            hoverTimer?.invalidate()
```
betterclever: Compilation is failing for me on these two lines. Looking more closely, I see it is not declared anywhere.

betterclever: Also, after removing these lines, hover does not work. Am I missing something?

XiEditor/HoverViewController.swift
```diff
@@ -0,0 +1,127 @@
...
+        let hoverContentSize = NSSize(width: hoverViewController.hoverPopoverWidth, height: hoverViewController.heightForContent())
+
+        guard !hoverContent.isEmpty else { return }
+
+        hoverViewController.scrollView.setFrameSize(hoverContentSize)
+        hoverViewController.scrollView.documentView?.setFrameSize(hoverContentSize)
+        infoPopover.contentSize = hoverContentSize
+        infoPopover.contentViewController = hoverViewController
+
+        if let event = hoverEvent {
```
cmyr: along with checking for `hoverEvent`, we should also make sure that the `request_id` we get back is the same is current. Otherwise we can have a situation where we request hover multiple times, and then the first result arrives and is displayed even though it isn't what we want.

XiEditor/HoverViewController.swift
```diff
@@ -121,6 +121,7 @@ extension EditViewController {
             let positioningSize = CGSize(width: 1, height: 1) // Generic size to center popover on cursor
 
             infoPopover.show(relativeTo: NSRect(origin: positioningPoint, size: positioningSize), of: self.view, preferredEdge: .minY)
+            hoverRequestID += 1
```
cmyr: Doesn't this mean multiple requests could be sent with the same ID?

nangtrongvuon: Yes, but I'm not quite sure what the request ID is meant to represent in the first place. As it is right now, the ID doesn't get incremented if a result wasn't shown, which happens when:

1. The result we get back is empty
2. The request ID in the RPC notification we get from core is different from what's currently expected

Since (in my personal opinion) I think that hover requests should return results instantly, it's fine even if we send many requests with the same ID?

XiEditor/HoverViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+
+extension EditViewController {
+
+    // Puts the popover at the baseline of the chosen hover symbol.
+    func showHover(withResult result: [String: AnyObject]) {
+        if infoPopover.isShown {
+            infoPopover.performClose(self)
+        }
+
+        let hoverContent = result["content"] as! String
```
betterclever: As per latest work in https://github.com/google/xi-editor/pull/753 , the result is now just a String, so you don't need to extract `content` out here.

PR #229, opened 2018-07-19
Multiple Search Queries

scholtzan:
This PR depends on https://github.com/google/xi-editor/pull/745 and adds support for multiple search queries.

~~I am trying to dynamically add additional search fields. To do this, currently the existing search field  including the search menu is copied and added into the search panel. 
I am not sure if that is a good approach but I couldn't come up with a better solution, yet. One problem with this is that if a search option in the search menu is selected/deselected it is not possible to determine in which search menu the option was selected. So, the search query options cannot be updated accordingly.~~

~~I'd appreciate any suggestions/hints/feedback on what would be a good way to add more search fields dynamically.~~

## Todo/Done

- [x] Dynamically add new search fields
- [x] Support use selection for find with multiple search queries
- [x] One search field by default present
- [x] `wrap_around`
- [x] Clearing search fields
- [x] Update search count
- [x] tab should focus next search field (`nextKeyView`)
- [x] Adjust top offset after adding/removing search fields (https://github.com/google/xi-mac/pull/225/)
- [x] Removing search fields
- [x] Fix `find_next` correct selection of current match
- [x] Colorization
    - [x] Nicer colors
- [x] Fix replace
- [x] Maximum queries
- [x] Cleanup, Tests, Documentation (remove/delete)
- [x] Add/delete button inline 
- [x] Load highlight colors from theme
- [x] Options for showing/hiding multiple search query feature
- [x] Hide/disable delete and add buttons
- [x] Find next fix
- [x] Fix ignore case
- [x] Add search field after search field in which + was pressed
- [x] +/- styling → outside
- [x] find next/previous bottom alignment

## Future Ideas
* Background color of search field same as highlight
* Avoid change of highlight color when search field is added/removed
* Colors for find highlights set in themes

## Todo

* Behaviour of "Replace all"
    * currently overlapping matches are considered as one and replace by replacement string

### DIFF
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -28,6 +28,7 @@ class FindViewController: NSViewController, NSSearchFieldDelegate, NSControlText
 
     var searchQueries: [SuplementaryFindViewController] = []
     var wrapAround = true   // option same for all search fields
+    var showMultipleSearchQueries = false   // activates/deactives 
 
     override func viewDidLoad() {
         addSearchField(true)     // by default at least one search query is present
@@ -87,6 +88,7 @@ class FindViewController: NSViewController, NSSearchFieldDelegate, NSControlText
             newSearchFieldController.disableRemove = disableRemove
             searchFieldsStackView.insertView(newSearchFieldController.view, at: searchQueries.count - 1, in: NSStackView.Gravity.center)
             newSearchFieldController.searchField.becomeFirstResponder()
+            (newSearchFieldController.searchField as! FindSearchField).showInlineButtons(show: (newSearchFieldController.parentFindView?.showMultipleSearchQueries)!)
 
             searchFieldsNextKeyView()
 
@@ -407,6 +409,13 @@ extension EditViewController {
         document.sendRpcAsync("selection_for_replace", params: [])
     }
 
+    @IBAction func multipleSearchQueries(_ sender: AnyObject?) {
+        findViewController.showMultipleSearchQueries = !findViewController.showMultipleSearchQueries
+        for query in findViewController.searchQueries {
+            (query.searchField as! FindSearchField).showInlineButtons(show: findViewController.showMultipleSearchQueries)
+        }
+    }
+
     @IBAction func performCustomFinderAction(_ sender: Any?) {
         guard let tag = (sender as AnyObject).value(forKey: "tag") as? Int,
             let action = NSTextFinder.Action(rawValue: tag) else { return }
@@ -478,6 +487,8 @@ class FindSearchField: NSSearchField {
     let defaultButtonWidth: CGFloat = 22
     let inlineButtonSpacing: CGFloat = 2
     var parentFindView: SuplementaryFindViewController? = nil
+    var inlineButtons: [NSButton] = []
+    var labelConstraint: NSLayoutConstraint? = nil
 
     private var _lastSearchButtonWidth: CGFloat = 22 // known default
     var id: String? = nil
@@ -508,7 +519,8 @@ class FindSearchField: NSSearchField {
         label.textColor = NSColor.lightGray
         label.font = NSFont.systemFont(ofSize: 12)
         label.translatesAutoresizingMaskIntoConstraints = false
-        label.rightAnchor.constraint(equalTo: self.rightAnchor, constant: -defaultButtonWidth - rightPadding).isActive = true
+        labelConstraint = label.rightAnchor.constraint(equalTo: self.rightAnchor, constant: -defaultButtonWidth - rightPadding)
+        labelConstraint?.isActive = true
         label.centerYAnchor.constraint(equalTo: self.centerYAnchor).isActive = true
     }
 
@@ -533,6 +545,23 @@ class FindSearchField: NSSearchField {
         button.bottomAnchor.constraint(equalTo: self.bottomAnchor, constant: -inlineButtonSpacing).isActive = true
         button.widthAnchor.constraint(equalToConstant: defaultButtonWidth).isActive = true
         rightPadding += defaultButtonWidth + inlineButtonSpacing
+        inlineButtons.append(button)
+    }
+
+    func showInlineButtons(show: Bool) {
+        rightPadding = 0
+        for button in inlineButtons {
+            button.isHidden = !show
+
+            if show {
+                rightPadding += (defaultButtonWidth + inlineButtonSpacing) * CGFloat(inlineButtons.count - 1)
+            }
+        }
+
+        labelConstraint?.isActive = false
+        labelConstraint = label.rightAnchor.constraint(equalTo: self.rightAnchor, constant: -defaultButtonWidth - rightPadding)
+        labelConstraint?.isActive = true
+        self.updateCell(self.cell!)
     }
 
     // required override; on 10.13(?) accessory icons aren't
@@ -558,7 +587,7 @@ class FindSearchField: NSSearchField {
         let rect = super.rectForSearchText(whenCentered: isCentered)
         let delta = max(0, _lastSearchButtonWidth - rect.origin.x)
         let originX = rect.origin.x + delta
-        var width = label.frame.minX - originX
+        var width = rect.width - originX - rightPadding - defaultButtonWidth
 
         // the label's frame is at 0,0 the first time we open this view
         if width < 0 {
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -469,6 +469,9 @@ class Label: NSTextField {
 
 class FindSearchField: NSSearchField {
     let label = Label(title: "")
+    var rightPadding: CGFloat = 0    // tracks how much space on the right side of the search field is occupied by buttons
+    let defaultButtonWidth: CGFloat = 22
+    let inlineButtonSpacing: CGFloat = 2
 
     private var _lastSearchButtonWidth: CGFloat = 22 // known default
     var id: String? = nil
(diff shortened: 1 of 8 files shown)


scholtzan:
This is ready for review. 
Currently the colors are hardcoded. I also decided to wait for https://github.com/google/xi-mac/pull/225 to get merged and then adapt it to my changes.

@cmyr @raphlinus

scholtzan:
This is ready for review. 
One thing that might need some future consideration is allowing users to define find highlight colors in the theme file. Currently, it is not possible to define custom rules in theme files since only a specified set of rules is parsed from the file. All custom rules will just be ignored. Allowing customization would however mean to change the syntect plugin, which is not really elegant in my opinion.

nangtrongvuon:
I don't see anything particularly wrong with the code, but I have a few comments:

I think the spacing of the arrow boxes and the replace box should be equal. I found that if you made the Equal Widths constant 8 for the Stack View and the Replace Stack View, the 2 buttons become equally spaced.
nh ch p man hinh 2018-09-09 luc 10 42 51

Also, you can use a centerY anchor on the replace stack view to silence the XCode storyboard warning. I don't think it will break anything and it will silence the red warning from the storyboard.

Other than that, excited to use multiple searches in Xi! It's particularly helpful when looking through logs and such, and I can't seem to find another editor that has this feature.

scholtzan:
Thanks for that! I didn't notice that the sizes of the text boxes were different, but fixed it now. 
I am not sure about the bug. If I change the `+` to a `-` then it scrolls in the opposite direction when the find bar is closed. For me it is working correctly as it is, but perhaps I am missing some corner case?

cmyr:
Okay, preliminary thoughts:

- there's something funky at the right side of the find text field:
render_bug
- with multiple find views, the add/remove buttons aren't styled correctly when `unifiedTitlebar = true`:
screen shot 2018-09-24 at 5 27 56 pm

- <kbd>tab</kbd> should toggle between the find and replace fields if visible? Apparently this isn't a new behaviour, so maybe can be a separate issue.

- the replace field can be the wrong color with dark themes?
screen shot 2018-09-24 at 5 30 57 pm

scholtzan:
Good points! I always had `unifiedTitlebar = false` while testing.

cmyr:
yea, there's some complexity here, feels sort of related to the config setting stuff actually.

My general impression, at this point:

- I appreciate that this is now not presented by default. This feels like a good choice for a 'power user' feature.
- It's confusing that, while I can search for multiple queries, I can't selectively go next/prev within a single query. That is, I can't chose to jump to the next instance of _this_ query, it's always the next instance of _a_ query.
- This seems to adds a reasonable amount of complexity to the protocol and to the implementation of search, which I worry will make things more difficult even for clients that might not choose to use this feature. (I'm not sure how legitimate this concern is; currently just a thought). I think something that's missing, currently, is some high-level documentation of how the search system works, and how a client should implement it.


I still need to look over the code more closely, but if we get the ux stuff cleaned up I think it should be reasonable to merge this. If it ends up being a big maintenance burden at some point in the future, we can reconsider then, but I think this is a reasonable addition in the 'experimental text editor' vein. I guess something else to think about is, if we do end up removing this at some point, how much work is involved in that? But I don't think that's a hugely important question, and we can deal with it down the road.


And a comment more appropriate for the xi-editor repo: I  would like to see more benchmarks and profiling of this code as well, it's definitely performance sensitive.

scholtzan:
I agree, that it might be useful to be able to jump to the next occurrence of a specific query. I am less concerned about the complexity that would add to the protocol (might be doable by adding an optional parameter for the query to `find_next`/`find_previous`) but more about the complexity it would possibly add to the UI. I imagine we'd have "Find next/Find previous/Find all" for each query then and additionally "Find next/Find previous/Find all" for all queries which might get confusing. Or maybe there are be better ways to provide that functionality in the UI? I am not really an UI expert... But that could also be follow-up work at some point.

XiEditor/Search.swift
```diff
@@ -86,139 +249,117 @@ class FindViewController: NSViewController, NSSearchFieldDelegate, NSControlText
...
+    func disableAddButton(disable: Bool) {
+        addButton.isEnabled = !disable
+    }
+
+    func disableDeleteButton(disable: Bool) {
+        deleteButton.isEnabled = !disable
+    }
+
+    @IBAction func addSearchField(_ sender: NSButton) {
+        let offset = parentFindView?.view.fittingSize.height
```
nangtrongvuon: Small bug here: When multiple queries are present and you paste in something, the edit view gets hidden.

I suspect it's because the scroll view's origin isn't set properly after the paste, or it doesn't account for the presence of the find view. 
screen shot 2018-09-27 at 09 29 34

scholtzan:
Thanks! Yes, currently having different colors is more of a hack. I would really like to be able to define the highlight colors in the theme, however syntect currently does not support that. I was planning to perhaps work on this  in the future, also to allow more customization of the front-end in general. I'll probably open a separate issue for this at some point, since I am not sure what the best way is to achieve this.

vlovich:
Exciting to see this coming together.  Not sure if the implementation already has it, but one thing that would put this over the top is saved queries.  This would make the feature a killer for log analysis.  Teams can have a predefined set of log queries checked in somewhere so the log experts can make even non-experts implicitly more efficient at analyzing logs.

I'm thinking something like the saved queries folder has a bunch of JSON (YAML?) files that represent available queries (can walk the tree for these files in case users want to add a folder structure).  These are then auto-added to the UI so you can easily just switch between the queries (or even assign shortcuts to your favourite queries).  Them you just symlink whichever JSON files you want into the saved queries folder to make them available in the editor (or symlink a folder of JSON files).

Anyway, just an idea I had.

ryanbloom:
One little UI issue: if you add some search fields, then disable Multiple Queries, the extra fields stay around without buttons to delete them. On the other hand, if the user disabled it by accident and the queries disappeared, that could be annoying. Maybe the ideal solution is to remember the queries somewhere in case the the user re-enables it?

Also, on the Multiple Search Queries menu item:
* It should probably indicate when the feature is enabled (`menuItem.state = .on`)
* If a menu item ends in "...", I expect it to open a panel or dialog or something. I don't think the ellipsis makes sense here, but maybe that's just me.

XiEditor/Search.swift
```diff
@@ -233,44 +374,63 @@ extension EditViewController {
...
     
     func findStatus(status: [[String: AnyObject]]) {
-        if status.first?["chars"] != nil && !(status.first?["chars"] is NSNull) {
-            findViewController.searchField.stringValue = status.first?["chars"] as! String
-        }
-        
-        if status.first?["case_sensitive"] != nil && !(status.first?["case_sensitive"] is NSNull) {
-            findViewController.ignoreCase = status.first?["case_sensitive"] as! Bool
-        }
+        for statusQuery in status {
```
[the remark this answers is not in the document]
scholtzan: I agree it's not very clean. I tried out a few things but they are all not really better than this :/

scholtzan:
@cmyr Regarding <kbd>tab</kbd>: currently if there are multiple search fields it switches to the next search field or if the cursor is in the last search field then it switches to the replace field (if shown). I thought that this behaviour is more intuitive compared to always switching to the replace field from any search field.

Apart from this, I think I addressed all the other issues that were mentioned.

cmyr:
@scholtzan okay great, will take a look shortly, actually, I promise!

XiEditor/Search.swift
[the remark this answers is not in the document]
cmyr: Well maybe we can at least document what's going on and what the expected structure of the input data is? Let's try and think of the poor soul who is going to want to modify this code down the road.. 😢

XiEditor/Search.swift
```diff
@@ -21,13 +21,169 @@ extension NSColor {
...
+            searchFieldsButtonsState()
+
+            return newSearchFieldController.searchField as? FindSearchField
+        }
+
+        searchFieldsButtonsState()
+        return nil
+    }
+
+    // Disables/enables add and delete buttons depending on whether some conditions are fulfilled.
```
cmyr: all of these `//` can be `///` maybe?

XiEditor/Search.swift
```diff
@@ -21,13 +21,169 @@ extension NSColor {
 }
 
 class FindViewController: NSViewController, NSSearchFieldDelegate, NSControlTextEditingDelegate {
-    weak var findDelegate: FindDelegate!
+    var findDelegate: FindDelegate!
```
cmyr: be aware of the problem addressed in #235; if this introduces a reference cycle we can end up leaking memory. cc @ryanbloom, thoughts?

ryanbloom: @cmyr The leak isn't happening on master. In this PR, I can find 5 classes that have suspicious instance counts (even after making `findDelegate` weak):

* `FindSearchField`
* `FindViewController`
* `SupplementaryFindViewController`
* `Document`
* `Label`

So I guess that's a place to start.

scholtzan: The fixes in https://github.com/xi-editor/xi-mac/pull/235 probably got lost during some rebase :/ I added them again. How can I get the number of instances to check whether the leak is fixed?

XiEditor/Search.swift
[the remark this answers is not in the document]
scholtzan: Added a comment about the structure. I wish there was a cleaner solution :(

XiEditor/Search.swift
ryanbloom: @scholtzan If you activate the [Xcode memory debugger](https://useyourloaf.com/blog/xcode-visual-memory-debugger/) the instance counts show up in the sidebar.

cmyr:
Yea that's annoying. 😞

Is this ready again? I haven't investigated the leak, but it looks like you've dug into it. @ryanbloom do you want to sign off?

scholtzan:
Okay, fixed :)

cmyr:
Okay, so I'm 99% happy!

My one lingering concern is with the fact that this change (well, the core change) is going to break all other frontends. I _think_ we can get around this by simply leaving the client->core `find` RPC as-is, and creating a new `multi_find` RPC; and in core, we will convert `find` into `multi_find` in `edit_types.rs`. It shouldn't be a problem going the other way (`find_status`) because this just adds a field, and clients are free to ignore it.

Does that sound reasonable?

scholtzan:
Sounds for sure reasonable. I'll get right on it.

PR #232, opened 2018-07-22
[WIP] Sidebar/File Navigator

### LINKED ISSUE #142: Question about contributing to a file explorer
I'd love to use Xi for Mac on a more regular basis. I love the goal of this project and am a huge fan of fast, platform-native software.

The one thing missing for me that would make Xi useful for for any real work is a (simple) file/project explorer. I would love to contribute to this. Has any thought been given to how this should be implemented? Would this be a plugin or part of the native Mac app? Should it be part of this project or separate?

### LINKED ISSUE #156: Add open folder
image

plz add folder view，This feature is useful for programmers。

lukakerr:
Currently a work in progress with a few things to get ironed out. I thought I'd open this PR just to get some initial feedback and/or suggestions regarding the UI, UX and the overall implementation.

Right now i'm just letting the user choose either a folder or file. If they choose a file, just open it as per usual, if they choose a folder then we render the contents of that folder into the sidebar.

The default system folder icon is being used, and the file icons are the specific icons for that file type using the `NSWorkspace.shared.icon(forFileType: item.fileType)` method. These can be changed to custom icons, along with the disclosure arrows next to the icons.

### Still Todo
- [ ] Sync the structure (expanded/collapsed/selected) of each windows sidebar. Currently when a file is opened in a new tab, the root folder is collapsed
- [ ] Possibly restore sidebar folders when the app it quit and relaunched

### Thoughts on behaviour

Currently when a file is clicked on it opens into a new tab (unless it is already open in the current tab). This mimics apps like Sublime Text and VS Code, although more native apps like Xcode (when tabs are open) open the selected file into the current tab. I'm not sure what a better user experience is, but since files aren't autosaved, opening a file into the current tab may not work the best.

Also, each window/tab has its own instance of the sidebar. This is a side effect of creating a new window per document and not much can be done at the moment about this. As a result, memory usage is increased for each new tab. Although, a singleton is used for the actual sidebar data so the complexity there is reduced slightly.

As discussed in #142 this is just a basic implementation to get a file tree rendering in a sidebar. Changes will probably have to be made later for when/if core gets support for workspaces.

### Screenshot of current UI

screenshot

Closes #142 
Closes #156

### DIFF
--- a/Sources/XiEditor/AppDelegate.swift
+++ b/Sources/XiEditor/AppDelegate.swift
@@ -298,6 +298,60 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         })
     }
 
+    @IBAction func openDocument(_ sender: Any?) {
+        let dialog = NSOpenPanel()
+
+        dialog.allowsMultipleSelection = false
+        dialog.canChooseFiles = true
+        dialog.showsHiddenFiles = true
+        dialog.canCreateDirectories = true
+        dialog.canChooseDirectories = true
+
+        guard
+            dialog.runModal() == .OK,
+            let result = dialog.url
+            else { return }
+
+        var isDirectory: ObjCBool = false
+
+        if FileManager.default.fileExists(atPath: result.path, isDirectory: &isDirectory) {
+            if isDirectory.boolValue {
+                DispatchQueue.global(qos: .userInitiated).async {
+                    let parent = FileSystemItem.createParents(url: result)
+                    let newItem = FileSystemItem(path: result.absoluteString, parent: parent)
+
+                    DispatchQueue.main.async {
+                        sidebarItems.addItem(newItem)
+
+                        // Don't have a window open
+                        if NSApp.keyWindow == nil {
+                            NSDocumentController.shared.newDocument(nil)
+                        } else {
+                            // Has a window open so show sidebar
+                            (NSApp.keyWindow?.windowController as? XiWindowController)?.showSidebar()
+                        }
+                    }
+                }
+            } else {
+                // When opening a single file, we want to open it in its own window
+                if #available(OSX 10.12, *) {
+                    Document.tabbingMode = .disallowed
+                }
+
+                NSDocumentController.shared.openDocument(
+                    withContentsOf: result,
+                    display: true,
+                    completionHandler: { (document, alreadyOpen, error) in
+                        if let error = error {
+                            print("error opening preferences \(error)")
+                        }
+                        // When opening a file in its own window we don't want to display the sidebar
+                        (NSApp.keyWindow?.windowController as? XiWindowController)?.hideSidebar()
+                });
+            }
+        }
+    }
+
     //- MARK: - helpers
 
     func getUserConfigDirectory() -> String {
--- a/Sources/XiEditor/Document.swift
+++ b/Sources/XiEditor/Document.swift
@@ -60,10 +60,9 @@ class Document: NSDocument {
     override func makeWindowControllers() {
         // save this before instantiating because instantiating overwrites with the default position
         let newFrame = frameForNewWindow()
-        var windowController: NSWindowController!
         let storyboard = NSStoryboard(name: NSStoryboard.Name("Main"), bundle: nil)
-        windowController = (storyboard.instantiateController(
-            withIdentifier: NSStoryboard.SceneIdentifier("Document Window Controller")) as! NSWindowController)
+        let windowController = (storyboard.instantiateController(
+            withIdentifier: NSStoryboard.SceneIdentifier("Document Window Controller")) as! XiWindowController)
 
         if #available(OSX 10.12, *) {
             windowController.window?.tabbingIdentifier = NSWindow.TabbingIdentifier("xi-global-tab-group")
@@ -73,10 +72,16 @@ class Document: NSDocument {
             Document.tabbingMode = .automatic
         }
 
+        if sidebarItems.isEmpty {
+            windowController.hideSidebar()
+        }
+
         windowController.window?.setFrame(newFrame, display: true)
         windowController.window?.minSize = Document.minWinSize
 
-        self.editViewController = windowController.contentViewController as? EditViewController
+        guard let splitViewController = windowController.contentViewController as? NSSplitViewController else { return }
+
+        self.editViewController = splitViewController.children.last as? EditViewController
         editViewController?.document = self
         editViewController?.xiView = XiViewConnection(asyncRpc: sendRpcAsync, syncRpc: sendRpc)
         windowController.window?.delegate = editViewController
--- a/Sources/XiEditor/EditViewController.swift
+++ b/Sources/XiEditor/EditViewController.swift
@@ -227,6 +227,11 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
 
     var hoverEvent: NSEvent?
 
+    private var sidebar: SidebarViewController? {
+        let splitViewController = self.parent as? NSSplitViewController
+        return splitViewController?.splitViewItems.first?.viewController as? SidebarViewController
+    }
+
     let statusBar = StatusBar(frame: .zero)
 
     // Popover that manages hover views.
@@ -254,6 +259,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         scrollView.hasHorizontalScroller = true
         scrollView.usesPredominantAxisScrolling = true
         (scrollView.contentView as? XiClipView)?.delegate = self
+        sidebar?.syncStructure()
     }
 
     override func viewDidAppear() {
@@ -903,6 +909,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
             subItem.state = NSControl.StateValue(rawValue: (subItem.title == theme) ? 1 : 0)
         }
         self.unifiedTitlebar = { self.unifiedTitlebar }()
+        sidebar?.themeChanged()
     }
     
     public func languageChanged(_ languageIdentifier: String) {
--- a/Sources/XiEditor/FileSystemItem.swift
+++ b/Sources/XiEditor/FileSystemItem.swift
@@ -1,0 +1,110 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Cocoa
+
+final class FileSystemItem {
+
+    fileprivate var relativePath: String!
+    fileprivate var parent: FileSystemItem?
+
+    public var numberOfChildren: Int {
+        return children().count
+    }
+
+    public var name: String {
+        return relativePath
+    }
+
+    static let rootFileSystemItem = URL(fileURLWithPath: "/")
+
+    var fullPath: String {
+        guard
+            let parent = self.parent,
+            let url = URL(string: parent.fullPath)?.appendingPathComponent(self.relativePath)
+        else { return self.relativePath }
+
+        return url.absoluteString
+    }
+
+    var url: URL {
+        return URL(string: self.fullPath)!
+    }
+
+    var fileURL: URL {
+        return URL(fileURLWithPath: self.fullPath)
+    }
+
+    var fileType: String {
+        return self.fileURL.pathExtension
+    }
+
+    var isDirectory: Bool {
+        var isDir: ObjCBool = false
+        FileManager.default.fileExists(atPath: self.fullPath, isDirectory: &isDir)
+        return isDir.boolValue
+    }
+
+    convenience init() {
+        self.init(path: FileSystemItem.rootFileSystemItem.path)
+    }
+
+    init(path: String, parent: FileSystemItem? = nil) {
+        self.relativePath = URL(fileURLWithPath: path).lastPathComponent
+        self.parent = parent
+    }
+
+    // MARK: - Public methods
+
+    public func child(at index: Int) -> FileSystemItem {
+        return children()[index]
+    }
+
+    // MARK: - Private methods
+
+    private func children() -> [FileSystemItem] {
+        let fileManager = FileManager.default
+        var isDirectory: ObjCBool = false
+        var children = [FileSystemItem]()
+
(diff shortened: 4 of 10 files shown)


cmyr:
Very cool @lukakerr, I'll give this a proper look in the next day or so. 🎊

cmyr (commented):
A few little nits and observations, but this looks good.

### `String` vs `URL`
One general Cocoa thought; it feels like there's a lot of confusion here about whether to represent file paths as `String`s or as `URL`s. The Cocoa APIs are pretty muddled on this point. The `String` APIs are definitely older, and the `URL` APIs are preferred. In any case, there are is quite a bit of logic here around jumping back and forth. My personal inclination might be to try and keep things as `URL`s as much as possible, and do string conversions only when we need to display things or deal with foreign code / RPC?

### Look and feel

There are a few things that I'd like to have before we merge this, most of which I think you're planning/tracking, but it might be useful to enumerate:

1. sync between documents in the same window
2. A menu item to hide the sidebar
3. No sidebar by default when opening a single file


And... that's maybe it?

### bigger questions

#### tabs

Looking at this sent me down a bit of a rabbit hole around how window tabbing works on Cocoa, which I hadn't thought about in a while.

It sure looks like the only way to really get the system tabbing behaviour is to have separate NSWindow instances. I'd be really curious of there was another mechanism. 

It would be nice to use the system tabbing stuff. Keeping our own implementation up to date would be annoying. This would require us to think more about what a 'window' is, for us, particularly with regards to split views: it would incline us towards an Xcode-style UI where each tab has multiple splits, instead of each split having multiple tabs. In this world a tab is more like a tab in vim, which has its own collection of open documents.

#### state, tracking changes, etc

Whatever route we go, having workspaces like this will probably lead us to supporting non-visible open files. This is something we would want to support in UI; some visual cue to the user that some file in the sidebar has unsaved changes, even if it isn't currently displayed.

------------

Anyway, that's just off the top of my head. I think this is a good direction, and I'm excited to see this get merged.

XiEditor/FileSystemItem.swift
```diff
@@ -0,0 +1,119 @@
...
+            return children
+        }
+        return []
+    }
+
+    // MARK: - Static methods
+
+    static func createParents(url: URL) -> FileSystemItem {
+        let parentUrl = url.deletingLastPathComponent()
+        let path = parentUrl.absoluteString
```
nangtrongvuon: Possibly a redundant observation at this point, but Swift API design guidelines dictates that things like URL (which is an initialism) should be uniformly be upper or lowercased. So right here, `url` can be lowercase but `parentUrl` should be changed to `parentURL`.

XiEditor/SidebarViewController.swift
```diff
@@ -0,0 +1,142 @@
...
+
+    // Height of each row
+    func outlineView(_ outlineView: NSOutlineView, heightOfRowByItem item: Any) -> CGFloat {
+        return 20.0
+    }
+
+    // When a row is clicked on should it be selected
+    func outlineView(_ outlineView: NSOutlineView, shouldSelectItem item: Any) -> Bool {
+        guard let item = item as? FileSystemItem else { return false }
+        return !item.isDirectory
```
nangtrongvuon: Even if it's a directory, it should still be clickable, and I also suggest that double clicking expands the directory, similar to the arrow click. I found clicking on arrows to expand directories to be jarring and inaccurate (especially for folders that have a lot of subfolders, much like our own xi-core repo), since the arrow isn't very big by design and it takes too much effort and accuracy to just expand directories in the sidebar. 

Double-clicking to expand folders has become quintessential UX (for good reason!) at this point, and users familiar with desktop environments expect it. You can see many other apps that make use of a directory sidebar (like other editors, or Finder, or XCode) allow users to double-click on directory rows to expand them.

XiEditor/SidebarViewController.swift
```diff
@@ -0,0 +1,142 @@
...
+            selector: #selector(self.updateItems),
+            name: NSNotification.Name(rawValue: "sidebarItemsChanged"),
+            object: nil
+        )
+
+        self.updateItems()
+    }
+
+    override func viewWillAppear() {
+        sidebar.backgroundColor = self.theme.background
```
nangtrongvuon: Little nit here - you don't change the text color (aka `theme.foreground`) here, which causes the sidebar to keep a dark text color even when the theme changes to a dark theme.

nangtrongvuon (commented):
Very cool, I might actually ditch VSCode if this gets merged, although I have a few observations, besides what @cmyr has already touched on:

- Opening a new folder should show that new folder in the foreground. 
- There should be a "view sidebar" menu item, although I think that's already on your mind.
- You should make more use of URL methods, as already commented on by @cmyr, since it's the preferred way going forward from Swift 4. 

Other than that, I'm really excited to see this make it into Xi. Glad to have you with us!

XiEditor/SidebarViewController.swift
lukakerr: @nangtrongvuon for this one I've just tested with the dark themes and the text color seems to automatically change to white/black depending on the background. Below is an example

screen shot 2018-07-26 at 11 06 01 am

XiEditor/FileSystemItem.swift
```diff
@@ -0,0 +1,119 @@
...
+        return self.fileUrl.pathExtension
+    }
+
+    var isDirectory: Bool {
+        var isDir: ObjCBool = false
+        FileManager.default.fileExists(atPath: self.fullPath, isDirectory: &isDir)
+        return isDir.boolValue
+    }
+
+    convenience init() {
```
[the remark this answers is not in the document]
lukakerr: Made a change to this based off your comments, not sure if its exactly what you were referring to so let me know if theres any issues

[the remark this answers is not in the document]
cmyr: Changes looks good. 👍

XiEditor/SidebarViewController.swift
ryanbloom: I tried reproducing this bug, and it only happens when there's a dark theme *and* `unified_titlebar = false`. In this case, the text stays dark because [the window has an `.aqua` appearance](https://github.com/lukakerr/xi-mac/blob/sidebar/XiEditor/EditViewController.swift#L155). If you make the sidebar's appearance dark, the text will automatically change to light.

lukakerr: Ah yep I can reproduce. Should be fixed now

jansol:
What's the status on this?

lukakerr:
The only thing left to do is sync the sidebar structure (which items are expanded/selected etc) across tabs. I haven’t had much time to work on it due to uni work though, but I haven’t forgot about it

XiEditor/SidebarViewController.swift
```diff
@@ -0,0 +1,165 @@
...
+        let rows = IndexSet(integersIn: 0..<outlineView.numberOfRows)
+
+        rows.compactMap { outlineView.rowView(atRow: $0, makeIfNecessary: false) }
+            .forEach { $0.backgroundColor = $0.isSelected ? .selectedControlColor : .clear }
+
+        if doc.isDirectory { return }
+
+        // Get the document controller and open the selected document into a new tab
+        if let appDelegate = NSApplication.shared.delegate as? AppDelegate {
+            appDelegate.documentController.openDocumentIntoNewTab(
```
nangtrongvuon: Just UX stuff, but I think the behavior should be if it's a single click, open in current window and if it's a double click, open in new tab. `doubleAction` should work here.

XiEditor/AppDelegate.swift
```diff
@@ -409,6 +409,57 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
...
+        dialog.canChooseFiles = true
+        dialog.canCreateDirectories = true
+        dialog.canChooseDirectories = true
+
+        guard
+            dialog.runModal() == .OK,
+            let result = dialog.url
+        else { return }
+
+        var isDirectory : ObjCBool = false
```
nangtrongvuon: Just general formatting nit, typically type colons are placed next to the variable, so this should look like `var isDirectory: ObjcBool = false`.

XiEditor/FileSystemItem.swift
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    init(path: String, parent: FileSystemItem? = nil) {
+        self.relativePath = URL(fileURLWithPath: path).lastPathComponent
+        self.parent = parent
+    }
+
+    // MARK: - Public methods
+
+    public func getChild(at index: Int) -> FileSystemItem {
```
nangtrongvuon: For these `get` methods, I would omit the `get` from these methods in general, since it should read as a noun phrase. You can read about this at the [Swift design guidelines's naming section.](https://swift.org/documentation/api-design-guidelines/)

For example, a small snippet of your code is `return item.getChild(at: Index)`. Since the get doesn't actually perform any side effects, this can just be `return item.child(at: Index)`, and it would look a bit more verbose in my opinion.

XiEditor/FileSystemItem.swift
```diff
@@ -0,0 +1,111 @@
...
+        return children[index]
+    }
+
+    public func getNumberOfChildren() -> Int {
+        return children.count
+    }
+
+    public func getName() -> String {
```
nangtrongvuon: Alternatively, for this `getName()` and `getNumberOfChildren` methods, I'd just use a computed property here.

For example, `getName()` would be something like `var name: String { return self.relativePath }`. This is more "Swift", in my opinion.

XiEditor/SidebarItems.swift
```diff
@@ -0,0 +1,53 @@
...
+
+import Foundation
+
+/// Singleton used across window sidebars containing an array of file system items
+/// Provides simple methods to modify the items
+/// Can listen to changes in the items by hooking into the "sidebarItemsChanged" notification
+final class SidebarItems {
+
+    static let sharedInstance = SidebarItems()
+    public var items: [FileSystemItem]
```
nangtrongvuon: Perhaps this can be `public var items = [FileSystemItem]()`, and you can remove the init that's below it for more clarity.

XiEditor/Document.swift
```diff
@@ -78,10 +78,9 @@ class Document: NSDocument {
     override func makeWindowControllers() {
         // save this before instantiating because instantiating overwrites with the default position
         let newFrame = frameForNewWindow()
-        var windowController: NSWindowController!
         let storyboard = NSStoryboard(name: NSStoryboard.Name(rawValue: "Main"), bundle: nil)
-        windowController = storyboard.instantiateController(
-            withIdentifier: NSStoryboard.SceneIdentifier(rawValue: "Document Window Controller")) as! NSWindowController
+        let windowController = storyboard.instantiateController(
+            withIdentifier: NSStoryboard.SceneIdentifier(rawValue: "Document Window Controller")) as! XIWindowController
```
nangtrongvuon: Should be `XiWindowController`.

XiEditor/Document.swift
```diff
@@ -94,7 +93,9 @@ class Document: NSDocument {
         windowController.window?.setFrame(newFrame, display: true)
         windowController.window?.minSize = Document.minWinSize
 
-        self.editViewController = windowController.contentViewController as? EditViewController
+        guard let splitViewController = windowController.contentViewController as? NSSplitViewController else { return }
```
nangtrongvuon: It seems like there's a bug here where at least 2 windows have to be opened before the initial window is shown. On current master, when opening xi-mac I get an "Untitled" window, but with this branch there is sometimes a dummy "Untitled 2" window before I can interact with a "Untitled 3" window.

nangtrongvuon (dismissed):
There are some things that need a bit more refining, but I'm really excited with the progress!

jansol:
Regarding general UX, one bit that I like in atom is that when I single-click on a file in the sidebar it opens a transient "preview" tab for it, or replaces the current "preview" tab if there is already one. Then, if I make an edit in that file, it turns that into a regular (persistent) tab. If I double-click instead, it opens a persistent tab directly. This is a nice way to avoid an unwieldy number of tabs cropping up quickly, as often I just want to skim through a few files to either find some section of the code or to look up details on something. Of course liberal use of ripgrep and RLS definitions could work to the same effect, but definitions require me to already know an exact name to type and ripgrep requires a context switch to a terminal (and is not necessarily the most intuitive solution for people accustomed to GUIs).

Just something that came to mind, perfectly fine with this getting ignored or postponed.

XiEditor/SidebarViewController.swift
lukakerr: @nangtrongvuon I thought this would be a good idea too, similar to how Xcode handles it. Although, how would we account for unsaved file changes to a document? The way Xcode does it is that it doesn't save the file and stores the changes in the background, so when the file is reopened, the changes are still there. And if a file has unsaved changes when Xcode quits, the changes get saved automatically. Does xi-editor support this kind of behavior?

XiEditor/Document.swift
lukakerr: I believe this should be fixed now

estk:
@lukakerr @cmyr, this is incredible! I'm with @nangtrongvuon, when this gets merged I'll be piloting xi as a vscode/atom replacement.

I built the project and have a few usability thoughts.

## Single click/ Double click
Atom has an interesting strategy for this, the current tab is only replaced when the file previously open was also opened with a single click, otherwise the new file is opened in a single tab. Probably there are two modes for a tab: `view` and `edit`. Double click on the file in the sidebar or on the filename in the tab indicates the file is in `edit` mode. Edit mode tabs are never replaced by newly opened files, View mode tabs on the other hand are always replaced. Hope this helps.

## Sidebar across tabs
Sidebar state should be synced, and if possible the data structure that indicates the status should be shared. Also, when a new folder is opened in the sidebar, it's default should be expanded.

## Sidebar hide/show
The animation to show the sidebar cuts off part of the open file before it slides right.

Hopefully that's helpful, thanks for all your hard work!

daleione:
any update?

lukakerr:
@daleione I've been pretty busy at uni to work on this. That said, I've actually implemented the sidebar structure syncing between tabs in my own editor project [Pine](https://github.com/lukakerr/Pine). So, i'll try to at least get the conflicts resolved and the sidebar structure syncing working in the next few days.

Apart from that, there are still a few minor things to fix up. Notably, the 'replace current document in tab' behaviour (similar to Xcode's behaviour) doesn't exist. This is because xi core doesn't (from my understanding) support file changes in the background without saving the file. So for the meantime, a new tab is created when a document is opened from the sidebar. This behaviour might also need to be worked on to iron the specifics out.

reykjalin:
I'm interested in adding a request to this discussion, although I'm not sure if this belongs in a new feature request: having the file navigator on the _right side_ instead of the left side.

The reason I personally like this is because when opening/closing the sidebar, instead of pushing the code to the side or pulling it back, it just cuts off or reveals the "back part" of the code.

I'm not sure how hard it would be to implement this, but it would be great if you could choose which side to have your sidebar on :)
Who knows, maybe someone likes it on the top of their window (although that might be getting a little out of hand 😬 ).

boozook (approved):

Issue #231, opened 2018-07-21
Memory leak when windows are closed


ryanbloom:
Xi's memory usage increases whenever a file is opened, but doesn't decrease substantially when files are closed. In the Xcode memory debugger, it looks like a reference cycle is keeping old `EditViewController`s in memory.

Here's the closest thing I've found to a fix (see [this branch](https://github.com/ryanbloom/xi-mac/commit/cd18ff80c83dacf58a5ad76ebc569b478488f46f)):

1. Make `EditView.dataSource` weak
2. Make `XiClipView.delegate` weak
3. Make `Document.editViewController` weak
4. Comment out the line where `EditViewController` tells `FindViewController` to update its background color

Unfortunately, this undoes #223 and also introduces a new crash:

`XiEditor(37590,0x1010f1380) malloc: *** error for object 0x6080000e5880: Invalid pointer dequeued from free list
*** set a breakpoint in malloc_error_break to debug
`

The crash has happened opening windows, closing windows, and editing text. The only constant factor seems to be that `malloc` is being called. If I change `XiClipView.delegate` back to strong, the problem goes away.

So I guess it'll take some more work, but I'm not sure what to do about the crash, or why change 4 was necessary.

jansol:
#235 seems to fix all of the points raised here.

PR #235, opened 2018-07-27
Fix memory leak when closing windows

ryanbloom:
This is a potential solution to #231. It fixes the `EditViewController` memory leak, and also the crash that surfaced when trying to fix the leak.

## Memory leak
Solving the memory leak only required weakening a few references to `EditViewController`. I still don't understand why commenting out `findViewController.updateColor(...)` helped, but that turns out not to be necessary if `FindViewController.findDelegate` is weak.

## Crash
What appears to have happened is that `CGLPixelFormat` objects were getting released one time too many, which caused a problem as soon as their memory was reallocated for something else. (The extra release didn't happen when all the ViewControllers were kept in memory.) The best fix I could find was to override `TextPlaneLayer.releaseCGLPixelFormat(...)` with an empty implementation, which prevents one of the releases. I'm not quite sure if that's an appropriate solution or a hacky workaround, but it seems to work pretty well. 😄

### DIFF
--- a/XiEditor/Document.swift
+++ b/XiEditor/Document.swift
@@ -60,7 +60,7 @@ class Document: NSDocument {
     var tabbingIdentifier: String
     
 	var pendingNotifications: [PendingNotification] = [];
-    var editViewController: EditViewController?
+    weak var editViewController: EditViewController?
 
     /// Returns `true` if this document contains no data.
     var isEmpty: Bool {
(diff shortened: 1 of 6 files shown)


cmyr:
@ryanbloom thanks for this! It might take a little while to get a proper review in, will need to get my head around the problem. Very happy to have you digging into this though, it's important to keep up on maintenance.

PR #236, opened 2018-07-27
Fix find view background not applying

lukakerr:
The call to set/update the find view background by @ryanbloom in c983d98 seems to have been removed. This PR just adds it back in, as currently the find view's backing layer doesn't get set:

screen shot 2018-07-27 at 8 44 17 pm


cmyr:
ah I didn't even notice this PR, thanks. will close #237.

PR #239, opened 2018-07-31
Add support to split selections into lines

scholtzan:
This depends on https://github.com/google/xi-editor/pull/770


XiEditor/Search.swift
```diff
@@ -138,7 +138,7 @@ class FindViewController: NSViewController, NSSearchFieldDelegate, NSControlText
         if NSEvent.modifierFlags.contains(.shift) {
             findDelegate.findPrevious(wrapAround: wrapAround)
         } else {
-            findDelegate.findNext(wrapAround: wrapAround, allowSame: false)
```
cmyr: is this change here intentionally? It does not seem like it should be part of this PR.

scholtzan: Oh, I think I just tried something out and forgot to change it back. I will undo it.

Issue #189, opened 2018-06-01
xi-mac is "Using Significant Energy" even when no windows are open


jansol:
xi-mac shows up in the "active battery drainers" list in the macos battery level dropdown almost all the time when it is running, regardless of open files. Activity monitor shows that xi-core is doing *something* even though it doesn't seem to have reason to.

Sample taken from Activity Monitor:
```
Sampling process 26331 for 1 second with 1 millisecond of run time between samples
Sampling completed, processing symbols...
Analysis of sampling xi-core (pid 26331) every 1 millisecond
Process:         xi-core [26331]
Path:            /Users/xxx/code/xi-mac/build/Release/XiEditor.app/Contents/Resources/xi-core
Load Address:    0x1043dd000
Identifier:      xi-core
Version:         0
Code Type:       X86-64
Parent Process:  XiEditor [26330]

Date/Time:       2018-06-01 23:19:24.101 +0900
Launch Time:     2018-06-01 23:11:29.589 +0900
OS Version:      Mac OS X 10.13.4 (17E202)
Report Version:  7
Analysis Tool:   /usr/bin/sample

Physical footprint:         1776K
Physical footprint (peak):  1784K
----

Call graph:
    835 Thread_14287431   DispatchQueue_1: com.apple.main-thread  (serial)
    + 835 start  (in libdyld.dylib) + 1  [0x7fff7a5fb015]
    +   835 main  (in xi-core) + 44  [0x10442aacc]
    +     835 std::rt::lang_start_internal::h336cd1da103a02e8  (in xi-core) + 242  [0x1045fe0d2]
    +       835 __rust_maybe_catch_panic  (in xi-core) + 31  [0x1046159af]
    +         835 std::panicking::try::do_call::hc7b9e6190a1d9f3e (.llvm.7628423903506348692)  (in xi-core) + 24  [0x1045ff7e8]
    +           835 std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::h2d63558f062ba807  (in xi-core) + 6  [0x1044286b6]
    +             835 xi_core::main::hd19e9726f90a7bea  (in xi-core) + 70  [0x10442a8d6]
    +               835 _$LT$xi_rpc..RpcLoop$LT$W$GT$$GT$::mainloop::h2c000e806680b824  (in xi-core) + 50  [0x10444a952]
    +                 835 crossbeam::scoped::scope::hb0d05522ea3d59e7  (in xi-core) + 1221  [0x104422945]
[...]

PR #240, opened 2018-08-03
Fix xi-editor GPU switching

### LINKED ISSUE #238: Xi-mac forces mac to switch to high performance GPU
Launching xi-mac unconditionally switches my mbp to use the dedicated GPU, causing it to heat up and likely being at least partially the reason for #189 .

Alacritty has the same issue (since it also uses OpenGL for its rendering) and an easy looking solution is proposed [here](https://github.com/jwilm/alacritty/issues/210#issuecomment-271249897).

lukakerr:
This fixes #238 and #189 by adding attribute to not switch to dedicated GPU for macs that have an integrated GPU

### DIFF
--- a/XiEditor/Info.plist
+++ b/XiEditor/Info.plist
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSSupportsAutomaticGraphicsSwitching</key>
+	<true/>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDocumentTypes</key>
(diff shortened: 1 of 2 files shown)


jansol:
LGTM

jansol:
It did at least put my laptop back on its iGPU. Idle CPU & power usage still remains nontrivial but as established on IRC that's a separate issue.

cmyr:
Okay, thanks for letting me know. I'll leave it up and see if @raphlinus wants to take a look, otherwise will merge in a few days.

raphlinus:
For me the only question is whether any of the other points in [Supporting Multiple GPUs on Mac OS X](https://developer.apple.com/library/archive/technotes/tn2229/_index.html) are needed. If we're covered, I'm happy with this going in. Thanks for investigating!

I'd also be curious about actual effects on both performance and power usage. On my Gigabyte Aero 14 with 1070 and HD 630, I find that the 1070 actually uses _less_ power for things like scrolling, but it's complicated.

jansol:
According to that, display reconfiguration needs to be dealt with (`CGDisplayRegisterReconfigurationCallback`). NSOpenGLView apparently handles that automatically but I didn't find any evidence that NSOpenGLLayer, which xi-mac uses, would do the same. Then again NSOpenGLLayer and its underlying CAOpenGLLayer are deprecated since macOS 10.15, along with OpenGL in general.

Triggering a switch to the dedicated GPU while xi-mac is running and back didn't cause any problems. Unless there's a possiblity that different monitors are driven by different GPUs, which I can't test. That should not be too different from triggering a GPU switch though so unless someone can test and report otherwise I'm fairly confident this can go in as-is.

cmyr:
Okay, I'm happy to defer to @jansol and @lukakerr, thanks for looking into this!

PR #241, opened 2018-08-05
Give all windows the same tabbing identifier

ryanbloom:
This fixes a minor bug I found with tabbing behavior: if you create two separate windows (⌘N ⌘N), Xi won't let you drag tabs between them. Apparently this is because the windows are given different tabbing identifiers—if I remove that code, the problem goes away. As far as I can tell, the new tab and new window behavior remains exactly the same.

### DIFF
--- a/XiEditor/Document.swift
+++ b/XiEditor/Document.swift
@@ -22,18 +22,8 @@ struct PendingNotification {
 
 class Document: NSDocument {
 
-    /// used internally to keep track of groups of tabs
-    static fileprivate var _nextTabbingIdentifier = 0
-
-    /// returns the next available tab group identifer. When we create a new window,
-    /// if it is not part of an existing tab group it is assigned a new one.
-    static private func nextTabbingIdentifier() -> String {
-        _nextTabbingIdentifier += 1
-        return "tab-group-\(_nextTabbingIdentifier)"
-    }
-
-    /// if set, should be used as the tabbingIdentifier of new documents' windows.
-    static var preferredTabbingIdentifier: String?
+    @available(OSX 10.12, *)
+    static var tabbingMode = NSWindow.TabbingMode.automatic
 
     // minimum size for a new or resized window
     static var minWinSize = NSSize(width: 240, height: 160)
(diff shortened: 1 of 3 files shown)


cmyr:
Awesome, I would be very happy to delete this hack. IIRC the original problem was with correctly associating new tabs with the window that had received the event, but it definitely didn't feel quite right.

cmyr:
Oh, I remembered something else about this: behaviour is highly dependent on the user's setting for 'Prefer tabs when opening documents'. Did you test this with that on always/manual?

cmyr:
Played around with this a little bit.

The first thing I noticed: after opening two new windows (after closing all windows), and with the second window focused, <kbd>cmd</kbd><kbd>T</kbd> creates a new tab in the _first_ window, not in the active window.

cmyr:
@ryanbloom yea, we're on the same page. The mess you're looking at is the best I could come up with when we first added tab support, which bought us n. 2 but cost us n. 4. My design at the time was to have cmd+N open a new tab if that was the user's system preference, but I can go either way on that?

jansol:
Haven't checked the behaviour yet but agree with having the option to always open a new window. Sometimes that just is what I want, even if I normally prefer otherwise. One thing worth considering is the exact shortcut for that: TextEdit uses ⌥⌘N whereas Atom uses ⇧⌘N. Xcode on the other hand is a bit confusing since it doesn't let you create scratch buffers to begin with but always requires you to go through the file creation wizard. It uses ⌥⌘N to create a new group and ⇧⌘N for a new project so if we want to follow the Atom/Sublime model of one window per project/workspace then ⇧⌘N is definitely the way to go. On the other hand there's TextEdit which doesn't have the concept of 'projects' (nor a folder sidebar).

jansol:
OK, playing around with this I made some observations:
1. ⌥⌘N seems to be accepted for 'new window' despite not being present in the menu (this is the case in TextEdit as well - but *not* in Atom, where it is also accepted as a new document shortcut but opens a tab instead of a window/workspace - so I suppose it has something to do with Cocoa itself)
2. Dragging tabs into other windows as tabs still only works if tone is 1) dragging a tab, not a window 2) the receiving window already has a tab bar.

Regarding latter point, 2) can probably be fixed by forcing a tab bar on all windows when one of our windows is being dragged. 1) is more difficult since we'd have to turn the dragged window into a tab automagically. There's also the case of dragging a window that already has tabs - how should it behave when dragged as a tab into another window? Should it be shorthand for moving all tabs to that window? Should it be allowed at all (i.e. *not* trigger the tab bar on other windows)?

As for my previous comment, the exact behaviour there is mostly a matter of what kind of editor xi aims to be. Given the heavy focus on code I guess the folder/workspace -centric approach makes more sense.

As for the changes in here, ⌘O now honours the system setting and ⌘N and ⌘T do what they say on the tin so LGTM, unless you want to fix the tab/window dragging issue here as well.

cmyr:
(also just noticed I've duplicated some of @jansol's comments; great minds etc etc)

PR #245, opened 2018-08-08
Fix Swift 4.0 build error

cmyr:


cmyr:
@nangtrongvuon no worries! The issue is that the most recent Xcode that sierra supports is 9.2. You can probably run later versions of swift, but you can't run later versions of Xcode. Apple is pretty ruthless about dropping support, so not much to do there.

Probably the better solution is to have CI set up for this repo, and have it build against a few different configs.

PR #256, opened 2018-09-09
Fix dismiss find causing overscroll

nangtrongvuon:
Currently on master, if the find view is opened and quickly dismissed without typing anything, the view is scrolled beyond the window. This commit fixes that issue.


cmyr:
Can you explain what the bug was? I couldn't figure it out when I was playing around...

PR #257, opened 2018-09-09
Fix flashing title bar color

ryanbloom:
When a new tab is created, it has a light gray title bar for a few dozen milliseconds, even if the theme is dark and `unified_titlebar = true`. This is because the config dictionary is sent to new tabs asynchronously after they're created; once this happens the title bar takes on the correct theme color.

Having not read through the entire config system, I think it makes sense for `AppDelegate` to cache the dictionary and send it to new windows/tabs immediately before they appear.

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -369,7 +369,14 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         }
     }
 
+    // Stores the config dict so new windows don't have to wait for core to send it.
+    // The main purpose of this is ensuring that `unified_titlebar` applies immediately.
+    var configCache: [String: AnyObject] = [:]
+
     func configChanged(viewIdentifier: ViewIdentifier, changes: [String : AnyObject]) {
+        for (key, value) in changes {
+            self.configCache[key] = value
+        }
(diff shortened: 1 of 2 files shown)


cmyr:
sorry for the delay, will get a look at this over the weekend. Also appreciate you're waiting for a decision on the xi-editor PR about movement, will try and make a decision there.

also not sure if you're aware but a few of us are fairly active in the #xi channel on irc.mozilla.org, I feel like the movement thing in particular might deserve to get talked through.

XiEditor/AppDelegate.swift
```diff
@@ -369,7 +369,12 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         }
     }
 
+    var config: [String: AnyObject] = [:]
```
cmyr: can we call this something like `configCache` or `lastConfig`, and include a brief comment describing the problem that we're working around?

The eventual solution will be to have a separate scope for client config settings that will be unrelated to buffer-specific settings, but this is reasonable for now.

An alternative option would be to _just_ cache these client settings, and handle them separately. That's clearer in its intent, but is a bit more code. I think either approach is fine, but it's good to document our hacks so they're easier to clean up later. :)

cmyr:
Thanks for the fixup!

PR #259, opened 2018-09-15
Fix transparent find bar

### LINKED ISSUE #258: [UI Bug] Find menu is transparent when opening more than one file in a session
The find menu (⌘+F) has the correct background color for the first file I open in a session, but every subsequent file I open has a transparent find menu.

I was able to replicate this issue every time I restarted Xi. I'm running the current version.

This might be relevant: https://github.com/google/xi-mac/issues/194#issuecomment-408353412

Thanks!

ryanbloom:
Fixes #258

### DIFF
--- a/XiEditor/Search.swift
+++ b/XiEditor/Search.swift
@@ -15,6 +15,11 @@
 import Cocoa
 import Swift
 
+extension NSColor {
+    // Used for the background of the find panel
+    static var veryLightGray = NSColor(white: 246.0/256.0, alpha: 1.0)
+}
+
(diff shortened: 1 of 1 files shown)


XiEditor/Search.swift
```diff
@@ -15,6 +15,11 @@
 import Cocoa
 import Swift
 
+extension NSColor {
```
cmyr: as an item of personal style I might've just had a `static var fallbackBackgroundGray` on the `FindViewController` class instead of doing this as an extension, but I don't have an argument beyond personal taste, so I think this is fine. :)

PR #284, opened 2018-10-02
Delay hiding of statusbar to avoid flicker

### LINKED ISSUE #234: statusbar should set a timer before hiding when item is removed
Currently with some of the LSP stuff we get a very blinky statusbar. I think ideally when the last item is removed from the statusbar, it sets a timer for a second or so in the future and then when that timer fires, if the statusbar is still empty, it gets hidden.

Statusbar showing/hiding is something that we also might consider animating at some point.

jansol:
Fixes #234

### DIFF
--- a/XiEditor/StatusBar.swift
+++ b/XiEditor/StatusBar.swift
@@ -101,7 +101,7 @@ class StatusBar: NSView {
     override init(frame frameRect: NSRect) {
         super.init(frame: .zero)
         self.translatesAutoresizingMaskIntoConstraints = false
-        checkStatusBarVisibility()
+        updateStatusBarVisibility()
     }
 
     required init?(coder decoder: NSCoder) {
@@ -120,7 +120,7 @@ class StatusBar: NSView {
         currentItems[item.key] = item
         self.needsUpdateConstraints = true
         checkItemsFitFor(windowWidth: self.bounds.width)
-        checkStatusBarVisibility()
+        updateStatusBarVisibility()
     }
 
     // Update a status bar item with a new value.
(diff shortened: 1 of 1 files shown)


XiEditor/StatusBar.swift
```diff
@@ -210,9 +212,20 @@ class StatusBar: NSView {
         self.needsDisplay = true
     }
 
-    // Hides the status bar if there is no item currently.
-    func checkStatusBarVisibility() {
-        self.isHidden = currentItems.isEmpty
+    // Shows/hides the statusbar based on whether it has any contents
```
nangtrongvuon: You should add a period at the end here for punctuation :p

jansol:
(Also renamed the function in question because "check" sounded rather awkward)

XiEditor/StatusBar.swift
```diff
@@ -210,9 +212,20 @@ class StatusBar: NSView {
...
+    func updateStatusBarVisibility() {
+        if !currentItems.isEmpty {
+            self.hideTimer?.invalidate()
+            self.isHidden = false
+        } else if !self.isHidden {
+            // Wait a moment before hiding the bar to avoid blinking in the case
+            // that a single item is added and removed repeatedly in rapid succession.
+            if #available(OSX 10.12, *) {
+                hideTimer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: false) {_ in self.isHidden = true}
```
nangtrongvuon: The block closure at the end should have some spaces to be easier to read, so should look like  `{ _ in self.isHidden = true }`

cmyr:
👍

nangtrongvuon:
Hey, as a last note, can you change `2dcefdb`'s commit message to say:  "Fixed up grammar and punctuation"?

cmyr:
@nangtrongvuon is nittier than me!

Issue #276, opened 2018-09-27
Support for macOS 10.14 dark mode


singularitti:
Will you consider support the macOS 10.14 Mojave's dark mode?
Expected features:
1. Dark title bar: just as what they do for Terminal.app:
<img width="639" alt="titlebar" src="https://user-images.githubusercontent.com/25192197/46174523-49366b00-c277-11e8-9522-e1e7197046b4.png">
2. Allow user to choose from dark and light titlebars
3. Dynamic title bar: The titlebar's color will change from light to dark when the sun is set (This feature may be too much so I do not really expect it to be implemented now)

cmyr:
yep, definitely would be nice to do, good to have an issue for this.

singularitti:
Great! Thanks! I think we should also allow to change the color of left sidebar(line numbers)? I forgot to mention it in the top.

singularitti:
For the third request, maybe we can let xi to detect the system theme and change according to it? I do not know how it is implemented but in some apps they could follow the system's theme.

jansol:
One thing I'd like to see that's kind of related to this is adopting translucent VisualEffectView + vibrancy (as demonstrated at [WWDC](https://developer.apple.com/videos/play/wwdc2018/218)) on *all the things*. (I love using vim in a translucent terminal.)

As for automatically changing between light and dark... xi-mac could allow that iff the current colorscheme has both a light and a dark version.

habovh:
AFAIK and according to [the docs](https://developer.apple.com/documentation/appkit/supporting_dark_mode_in_your_interface) if the app uses semantic colors from `NSColor`, the dark mode support and the transition would be automatically handled by the system.

jansol:
There was a mention about having to use a property instead of setting the color manually in the WWDC video. Setting it manually would mean having to update it manually, whereas the property is supposed to propagete updates from the system automagically.

habovh:
If you comment out [these lines](https://github.com/xi-editor/xi-mac/blob/1fa66c82f52c49137a61a286987451ee28ea66c3/XiEditor/EditViewController.swift#L152-L156), the title bar follows the system theme on first opening window as well. I'm not sure what these lines are here for though, so it definitely needs a more insightful research.

[...]

PR #287, opened 2018-10-03
[WIP] Add macOS 10.14 Dark Mode support

habovh:
# WIP

capture d ecran 2018-10-03 a 00 09 52

As suggested in #276 by @jansol and @nangtrongvuon, I'm creating a PR so we can push the discussion further. Bear in mind this is my first contribution to a Swift project. 🙈

This is a work in progress, and should probably not be merged as-is.

## Set theme flow

The Theme changes originally follow this flow during the app lifecycle:
- Select theme
  - App initialization
    - Get theme list from `core`
    - Get current theme from preferences
  - Manual theme setting
     - Get theme name from menu item
- Apply theme
  - Send a *set theme* message to `core`
  - `core` responds to `mac` with a JSON-like theme object
  - The provided theme object is parsed by `mac` and sets the menu item as *active*

### Working stuff

What works in the current PR, is selecting the `macOS` theme after app initialization. Doing so will inhibit the *set theme* message usually sent to `core`, and update the UI with a `Theme` object directly.

### Not working stuff

I found that setting the theme this way does not work for some keys (they are explicitly listed in comments in the code).
Opening the app when `macOS` theme is selected results in an error from `core`, because of the *set theme* message sent on startup with a theme name that does not exist in `core`. We could inhibit this as well.
`core` seems to still have a word to say in what's displayed on the screen somehow, and the previously selected theme will apply to the editor text color.

## Thoughts

As you can see, this PR is pretty hacky, so we may want to come up with a better solution. But I feel this is definitely doable.

I feel the best solution (for a working v1) would be to have a dedicated theme menu entry that is treated like a special case that sends a *set theme* message to core with an existing theme (e.g. `base16-ocean.light`) based on the current selected system color scheme, so the editor can use the text color from core, and then set the rest of the app UI with a Theme object using semantic `NSColors`.

If any of you guys have insights or general tips regarding Swift development please feel free to share your feedback and contribute!

😃

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -148,11 +148,15 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
 
                 statusBar.updateStatusBarColor(newBackgroundColor: self.theme.background, newTextColor: self.theme.foreground, newUnifiedTitlebar: unifiedTitlebar)
                 findViewController.updateColor(newBackgroundColor: self.theme.background, unifiedTitlebar: unifiedTitlebar)
-
-                if color.isDark && unifiedTitlebar {
-                    window.appearance = NSAppearance(named: NSAppearance.Name.vibrantDark)
+                
+                if #available(OSX 10.14, *) {
+                    // Let system decide, disregarding active theme color
                 } else {
-                    window.appearance = NSAppearance(named: NSAppearance.Name.aqua)
+                    if color.isDark && unifiedTitlebar {
+                        window.appearance = NSAppearance(named: NSAppearance.Name.vibrantDark)
+                    } else {
+                        window.appearance = NSAppearance(named: NSAppearance.Name.aqua)
+                    }
                 }
             }
         }
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -50,6 +50,8 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
 
     @IBOutlet weak var editViewHeight: NSLayoutConstraint!
     @IBOutlet weak var editViewWidth: NSLayoutConstraint!
+    
+    let appDelegate = NSApp.delegate as! AppDelegate
 
     lazy var findViewController: FindViewController! = {
         let storyboard = NSStoryboard(name: NSStoryboard.Name(rawValue: "Main"), bundle: nil)
@@ -618,6 +620,11 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     // MARK: - Debug Methods
     @IBAction func debugSetTheme(_ sender: NSMenuItem) {
         guard sender.state != NSControl.StateValue.on else { print("theme already active"); return }
+        if sender.title == "macOS" {
+            let theme = Theme.systemTheme()
+            appDelegate.themeChanged(name: sender.title, theme: theme)
+            return
+        }
         let req = Events.SetTheme(themeName: sender.title)
         document.dispatcher.coreConnection.sendRpcAsync(req.method, params: req.params!)
     }
@@ -759,6 +766,12 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
             .first?.title
 
         pluginsMenu.removeAllItems()
+        
+        let systemThemeMenuItem = NSMenuItem(title: "macOS", action: #selector(EditViewController.debugSetTheme(_:)), keyEquivalent: "")
+        systemThemeMenuItem.state = NSControl.StateValue(rawValue: ("macOS" == currentlyActive) ? 1 : 0)
+        
+        pluginsMenu.addItem(systemThemeMenuItem)
+        pluginsMenu.addItem(NSMenuItem.separator())
         for theme in themes {
             let item = NSMenuItem(title: theme, action: #selector(EditViewController.debugSetTheme(_:)), keyEquivalent: "")
             item.state = NSControl.StateValue(rawValue: (theme == currentlyActive) ? 1 : 0)
--- a/XiEditor/Theme.swift
+++ b/XiEditor/Theme.swift
@@ -75,6 +75,29 @@ extension Theme {
               shadow: nil
             )
     }
+    
+    static func systemTheme() -> Theme {
+        if #available(OSX 10.13, *) {
+            return Theme(foreground: NSColor.textColor,
+                         background: NSColor.textBackgroundColor,
+                         caret: NSColor.red, // Seems to be ignored
+                         lineHighlight: NSColor.green, // Seems to be ignored
+                         findHighlight: NSColor.blue, // Seems to be ignored
+                         findHighlightForeground: NSColor.purple, // Seems to be ignored
+                         gutter: NSColor.textBackgroundColor,
+                         gutterForeground: NSColor.secondaryLabelColor,
+                         selection: NSColor.selectedTextBackgroundColor,
+                         selectionForeground: NSColor.systemPink, // Seems to be ignored
+                         selectionBorder: NSColor.yellow, // Seems to be ignored
+                         inactiveSelection: NSColor.brown, // Seems to be ignored
+                         inactiveSelectionForeground: NSColor.cyan, // Seems to be ignored
+                         shadow: NSColor.magenta // Seems to be ignored
+            )
+        } else {
+            // Fallback on earlier versions
+            return defaultTheme()
+        }
+    }
 
     init(jsonObject dict: [String: AnyObject]) {
         let foreground = NSColor(jsonRgbaColor: dict["foreground"] as? [String: AnyObject] ?? [:])


XiEditor/EditViewController.swift
```diff
@@ -50,6 +50,8 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
 
     @IBOutlet weak var editViewHeight: NSLayoutConstraint!
     @IBOutlet weak var editViewWidth: NSLayoutConstraint!
+    
+    let appDelegate = NSApp.delegate as! AppDelegate
```
habovh: Not sure if this is the right way to access AppDelegate. Not even sure if I'm supposed to access the AppDelegate in a VC.

cmyr: I would avoid storing the variable, and instead just call `(NSApp.delegate as? AppDelegate)?.whatever()` in line, as needed.

cmyr:
Example of use when the previously active theme clashes:
screenshot 2018-10-03 13 35 14

jansol:
I think this should use the vibrant blending mode, covered in the [WWDC video, starting from around 28:30](https://developer.apple.com/videos/play/wwdc2018/218?time=1704) -- it's supposed to avoid this kind of issues.

Besides that, I think making this a special theme is the wrong approach. Instead I would make it a flag that when enabled uses the theme colors and vibrant blending to tint the transparent window. It should be fine even when switching between dark and light, but in case it's not (and even if it is, because this would be quite cool) there could be an option to specify two active themes (dark and light) and have xi-mac automatically switch when the system appearance changes.

cmyr:
Okay, I haven't dug into the mojave docs/talks yet so I defer to @jansol's suggestion!

habovh:
@cmyr I'm glad I came out with the same analysis as your high-level thought!

I originally had in mind what you just said in your *for now* paragraph. I was going for the option to enable the system dark/light mode, which would use one of core's themes for each corresponding state and override the actual app properties.

I also like the *long term idea* you have, to bundle 2 special themes for each corresponding system theme. However there's a few things I don't get. First, I couldn't find out how/where themes are stored/defined in core. I obviously need to understand this in order to bundle themes from xi-mac. Also, I guess a theme (syntax highlighting-speaking) has a fair amount of colors that have to work in harmony. I wouldn't say I'm a bad UI designer, but there's a high chance people are used to be able to choose their syntax-highlighting independently of their app UI.

What I'm thinking is similar to what @jansol said, we could consider the best of both worlds:
- Allow users to set a flag that they want the UI to match the system light/dark setting
- Allow them to choose which syntax highlighting they want to use independently

Here's a mockup of how the user could use this:

| Light system theme | Dark system theme |
| ---- | ---- |
| capture d ecran 2018-10-04 a 11 50 44 | capture d ecran 2018-10-04 a 11 49 50 |

Thinking long-term, we could imagine that users could import their own themes, flag them as made for dark or light backgrounds, and xi-mac can automagically use them depending on the system's preferences.

However, I do not think the vibrant blending mode is quite suitable for the use case we have. As I understand it, vibrant blending mode should be used to modify colors that are displayed on top of a potentially changing background color. So one might think it would be useful to update the syntax highlighting colors, but there's a few issues:
- vibrant blending mode should not be used for non-grayscale colors (read: do not use vibrant blending mode on colored items)
- the only colors we should override are the background, gutter, and default text. Using semantic `NSColors` is IMHO the way to go there.

Do you guys think the two active themes and the flag solution is the way to go there?

XiEditor/EditViewController.swift
habovh: Thanks for your feedback on this, always interesting to see various practices!

jansol:
> Do you guys think the two active themes and the flag solution is the way to go there?

I believe so. An interesting thing to try would be to allow the user to specify just one theme and then automatically invert its colors (or maybe just their brightness/value?) for the other appearance. Some syntax themes also work quite well on both bright and dark backgrounds, especially when the default color is automatically updated to match (Monokai for example looks quite nice against either background) so automatic inversion should be optional too, if it is supported. This is of course an optional bit of functionality at this stage.

habovh:
While inverting colors would keep the same contrast compared to the other type of background, it also breaks any color semantics the theme could've set. The rule of thumb here would be to invert only the grayscale colors, because they're supposedly the ones that are going to lack contrast against an inverted background. But again, there's no guarantee that it will actually look nice. I'm not a fan of trying to guess what the theme should look like. While I agree some themes can look good on either backgrounds, I still think that being able to choose one light and one dark theme would be better for the user.

cmyr:
A drive-by comment: something else to keep in mind is the "unified_titlebar" feature, and how that works with this?

cmyr:
also @habovh there's some info on how themes work here: https://github.com/xi-editor/xi-editor/issues/883

habovh:
I guess unified titlebar should update according the system theme **if** the flag is set to follow the system theme, otherwise we can still update it according to the theme if it *isDark*?

@cmyr thanks for the heads up! Will definitely read this!

jansol:
> While inverting colors would keep the same contrast compared to the other type of background, it also breaks any color semantics the theme could've set.

Hence value invert aka keep the hue & saturation but invert the lightness.

Anyway this is mostly a "could be nice" gimmick for people who are too lazy to define two separate themes.

habovh:
Oh sorry @jansol I didn't get the invert lightness only 🙈 My bad, this is definitely a could be nice feature!

XiEditor/EditViewController.swift
ollieatkinson: It's generally not good practice to reference the AppDelegate inside of your view controllers (directly).

How about using a protocol?

e.g.

```swift
protocol ThemeChangeDelegate {
    func themeChanged(name: String, theme: Theme)
}


...

extension AppDelegate: ThemeChangeDelegate { }

...

weak var themeChangeDelegate: ThemeChangeDelegate? = NSApp.delegate as? ThemeChangeDelegate

```

It will make it much easier to write a unit test and decouples the editor view.

terhechte: Adding to what @ollieatkinson wrote, the best option is to use dependency injection to add the delegate into the view controller:

Here, the main view controller is setting its `themeChangeDelegate` on the `ThemeViewController`, so it is just copying its dependency into the `ThemeViewController`
``` swift
class MainViewController: NSViewController {
  private var themeViewController: ThemeViewController?
  private var themeChangeDelegate: ThemeChangeDelegate?
  override func viewDidLoad() {
    super.viewDidLoad()
    self.themeViewController = ThemeViewController()
    self.themeViewController?.themeChangeDelegate = self.themeChangeDelegate
  }
}
```

Here, the AppDelegate sets up the `MainViewController` and installs itself as the dependency. That way, nobody ever accesses the AppDelegate except for the appdelegate itself.
``` swift
class AppDelegate: NSObject, NSApplicationDelegate, ThemeChangeDelegate {
  var mainViewController = MainViewController()
  func applicationWillFinishLaunching(_ aNotification: Notification) {
    mainViewController.themeChangeDelegate = self
  }
}
```

cmyr:
@habovh @jansol is there consensus here? I'm happy to more or less stay out of it, if you two agree on an approach you have my blessing. 🙈

jansol:
I think so. To recap:
* Have a flag for switching between theme-defined background color and NSAppearance
* User can select separate light and dark syntax themes
* Leaving one of those unset causes xi to derive it from the other one by inverting the lightness of the entire palette

What I'm not sure is what the best notation for this would be in the config.. Simply presence/absence of theme_light/theme_dark keys? Using a separate theme key? That one would require the lightness to be tagged directly in the theme, which would require some more refactoring.

cmyr:
We could just let the user pick whatever theme they want for 'light' and 'dark' and if it looks bad that's just user error?

ollieatkinson:
Would it be too much work for each theme to define a light and dark variant?

jansol:
It's mostly a question of knowing whether lightness should be inverted for the light or the dark variant or neither. It could be configured like this:

```toml
theme_light = "Tomorrow Night Eighties"
theme_light_invert = "true"
#theme_dark = "Tomorrow Night Eighties"
theme_dark_invert = "false"
```

We'll want some slightly more sophisticated logic for defaults though:
* If a theme variant is specified, default to `invert = "false"`
* If a variant is *not* specified, default to the other variant
  * if `invert` is not specified, default to `"true"`

This way it is enough to specify `theme_light = "InspiredGitHub"` and you get a reasonable default for both light and dark variants.

@ollieatkinson For bundled themes that is a reasonable thing to do. However, if you found a nice theme online and wanted to use that you'd end up having to manually create a copy and replace every single color. So yes, I'd say that would be needlessly much work, at least if automatic inverting turns out to look anywhere near as good as I expect.

XiEditor/EditViewController.swift
```diff
@@ -148,11 +150,15 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
 
                 statusBar.updateStatusBarColor(newBackgroundColor: self.theme.background, newTextColor: self.theme.foreground, newUnifiedTitlebar: unifiedTitlebar)
                 findViewController.updateColor(newBackgroundColor: self.theme.background, unifiedTitlebar: unifiedTitlebar)
-
-                if color.isDark && unifiedTitlebar {
-                    window.appearance = NSAppearance(named: NSAppearance.Name.vibrantDark)
+                
+                if #available(OSX 10.14, *) {
+                    // Let system decide, disregarding active theme color
```
ryanbloom: When `unifiedTitlebar == true`, we should be using the theme color instead of the system appearance here.

cmyr:
@habovh What's the status on this? Are you blocked on anything?

cmyr:
@habovh no worries! this is definitely not critical right now, if you get around to it eventually that's great, if someone else grabs it that's great too.

cmyr:
@habovh I'm going to close this for now; if you're interested in picking it up again at some point feel free to reopen!

Issue #279, opened 2018-10-01
when built from Xcode, display commit # in titlebar (or something?)


cmyr:
I consistently have several versions of xi-mac running, and I am regularly confused about which is which. It would be great if, when build from Xcode, xi-mac could display the HEAD hash in the menubar, or could inject an alternate app icon or something like this.

jansol:
I'd say make it specific to Debug builds (aren't those the default when building from Xcode? [seems to be easy to detect too](https://stackoverflow.com/questions/25916396/how-to-identify-if-swift-was-compiled-with-optimization)) Alternatively there might be a way to define variants of assets (i.e. app icon) based on build type (á la xamarin), or just a custom resource copy step to choose the right icon manually.

nangtrongvuon:
It could a part of the status bar, similar to Atom and VSCode in some sense.

terhechte:
I gave this a go in this PR:
https://github.com/xi-editor/xi-mac/pull/289

cmyr:
closed in #289

PR #289, opened 2018-10-03
Feature/279 when built from Xcode, display something

terhechte:
#279 

This PR includes two indicators to show the user that Xi-Mac is being run from within Xcode:

First, a `[Debug] ..` string in the window titlebar:

screenshot 2018-10-03 at 18 35 05

Second, a different icon that kinda mimics the look of the Xcode icon:

xi-debug-512 copy


XiEditor/XiWindowController.swift
```diff
@@ -0,0 +1,10 @@
+import AppKit.NSWindowController
```
cmyr: You can copy over the header from one of the existing files (with the current year)

terhechte:
@cmyr I broke something when I commited. The icon should also work now. I also added the license header.

cmyr:
@terhechte awesome, this is working for me now.

My only lingering concern is that I'm seeing two new compiler errors (Xcode 10). The first is about your `#if debug` causing an unreachable line, and the other is about a downcast to `NSWindowController`. If you'd like to clean those up I'd be happy to merge.

Also it's a little embarrassing that our debug logo maybe looks a bit better than our default logo.. 😅💦 (aka, good job on the logo)

terhechte:
@cmyr Thanks! Glad you like the icon :) I've fixed the compiler warnings. If there's interest, I could give the main icon a refresh as well. I'd love to do that.

nangtrongvuon:
There's definitely interest from me for that new icon :)

jansol:
See also xi-editor/xi-editor#51

That issue has a lot of suggestions but sadly no consensus has been reached. I think people expected @raphlinus to make an authoritative decision there but things are quite different now.

PR #293, opened 2018-10-07
Render background colors for text spans

ryanbloom:
Depends on xi-editor/xi-editor#894.

Some themes set a red background color for illegal characters. Xi doesn't currently render this, which can make the text illegible. This pull request aims to fix that.

### Complications with blending
My approach here was to simply render theme styles first, then find styles, then selection styles on top. With translucent find highlight and selection colors, this can look alright. Highlights and selections blend with the span's background color, but don't obscure it. Here's a theme I've been using for testing:

[One Light.tmTheme.zip](https://github.com/xi-editor/xi-mac/files/2454265/One.Light.tmTheme.zip)

With an opaque selection color (i.e. with the included syntect themes), this doesn't look so good; selections just override the span's background color. To make existing themes look better, we could add logic to blend only as needed, and use the opaque color the rest of the time. However, if we're moving away from syntect's themes anyway (xi-editor/xi-editor#883), it might make sense to just assume that themes have been adapted for xi.

One other consideration is that the two blending styles will look slightly different. For instance, if a theme wants a white background with a light gray selection color, the selection color could either be an opaque light gray (with custom blending logic) or a translucent dark gray. The former would lighten background spans; the latter (what I did here) darkens them.

Anyway, I'm not totally sure about this. Let me know if I'm headed in the wrong direction.

### DIFF
--- a/XiEditor/StyleMap.swift
+++ b/XiEditor/StyleMap.swift
@@ -165,17 +165,21 @@ class StyleMapState: UnfairLock {
     }
 
     func applyStyle(builder: TextLineBuilder, id: Int, range: NSRange, selColor: ARGBColor?) {
-        if id >= styles.count {
+        if id < 0 || id >= styles.count {
             print("stylemap can't resolve \(id)")
             return
         }
 
-        if id >= 0 && id < Style.N_RESERVED_STYLES {
-            builder.addSelSpan(range: convertRange(range), argb: selColor!)
+        if id < Style.N_RESERVED_STYLES {
+            // Ignore foreground color for selection and highlight styles
+            builder.addBgSpan(range: convertRange(range), argb: selColor!)
         } else {
             guard let style = styles[id] else { return }
             if let fgColor = style.fgColor {
                 builder.addFgSpan(range: convertRange(range), argb: colorToArgb(fgColor))
+            }
+            if let bgColor = style.bgColor {
+                builder.addBgSpan(range: convertRange(range), argb: colorToArgb(bgColor))
             }
             if let font = style.font {
                 builder.addFontSpan(range: convertRange(range), font: font)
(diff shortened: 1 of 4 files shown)


cmyr:
cc @scholtzan, who might be interested in this:

@ryanbloom Sorry for the delay! this looks reasonable, and maybe points in a direction I've been thinking about, which is to actually just remove selections & highlights from the spans entirely. Instead we would send this information as a separate field in the update.

This avoids a bunch of logic in core where we grab the selections and insert them into the spans; it also lets us avoid the RESERVED_STYLES.

Importantly, it _also_ lets us keep more of the semantic information of the selection/find results, which will be important for things like accessibility; it will be easier for a screen reader (for instance) to know what lines are selected.

We'll also at some point want to send _other_ information this way; for instance if we want to show compiler warnings/errors or other linting output inline, we could use the same mechanism.

cmyr:
So I'm certainly okay with this going in, but it isn't _currently_ going to work with most themes.

I would be okay with a workaround where we take the selection color and reduce it's alpha, or some similar hack?

cmyr:
Braindump:

- VSCode does something we might want to consider, which is that it uses textmate bundles for _syntax_ highlighting, but has its own format for UI theming. (`.tmTheme` files contain both syntax highlighting colors and UI theme stuff.) Although you're right that it doesn't seem to show error spans.

- _if_ we allow background color to be a thing that a syntax highlighter can define, we need some way of showing a selection or find result on a span that has a background color. I don't have strong feelings about what the best approach is, but if we want one, we'll want the other.

- One possible work around is that if we draw any background colors while resolving normal styles, we set some flag; and then in this case we take a different code path when drawing selections that figures out if a selection overlaps an existing background span and blends the color for that region?

I'm happy to defer to you on what the best approach from here is. Having the highlighter include background colors that we then ignore isn't good, so pretty much any decision from there will be an improvement.

ryanbloom:
I tried this in Sublime Text. Most of the default color schemes have opaque selection colors, which obscure any error spans (like this pull request does with syntect's themes). The Celeste scheme uses a translucent color, which works like the modified One Light theme I attached above. So it seems this implementation is more Sublime-compatible than I thought, at least with selections. It'll look better once we bundle the tweaked themes. I might try working on that at some point (unless @scholtzan has claimed the issue?)

Anyway, now that I see Sublime's behavior, I think this is probably the way to go. If/when you think it's ready, I'll resolve the merge conflicts.

cmyr:
@ryanbloom sure, lets do that.

I think tweaked themes is available; one thing I would look at is VSCode's [extension generator](https://code.visualstudio.com/docs/extensions/yocode); one of the things it can do is generate a new VSCode theme from an existing `.tmTheme`; we may want to try something similar?

PR #294, opened 2018-10-08
Add menu option to select a language

### LINKED ISSUE #282: Add menu option to select language
This ~will depend on https://github.com/xi-editor/xi-editor/issues/863 and is related to https://github.com/xi-editor/xi-editor/issues/866; and the two might be worked on together.~ is ready!

This menu should be populated when the `available_languages` notification is received, shortly after startup. Selecting a language from this menu should apply a checkmark indicator, and should send the `set_language` RPC (described in 866 above).

akxcv:
Added a language menu in `View > Language`

Closes #282 
~Is blocked by [#891](https://github.com/xi-editor/xi-editor/issues/891)~

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -171,6 +171,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
                                                   textColor: theme.foreground)
         }
     }
+    
+    var languageId: String?
 
     override init() {
         ValueTransformer.setValueTransformer(
@@ -284,6 +286,14 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
                 guard let doc = doc as? Document else { continue }
                 doc.editViewController?.themeChanged(name)
             }
+        }
+    }
+    
+    func languageChanged(viewIdentifier: String, languageId: String) {
+        DispatchQueue.main.async { [weak self] in
+            self?.languageId = languageId
+            let document = self?.documentForViewIdentifier(viewIdentifier: viewIdentifier)
+            document?.editViewController?.languageChanged(languageId)
         }
     }
 
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -295,6 +295,15 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
             }
         }
     }
+    
+    func availableLanguages(languages: [String]) {
+        DispatchQueue.main.async {
+            for doc in NSApplication.shared.orderedDocuments {
+                guard let doc = doc as? Document else { continue }
+                doc.editViewController?.availableLanguagesChanged(languages)
+            }
+        }
+    }
 
     func pluginStarted(viewIdentifier: String, pluginName: String) {
         let document = documentForViewIdentifier(viewIdentifier: viewIdentifier)
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -171,8 +171,6 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
                                                   textColor: theme.foreground)
         }
     }
-    
-    var languageId: String?
 
     override init() {
         ValueTransformer.setValueTransformer(
@@ -289,11 +287,10 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         }
     }
     
-    func languageChanged(viewIdentifier: String, languageId: String) {
+    func languageChanged(viewIdentifier: String, languageIdentifier: String) {
         DispatchQueue.main.async { [weak self] in
-            self?.languageId = languageId
             let document = self?.documentForViewIdentifier(viewIdentifier: viewIdentifier)
-            document?.editViewController?.languageChanged(languageId)
+            document?.editViewController?.languageChanged(languageIdentifier)
         }
     }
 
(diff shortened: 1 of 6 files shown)


XiEditor/EditViewController.swift
```diff
@@ -805,6 +814,22 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         }
         self.unifiedTitlebar = { self.unifiedTitlebar }()
     }
+    
+    public func availableLanguagesChanged(_ languages: [String]) {
+        let languagesMenu = NSApplication.shared.mainMenu!.item(withTitle: "Debug")!.submenu!.item(withTitle: "Language")!.submenu!;
```
cmyr: I always get a bit scared whenever I see this many `!`s. Can it happen that this RPC arrives before the menu has loaded, but the view controller hasn't? I'd _hope_ not.

In any case, I think it's probably fine, at least this way we'll find out if/when it happens.. 😏

akxcv: > Can it happen that this RPC arrives before the menu has loaded, but the view controller hasn't?
I'm not sure about that, but it never happened on my machine 😅 
Also, the same goes for `available_themes`, and it's been in the code for a while, so I think we're good here

XiEditor/AppDelegate.swift
```diff
@@ -171,6 +171,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
                                                   textColor: theme.foreground)
         }
     }
+    
+    var languageId: String?
```
cmyr: In general when I'm writing Cocoa stuff I adhere to the Cocoa naming conventions, which prefer spelling out whole names, so 'Identifier' instead of 'Id'. We aren't super consistent about this, so 🤷‍♂️, just a note.

XiEditor/AppDelegate.swift
```diff
@@ -286,6 +288,14 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
             }
         }
     }
+    
+    func languageChanged(viewIdentifier: String, languageId: String) {
+        DispatchQueue.main.async { [weak self] in
+            self?.languageId = languageId
+            let document = self?.documentForViewIdentifier(viewIdentifier: viewIdentifier)
```
cmyr: actually, what's going on here? Is there a reason we need to store the `languageId` at all outside of the particular view that's effected?

XiEditor/EditViewController.swift
```diff
@@ -127,6 +127,13 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
             updatePluginMenu()
         }
     }
+    
+    /// Current language used for syntax highlighting
+    var currentLanguage: String? {
```
cmyr: I might need to think this through a bit more, but I think we can safely just init this to "Plain Text". xi-mac, as a product, will always includes a bundled syntax highlighting plugin, which should always support this syntax?

XiEditor/EditViewController.swift
```diff
@@ -752,6 +770,24 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
+            if let currentLanguage = self.currentLanguage {
+                subItem.state = NSControl.StateValue(rawValue: (subItem.title == currentLanguage) ? 1 : 0)
+            } else {
+                subItem.state = NSControl.StateValue(rawValue: 0)
+            }
+        }
+    }
+    
+    // Gets called when active window changes
+    func updateMenuState() {
```
cmyr: yep this is nice and clean 👍

XiEditor/Main.storyboard
```diff
@@ -1015,6 +1014,20 @@ Gw
                                 <modifierMask key="keyEquivalentModifierMask"/>
                                 <menu key="submenu" title="View" id="SBh-Xs-jHf">
                                     <items>
+                                        <menuItem title="Theme" id="SUL-oD-daT">
```
cmyr: Okay so this is totally my fault (going back and rereading my comment about moving the Theme menu):

I think this change might make sense, but I'm not sure `Theme` should go under `View` (it might belong with preferences under the main 'XiEditor' menu?), and in any case it feels like a change that is pretty firmly unrelated to this PR, so it would make more sense to do it separately. Sorry for the confusion!

akxcv: Yeah, I thought about it being unrelated to the PR, but figured that if I mentioned that in the PR description it would be okay. But sure, let's do this separately, that would be better.

XiEditor/EditViewController.swift
akxcv: I think this is fine for now, although not fine in the future. What if we change how language identifiers are spelt, for instance? Like "text/plain" instead of "Plain Text" or something.

I think this logic (default language) belongs to the core, but currently, it's hard to implement because of the way we're loading languages there.

XiEditor/AppDelegate.swift
akxcv: You're right, that's weird. This shouldn't be in the code.

XiEditor/EditViewController.swift
akxcv: There's also an issue that we're sending empty language from the core down the wire (`language_changed {"view_id": "view-id-1", "language_id": ""}`), so whatever default we provide will be overwritten, and checking this in Swift code feels ugly. Maybe we should tackle this in a separate issue? Rethink how the core is loading languages?

cmyr: yep that sounds reasonable.

cmyr:
Thanks!

Issue #277, opened 2018-09-28
View controller should access config on creation


scholtzan:
This issue is related to https://github.com/xi-editor/xi-mac/pull/274
When creating the view controller `configChanged` is currently called, which is causing some problems. Instead created view controlleres should start with access to a config, which they use for initial setup; and then `configChanged` should only be called when things need to be updated.

PR #303, opened 2018-10-14
Set config for EditViewController in redrawEverything

scholtzan:
This is an attempt to fix https://github.com/xi-editor/xi-mac/issues/277

### DIFF
--- a/XiEditor/EditViewController.swift
+++ b/XiEditor/EditViewController.swift
@@ -188,7 +188,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
                 statusBar.updateStatusBarColor(newBackgroundColor: self.theme.background, newTextColor: self.theme.foreground, newUnifiedTitlebar: unifiedTitlebar)
                 findViewController.updateColor(newBackgroundColor: self.theme.background, unifiedTitlebar: unifiedTitlebar)
 
-                if color.isDark && unifiedTitlebar {
+                if color.colorSpace.colorSpaceModel == NSColorSpace.Model.RGB && color.isDark && unifiedTitlebar {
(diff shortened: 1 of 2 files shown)


XiEditor/EditViewController.swift
```diff
@@ -188,7 +188,7 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
                 statusBar.updateStatusBarColor(newBackgroundColor: self.theme.background, newTextColor: self.theme.foreground, newUnifiedTitlebar: unifiedTitlebar)
                 findViewController.updateColor(newBackgroundColor: self.theme.background, unifiedTitlebar: unifiedTitlebar)
 
-                if color.isDark && unifiedTitlebar {
+                if color.colorSpace.colorSpaceModel == NSColorSpace.Model.RGB && color.isDark && unifiedTitlebar {
```
scholtzan: This is a bit awkward. The reason why I check for the color model is, that when `redrawEverything` is first called and `unifiedTitlebar` is set, the color model is not RGB. In `isDark` values for `redComponent` etc are not available which will results in no window being shown. Maybe there is a better way to handle this?

lukakerr: @scholtzan I put in a fix for the color space in my sidebar branch https://github.com/xi-editor/xi-mac/blob/9cf3304bfc94227f9e898824b8d3f11287dc1f07/XiEditor/EditViewController.swift#L875

Perhaps I should open a different PR for that on its own. In the sidebar branch it was also causing the issue you mentioned about windows not showing

scholtzan: Oh nice! If you could open a separate PR which fixes the color space issue then that would be great!

Issue #299, opened 2018-10-10
track recently used/selected languages


cmyr:
Now that we have a 'languages' menu for selecting the syntax for the current document, there's one obvious improvement we could make: it would be nice if the top had a small 'Recently Used' section, which could show the 4/5 most recent items. This could be based on either what has been sent by "language_changed", or on what has actually been selected by the user. These could be persisted between launches by using `NSUserDefaults`.

Issue #275, opened 2018-09-27
Move to a protocol-based interface for core


cmyr:
A refactor that's long overdue: just as we define a client interface in `Client.swift`, we should define a core interface. All the various core methods would be presented through this interface; behind the surface it would wrap something like the existing `CoreConnection` class.

This will make it easy for us to change how the core connection is managed at some point in the future, and will also let us write tests in xi-mac that don't need to spin up a real instance of the core.

jkaan:
@cmyr Hi there! I could pick this one up. Would it for now just involve putting the public interface of the `CoreConnection` into a protocol and typehinting to that interface instead of the concrete implementation?

cmyr:
`CoreConnection` doesn't really have a typed interface at all; it just exposes functions for sending raw arbitrary RPCs. What we'd really like is to have a protocol that describes the legal, existing RPCs, and then implements that for `CoreConnection` by translating them into the raw RPCs.

We might also consider having a protocol for the `CoreView` interface, which describes just the RPCs that are scoped to a view.

There is a lot of RPC surface area (especially considering all of the movement commands) so I think it would be okay to go halfway, initially; for instance you could have a protocol method "raw_notification" that still lets you specify a command by it's (string) method name.

cmyr:
oh, also: there was an earlier attempt at something like this, in `Dispatcher.swift`. This predates my involvement in the project, and doesn't seem to have really ever been consistently used; I would expect this new work to also result in us removing that file, I think?

cmyr:
closed by #350.

PR #305, opened 2018-10-14
[WIP]: Track recently selected languages

demalex:
### Description

Implementation of enhancement #299 suggested by @cmyr. Every time user makes a language selection, the language is stored to `UserDefaults`. 

Using `UserDefaults` directly makes classes tightly coupled with it and makes it hard to maintain and test. To get rid of problems like this, `UserDefaults` was covered with additional abstraction. The idea is to keep settings provider in one place (`AppDelegate` for now) and inject it as dependency into other consumers. There are other places where `UserDefaults` is used directly and from my perspective it requires to be refactored, but such refactoring is out of this enhancement scope.

`SettingsProvider` is responsible for communication with `KeyValueStorage`. It executes save and fetch operations, while `SettingsProvidingProcessor` decorates `SettingsProvider ` and brings additional logic for save languages operation. Appropriate test cases also have been added.

### Screenshot

screen shot 2018-10-16 at 8 39 39 pm

### DIFF
--- a/Sources/XiEditor/Search.swift
+++ b/Sources/XiEditor/Search.swift
@@ -480,7 +480,7 @@ extension EditViewController {
     }
 
     @IBAction func selectionForReplace(_ sender: AnyObject?) {
-        document.sendRpcAsync("selection_for_replace", params: [])
+        xiView.selectionForReplace(caseSensitive: false)
     }
 
     @IBAction func multipleSearchQueries(_ sender: AnyObject?) {


cmyr:
Cool, thanks for this. Will have to give it a closer read, but looks nice and tidy at first glance, and thanks for the tests!

(if anyone else @xi-editor/xi-mac wants to take a look here that would be welcome!)

XiEditor/Storage/SettingsProvider.swift
```diff
@@ -0,0 +1,36 @@
...
+//  XiEditor
+//
+//  Created by Alex Demishkevych on 10/13/18.
+//  Copyright © 2018 Raph Levien. All rights reserved.
+//
+
+import Foundation
+
+protocol SettingsProviding {
+    func fetchingRecentlyUsedLanguages() -> [String]
```
jansol: This should probably be called `fetchRecentlyUsedLanguages` ("fetching" sounds very weird to me).

XiEditor/Storage/SettingsProvider.swift
```diff
@@ -0,0 +1,36 @@
...
+    func saveRecentlyUsed(languages: [String])
+}
+
+class SettingsProvider: SettingsProviding {
```
jansol: This architecture, it seems needlessly convoluted. Yes, it looks a bit like MVC but that falls apart because in practice you can only access the `SettingsProvider` through the `SettingsProviderProcessor`. I'm not sure what benefits this separation is supposed to bring.

Also having `SettingsProvider` and `SettingsProviding` which are almost the same thing is confusing. I'd rather call the protocol `SettingsProvider` or even `ISettingsProvider` (to clearly indicate that it's an interface/protocol) and this class something like `SettingsStore`, since all it does is provide access to the KeyValueStore. Or you could probably remove this class entirely since it's just a very thin wrapper over a dictionary.

XiEditor/Storage/SettingsProvider.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    enum Keys: String {
+        case recentlyUsedLanguages
+    }
+
+    private let keyValueStorage: KeyValueStorage
```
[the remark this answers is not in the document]
demalex: Hi @jansol, thanks for the comment! I have introduced `KeyValueStorage ` protocol in order to have ability easily switch implementations without touching client code.

First benefit of such approach is that It allows to choose whatever key-value storage is needed (UserDefaults, Dictionary etc.). It is a rare case to be honest, but still.
Second benefit goes when it is needed to mock and stub such storages in tests. The main disadvantage working with `UserDefaults` is that `UserDefaults` persists the data between tests launch. From my experience it is very easy to forget to remove user defaults suite or  some value and eventually find out that not related test cases are falling because of not properly cleaned up `UserDefaults`. So, it is very easy to mess it up.

Having such protocol allows to create mock `DictionaryStorage` and use it as storage in tests.

What do you think? Does it make sense?

XiEditor/Storage/SettingsProvider.swift
demalex: @jansol yes, `fetchRecentlyUsedLanguages` definitely sounds better 👍🏻

XiEditor/Storage/SettingsProvider.swift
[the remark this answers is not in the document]
jansol: It makes sense for mocking, yes. But I didn't see any case where this *actually* is anything other than `DictionaryStorage` (and no other implementations either), hence my suspicions.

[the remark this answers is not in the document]
demalex: @jansol yes, I agree. So, do we keep the protocol? I think it worth keeping it, at least for mocking purposes. I am open to any other suggestions if any.

XiEditor/Storage/SettingsProvider.swift
demalex: @jansol to be honest I did not get what you meant by `MVC` and access the `SettingsProvider` through the `SettingsProviderProcessor`. Could you please explain what did you mean?

The idea behind `SettingsProvider` is that it responsible for only for interacting with storage without any additional logic while `SettingsProvidingProcessor` decorates `SettingsProvider` adding level of logic.

This solutions are flexible from my point of view, because `SettingsProvider` implementation can be easily replaced with new one (for example new implementation could fetch settings from `SQLite` instead of `UserDefaults`) and injected into `SettingsProvidingProcessor`. So, final configuration has new provider with new storage and with same processor that contains unchanged languages processing logic. As the project evolves, `UserDefaults` may not be enough and switching to some type of data base could be done quite easy.

What do you think? Does it worth to have such solution?

Yes, naming definitely should be improved. I would not use `ISettingsProvide` since it is not a best practice for naming protocols in Swift. Also I am trying to use this guideline during API design [API Design Guidelines](https://swift.org/documentation/api-design-guidelines/).

XiEditor/Storage/KeyValueStorage.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  KeyValueStorage.swift
+//  XiEditor
+//
+//  Created by Alex Demishkevych on 10/14/18.
+//  Copyright © 2018 Raph Levien. All rights reserved.
+//
```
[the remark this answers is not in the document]
demalex: @jansol Updated

demalex:
@cmyr @jansol I have updated license headers and changed components naming a bit. Also I have added some info for conversations above. Please, have a look. Thanks!

cmyr:
I've given this a read through, and I do see @jansol's point: it feels like there's maybe _one_ more layer of abstraction than is necessary here? I like the idea of using a protocol for the main interface, but I did find the overall inheritance chain tricky to follow, in a way that didn't seem _totally_ necessary.

For instance: if `SettingsProcessor` was combined with `SettingsStore`, I don't think this would make the code any harder to maintain?

demalex:
@cmyr Yes, it makes sense. The initial idea was just to separate store interaction logic and business logic, so they both could be modified independently. But we can definitely combine this two classes and it wont make the codebase any harder to maintain. Lets keep it simple at least at this stage of project.

demalex:
@cmyr @jansol Guys, I have removed `SettingsProcessor` as separate entity. Now there are only `SettingsStore` that conforms to `SettingsProvider` protocol. As we discussed with @jansol `KeyValueStorage` protocol is good to have for testing and mocking, so this protocol remains the same. Also, I have attached a screenshot to initial post. Please, have a look! Thank you!

XiEditor/EditViewController.swift
```diff
@@ -850,23 +840,65 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
+
+    private func languagesMenu() -> NSMenu? {
+        let mainMenu = NSApplication.shared.mainMenu
+        let viewMenu = mainMenu?.item(withTitle: "View")?.submenu
+        return viewMenu?.item(withTitle: "Language")?.submenu
+    }
+
+    private func buildRecentlyUsedLanguagesSection(from languages: [String]) -> [NSMenuItem] {
+        var items: [NSMenuItem] = []
+        let titleItem = NSMenuItem(title: NSLocalizedString("Recently Used", comment: "Recently Used"), action: nil, keyEquivalent: "")
```
jansol: Might want to clarify the comment with something like "[Language]"

jansol:
So languages are supposed to show as in use in both "recently used" and the general list at the same time? That seems like reasonable behavior, but I don't see where the recently used one is marked as in use in the code. Missed when committing?

XiEditor/EditViewController.swift
demalex: @jansol I did not get your comment. Could you please explain?

jansol: The comment is meant to provide context to translators so they don't need to look at (and understand!) the code when writing translations. (They should only need to work with the strings file)

Actually on second thought, "Language list subtitle" would be an even better description.

demalex:
@jansol Hmm, no in use language is supposed to be selected only in general list section. I am sorry, but I think screenshot is misleading and I even do not know how I made it. 
Do you think it makes sense to show language as selected in both sections?

jansol:
Yeah, I think it's more consistent that way.

demalex:
@cmyr @jansol Updated screenshot and implementation. Please, have a look!

demalex:
@jansol Great! Thanks a lot! As for that indentation, I just had a look at macOS system implementations and the indentation is used in very rare variety of cases, so I will probably remove it.

Here are few examples from system implementation:
screen shot 2018-10-17 at 11 46 48 am
screen shot 2018-10-17 at 11 48 03 am

Also, do you think we need to add available languages title? I think it will be more consistent that way. What do you think?

jansol:
Yeah looks like not indenting is more in line with the rest of the system. (And we can always put the indentation back if it turns out problematic)

I don't think a subtitle for the available languages is necessary - the menu entry that opens this menu is already called `Language` so that seems quite obvious...

cmyr:
I agree that the indentation might not be necessary.

I've played around with this, and I do have a couple notes:

- Ideally the list would always be the five most recently used languages. However, if I select a language that's already recently used, its state isn't updated; so if I select the fifth most recently used language, and then select a new language, the language I was just using is removed from the list. My solution might be to always have the recents list be in descending order, most recently used to last?

- I was able at some point to get a language called 'Language' added to the recents list. This was on first launch. I'm not sure how I did this. 😕

Overall this looks good, and feels like a nice improvement; I look forward to getting it merged. 👍

XiEditor/Storage/KeyValueStorage.swift
```diff
@@ -0,0 +1,23 @@
+// Copyright 2016 The xi-editor Authors.
```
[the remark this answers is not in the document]
demalex: @cmyr 😆 You are right. I have just copy pasted the header. We might need to have custom header and automatically add it to all newly created source files. What do think?

XiEditor/Storage/KeyValueStorage.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+protocol KeyValueStorage {
```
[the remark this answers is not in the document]
demalex: @cmyr I will add a comment 👍🏻

demalex:
@cmyr I just played with the application and feature for a while and did not have a change to reproduce the issue with initial `Language` value. Could you please provide more details?

As for descending ordering I think it would be better to have it. So, I will apply change and let you know when it is ready! Thanks.

XiEditor/Storage/KeyValueStorage.swift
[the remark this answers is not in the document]
cmyr: This is an Xcode feature? I'd forgotten about that, it could be helpful though!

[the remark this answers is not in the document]
jansol: It is.

demalex:
@cmyr I have updated implementation. Please, have a look at it! Now most recent language appears at the top of the list and pushes previously used languages to the bottom. If one of the recently used languages is selected it is moved to the top of the list again. Thus, recently used languages list represents ordered history of languages which were used.

demalex:
@cmyr Yes, I can see the issue. It appeared because I have changed Language menu building logic a bit. 
From what I see, when application started it receives `selected languages` and `available languages` notifications. Bu, when main window is closed and then opened again only `selected language` notification is received. Could you please tell, whether it is right behavior?

cmyr:
@demalex Ah, right, this makes sense.

What we should do is handle `available_languages` in the app delegate, and then make it accessible from there to all view controllers, instead of trying to handle it in the view controller itself.

basically: `available_languages` is really a global setting, but `selected_language` is scoped to the individual document.

demalex:
@cmyr Initially I had same thought, but then I decided to clarify this behaviour. Thanks for the explanation! 

Basically I renamed `SettingsStore` to `LanguagesStore` and now it can provide recently used languages as well as available languages. I think `Settings` it is another entity and it requires separate abstraction to be created (it could be very similar to `LanguagesStore`).

I do not think that it is a good idea to keep all shared objects in `AppDelegate` and get access them by calling to shared application and then make a cast to `AppDelegate`. Such approach violates design rules and makes code hard to maintain, change and test. I would rather pass dependencies explicitly, but it requires code to be refactored and such improvement out of this feature scope. 

Nevertheless, I have updated the implementation. Please have a look and let me know what you think.

Thanks!

XiEditor/Storage/LanguagesStore.swift
```diff
@@ -0,0 +1,63 @@
...
+
+    func fetchRecentlyUsedLanguages() -> [String] {
+        return keyValueStorage.stringArray(forKey: Keys.recentlyUsedLanguages.rawValue) ?? []
+    }
+
+    func saveRecentlyUsed(language: String) {
+        
+        let newLanguage = language.trimmingCharacters(in: .whitespacesAndNewlines)
+        
+        if newLanguage.count == 0 {
```
mmatoszko: This check seems like a perfect job for a `guard` 😄 
`guard newLanguage.count > 0 else {
    return
}`

XiEditorTests/Mocks/DictionaryStorage.swift
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+class DictionaryStorage: KeyValueStorage {
+
+    var dictionary: [String : Any]
```
mmatoszko: why not private?

demalex: @mmatoszko `DictionaryStorage` should be used for tests when you need to mock object that conforms to `KeyValueStorage` protocol. It is added to Tests target only.

XiEditor/Storage/LanguagesStore.swift
demalex: @mmatoszko you think `if` can't handle it? 😎 It is not a big difference in this situation what to use. I think it easy to read simple `if` than `guard`. `guard` is good when you need to declare variables but do not want to increase cyclomatic complexity. Yes, compiler checks and if there is no `return` at some point it breaks compilation, but in such case (having tests) it is not so critical.

mmatoszko: totally agree :+1:

cmyr:
@demalex I definitely agree that our use of app delegate is pretty sketchy, and I'm open to refactoring. One related project is #275, which would move most of the RPC handling stuff out of the app delegate as well. 

Have been a bit distracted the past few days, but will try to give this another proper review either tonight or tomorrow. Thanks for your patience!

demalex:
@cmyr Thanks a lot and take your time! Let me know if you think that something needs to be improved.   I will have a look at another task for now.

demalex:
@cmyr Thanks a lot and take your time! Let me know if you think that something needs to be improved.   I will have a look at another task for now.

XiEditor/EditViewController.swift
```diff
@@ -133,11 +133,10 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
     /// Current language used for syntax highlighting
-    var currentLanguage: String? {
-        didSet {
-            updateLanguageMenu()
-        }
-    }
+    private var currentLanguage: String = ""
+
+    /// Object provides available and recently used languages
+    var languagesProvider: LanguagesProvider!
```
cmyr: Okay, so I've been thinking about this a bit:

I think it's important that we have a clear sense of what the source of truth is for language information. I'm thinking of the situation where (theoretically) a new plugin is loaded in core that adds a new language: What happens in this case? As currently implemented, the `LanguagesProvider` in the `AppDelegate` would get updated, but the individual views wouldn't end up being notified?

There are a variety of approaches to this. One simple one would be to have all access to the `languagesProvider` be explicitly shared; the view controller would be returning a reference to the document, which would be returning the shared object in the AppDelegate. Then when the state changed, we would iterate through the view controllers and notify them of the state change, and they can reload menus etc as needed.

Alternatively, when the state gets set in core we could go and set the new `languageProvider` on each view controller, and then handle any updates in a `didSet` method.

Another confusion here is that because `LanguagesStore` is a class, everyone should currently be sharing a single instance, and so the state _should_ be in sync; but `LanguagesProvider` isn't declared as a class protocol, so if it were implemented as a struct this would no longer be true?

Another thing I might think about is whether this state belongs more in the document or in the view controller. I think it might be reasonable to have it only be in the document? I could go either way, though.

XiEditor/Storage/LanguagesStore.swift
```diff
@@ -0,0 +1,63 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+protocol LanguagesProvider {
+    func fetchRecentlyUsedLanguages() -> [String]
+    func saveRecentlyUsed(language: String)
+    var availableLanguages: [String] { get set }
```
cmyr: I'm not sure if we really want the `set` method exposed here? `availableLanguages` should only really be set in response to a core RPC, so exposing the setter to other users feels a bit iffy.

demalex: @cmyr You are absolutely right. I think we need to change data flow a bit. I have some ideas and will share it soon.

XiEditor/EditViewController.swift
demalex: @cmy Thank you for review! 

You are right, class protocol constraint definitely needs to be added to `LanguagesProvider` 👍🏻 

Ideally `ViewController` should not contain any state. ViewController is an entity of UI / Presentation layer and its main responsibility to be a glue between model (view-model) and view and handle view's lifecycle. I definitely would not keep state in view controller and most likely it is Document object's responsibility.

I have been thinking about it a bit and I came up with possible solution for the problem. The idea is to have `LanguagesStore` as single source of truth and listeners which can listen to notifications about the updates. The `LanguageStore` itself can be updated by receiving notification from the core. I have created high level UML diagram to make it a little bit clear. Please, have a look:

languages store-2

Diagram is high level in some cases (Document and XiDocumentController, Core), but still brings visibility how data flows from presentation layer to core and to `LanguagesStore`.

What do you think?

cmyr: So I think we're broadly on the same page; what we want is a single source of truth and a notification mechanism.

We have options in terms of what our notification mechanism is. The simplest is that when the state changes, we iterate through our documents and call some `stateChanged()` function; then each document goes and checks the new state and performs some update logic. Another reasonable option is to lean on the Cocoa `NotificationCenter` infrastructure, which you're doing here. In _this_ case, though, I think it makes more sense to pass some reference type containing the state along as the notification's object; so in this case I would register for, say, the 'availableLanguagesChanged' notification, and the notification would contain a reference to some type, say, `CoreState`, (which can be a protocol) which has a member `var availableLanguages { get }`.

I think this is mostly what you're suggesting? The main thing I want to push is that we'd like to balance the benefits (clear separation of responsibilities, clear flow of data) against the costs of increasing architectural complexity; so whatever we do here should be something that we anticipate being the general model for communicating changes to shared state.

demalex: @cmyr Yes, its true! This is what I am suggesting. Please, take a note that `LanguageStoringNotification` is actually an interface. Interface name is misleading and will be changed. Having interface makes a loose coupling and allows to do checking during compile time, while `NSNotification.Name` is actually a `String` and it is easy to break an implementation. So, I would not use `NotificationCenter`.

So, the basic idea is very simple: 

1. Subscriber implements protocol and be subscribed for receiving updates from `LanguagesStore`
2. `LanguagesStore` receives updates from the `Core`, updates it's state and notifies all subscribers by iterating over the collection and calling appropriate function
3. As you mentioned above, updated state of the store (as a type of lets say `LanguageStoringOutput`) can passed to the function as parameter

What do you think? Does it make sense?

Architecture is what allows you to manage complexity and makes it easy to adapt the code base for new requirements. So, there is always some level of architecture complexity. But, I agree, there should be a trade off.

Please, let know what do you think about my suggestion. Thank you!

cmyr: Okay, thanks for the explanation. I think we're on the same page, and this approach sounds reasonable. 👍

cmyr: My one thought is that we might want to eventually expand this so that we have a single mechanism for various types of state (themes are an obvious example), so I'd keep that in mind; for instance we might want multiple types of events that can be subscribed to, _or_ we might want them to be a package deal.

demalex: @cmyr Thank you very much for the feedback! Yes, I am also thinking about generic solution that can be used in variety of cases. I was busy at work these days and going to have look at it on the weekend. Thanks!

cmyr:
@demalex just checking in, if this is ready for a review just :bell: me!

demalex:
@cmyr Sorry for the delay in answering. I am busy at work lately, but I am moving on with implementation. I hope, I can finish it soon. Let me know if it acceptable.

cmyr:
@demalex absolutely, work at whatever pace is good for you. I'm generally more worried that things are blocked on review or some question I've missed, so wanted to make sure that wasn't the case.

cmyr:
@demalex are you still making progress on this? It feels like it's really close, and it would be nice to get this feature done, but no worries if you're busy. 😃

demalex:
@cmyr Hi! Excuse me that it is taking so long to complete this. I a little bit overloaded at work at the end of the year, but I am making a progress and submit a new patch set soon. Hope it works for you.
Happy holidays! 🎄

cmyr:
@demalex amazing, glad to hear you're still hacking away. Happy holidays!

demalex:
Closed it due to big difference between master and feature branch. Will resubmit it soon

jansol:
@demalex ping?

Issue #308, opened 2018-10-19
Showing dot files


AlanFoster:
From what I can tell, when you attempt to open a dot file - it doesn't show up in the file explorer.

## Reproduction steps

- Create a file called `.travis.yml`
- Open the file explorer within Xi
- Attempt to edit the dot file

## What happens

There is no `.travis.yml` file within the explorer

## What I expected to happen

I can click the `.travis.yml` file and open it successfully for modification

AlanFoster:
I decided to learn how the NSDocumentController/NSDocument/NSWindowController pattern worked, but I don't believe there's a way to programmatically show hidden files.

There is however a short cut that can show hidden files:

> Press and hold the ⌘ + shift and . keys at the same time.

lukakerr:
Just an aside: if you implement your own `openDocument` method (which is what i've done while working on the sidebar) then you can programatically show hidden files in the open file dialog. So when the sidebar (eventually) gets merged in, hidden files will be shown

AlanFoster:
Interesting! I think I've found a way to show hidden files then :+1:

PR #310, opened 2018-10-21
Show hidden files

AlanFoster:
## Before

The open dialog would not show hidden files, such as `.travis.yml`

## After

All hidden files are to shown to the user when the open dialog is presented

Note: I'm not an iOS developer, I hope this is roughly correct 🤞 

---

Closes https://github.com/xi-editor/xi-mac/issues/308

### DIFF
--- a/XiEditor/XiDocumentController.swift
+++ b/XiEditor/XiDocumentController.swift
@@ -106,6 +106,11 @@ class XiDocumentController: NSDocumentController {
         super.removeDocument(document)
     }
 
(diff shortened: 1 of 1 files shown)


cmyr:
@lukakerr I saw that you have a commit handling this in #232, so maybe you can make the call as to whether this is the correct approach / will be easy to incorporate into that work?

lukakerr:
Sorry for the multiple reviews above! The GitHub outage today caused my review(s) not to show up, so I was under the impression that they just didn't go through. I'll merge this in though, thanks :)

Issue #313, opened 2018-10-24
Add CLI so xi can be used as your $EDITOR


mads-hartmann:
I want to start using xi for smaller text-editing tasks just to see how it feels. Specifically I want to use it to edit git commit messages.

xi currently bundles a `xi` bash script that simply invokes `open -a XiEditor "$@"`. To be able to use this as my git $EDITOR we'd need to extend it with a few more capabilities.

- Read from stdin
- Add a --wait flag. The `-W` flag to `open` doesn't quite get you all the way as it waits for the entire application to close rather than the window.

I'd be happy to help

cmyr:
@mads-hartmann this is totally something we'd like, and if you wanted to explore that would be great! I'm not totally sure what would be involved, but I suspect it will require writing a custom little binary?

mads-hartmann:
@cmyr Awesome, I'll have a look. I believe you're right, at least that's how Textmate [does it](https://github.com/textmate/textmate/blob/master/Applications/mate/src/mate.cc). I haven't written a single line of Rust before but I'll give it a go 😉

cmyr:
@mads-hartmann Although rust is definitely a reasonable choice, I think swift would also make sense? This is going to be mostly interfacing with the mac app.

mads-hartmann:
@cmyr Of course, you're right. Haven't written a single line of Swift either though 😉

liamrosenfeld:
I've gotten a basic CLI written on my [fork](https://github.com/liamrosenfeld/xi-mac/tree/cli). It's currently able to open a given file in Xi. I'm still trying to figure out how to implement --wait.

liamrosenfeld:
[Heres](#358) the CLI PR. It currently still needs --wait functionality.

PR #314, opened 2018-10-24
[WIP] Add CLI

mads-hartmann:
This is a WIP PR for #313. Opening it now so we can discuss things along the way. Right now it's just a placeholder CLI.

Other editor CLI implementations that we can draw inspiration from

- [mate](https://github.com/textmate/textmate/blob/master/Applications/mate/src/mate.cc) for TextMate
- [code](https://github.com/Microsoft/vscode/blob/master/src/vs/code/node/cli.ts) for VSCode

Things that are missing

- [ ] Tie the CLI into the existing build process
- [ ] Open Xi with the given path
- [ ] Some sort of IPC between the CLI and the Xi app so the app can communicate when a file is closed (to implement the `--wait` flag)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 xcuserdata/
 .DS_Store
 build/
+.build/
 .env
 xi-editor/
--- a/xi/Package.resolved
+++ b/xi/Package.resolved
@@ -1,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "SwiftPM",
+        "repositoryURL": "https://github.com/apple/swift-package-manager.git",
+        "state": {
+          "branch": null,
+          "revision": "235aacc514cb81a6881364b0fedcb3dd083228f3",
+          "version": "0.3.0"
+        }
+      }
+    ]
+  },
+  "version": 1
+}
--- a/xi/Package.swift
+++ b/xi/Package.swift
@@ -1,0 +1,16 @@
+import PackageDescription
+
+let package = Package(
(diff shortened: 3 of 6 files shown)


dustinknopoff:
I'd like to help but don't really know Swift or Cocoa.

This is my understanding of how implementing this would work:

1. Parse command line arguments (which you've mostly setup)
2. Send to AppDelegate? [Launching to AppDelegate](https://stackoverflow.com/questions/24046164/how-do-i-get-a-reference-to-the-app-delegate-in-swift)
3. Based on arguments call certain methods of XiDocumentController. (i.e. --line would call `AppDelegate!.documentController.openDocument(arg, true, completion handler).EditViewController?.scrollTo(line: arg, col: 0)`

mads-hartmann:
@dustinknopoff I don't know Swift either 😅I think you've gotten the steps right. Please feel free to hack around on this branch 👍

dustinknopoff:
What branch is this on?

mads-hartmann:
@dustinknopoff It's on the ci branch in my fork. https://github.com/mads-hartmann/xi-mac/tree/cli. I'll happily add you as a contributor if that makes collaboration easier :+1:

liamrosenfeld:
If the CLI is in a separate target I'm not sure if it would be possible to send a call directly to app delegate. [Launch Services](https://developer.apple.com/documentation/coreservices/launch_services) though looks like a very promising API for this type of task.

cmyr:
closing this; a new cli was added in #358

PR #317, opened 2018-10-29
Marker Bar

scholtzan:
This adds support for showing markers on top of the scroll bar. This allows to show indicators for find highlights, errors, ...

screen shot 2018-10-28 at 20 07 00

## Todo

- [x] Only show when scrollbar is shown (XCode behaviour)? Or show always (IntelliJ behaviour)? Or have a configuration for that?
- [x] Refine marker positioning
- [ ] When hovering show description?
- [ ] Find marker color should be defined in theme (required support of custom attributes)
- [ ] ...?

### DIFF
--- a/Sources/XiEditor/MarkerBar.swift
+++ b/Sources/XiEditor/MarkerBar.swift
@@ -29,7 +29,7 @@ class MarkerBar: NSScroller {
     var backgroundColor: NSColor = NSColor.red
     let markerLayer = CAShapeLayer()
     var overlayScrollerLayer: CALayer?
-    weak var markerDelegate: EditViewController?
+    weak var markerDelegate: MarkerDelegate?
 
     override func viewWillDraw() {
         super.viewWillDraw()
@@ -70,7 +70,7 @@ class MarkerBar: NSScroller {
     func drawMarkers() {
         guard let markerBarHeight = layer?.bounds.height,
             let maxScrollerWidth = layer?.bounds.width,
-            let totalLines = markerDelegate?.lines.height,
+            let totalLines = (markerDelegate as? EditViewController)?.lines.height,
             let width = overlayScrollerLayer?.bounds.width else { return; }
 
         let path = CGMutablePath()
--- a/Sources/XiEditor/MarkerBar.swift
+++ b/Sources/XiEditor/MarkerBar.swift
@@ -32,17 +32,17 @@ class Marker {
 class MarkerBar: NSScroller {
     var markers: [Marker] = []
     var backgroundColor: NSColor = NSColor.red
-    var parent: EditViewController!
-    var markerLayer = CAShapeLayer()
-    var overlayScrollerLayer: CALayer?
+    let markerLayer = CAShapeLayer()
+    var overlayScrollerLayer: CALayer!
+    weak var markerDelegate: EditViewController?
 
     override func viewWillDraw() {
         super.viewWillDraw()
 
         // last child layer will be used for drawing the overlay scroller
-        overlayScrollerLayer = self.layer?.sublayers?.last
-        overlayScrollerLayer?.addObserver(self, forKeyPath: "opacity", options: .new, context: nil)
-        self.layer?.addSublayer(markerLayer)
+        overlayScrollerLayer = self.layer!.sublayers!.last
+        overlayScrollerLayer.addObserver(self, forKeyPath: "opacity", options: .new, context: nil)
+        self.layer!.addSublayer(markerLayer)
     }
 
     override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {
@@ -51,7 +51,7 @@ class MarkerBar: NSScroller {
         }
     }
 
-    var markerHeight: CGFloat = 1.0
+    static let markerHeight: CGFloat = 1.0
 
     override class var isCompatibleWithOverlayScrollers: Bool {
         return true
@@ -69,18 +69,21 @@ class MarkerBar: NSScroller {
 
     func drawMarkers() {
         let path = CGMutablePath()
-        let totalLines = CGFloat(parent.lines.height)
-        let markerBarHeight = self.layer?.bounds.height
-        let maxScrollerWidth = self.layer?.bounds.width
+        let totalLines = CGFloat(markerDelegate!.lines.height)
+
+        print(totalLines)
+
+        guard let markerBarHeight = layer?.bounds.height,
+            let maxScrollerWidth = layer?.bounds.width,
+            let width = overlayScrollerLayer?.bounds.width else { return; }
 
         for marker in markers {
             markerLayer.fillColor = marker.color.cgColor
 
-            let width = (overlayScrollerLayer?.bounds.width)!
-            let x = maxScrollerWidth! - width
-            let y = ((CGFloat(marker.line) - 1) / totalLines) * markerBarHeight!
+            let x = maxScrollerWidth - width
+            let y = ((CGFloat(marker.line) - 0.5) / totalLines) * markerBarHeight
 
-            path.addRect(CGRect(x: x, y: y, width: width, height: markerHeight))
+            path.addRect(CGRect(x: x, y: y, width: width, height: MarkerBar.markerHeight))
             path.closeSubpath()
         }
 
--- a/Sources/XiEditor/MarkerBar.swift
+++ b/Sources/XiEditor/MarkerBar.swift
@@ -22,11 +22,6 @@ class Marker {
         self.line = line
         self.color = color
     }
-
-    @available(*, unavailable)
-    required init?(coder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
-    }
 }
 
 class MarkerBar: NSScroller {
--- a/Sources/XiEditor/MarkerBar.swift
+++ b/Sources/XiEditor/MarkerBar.swift
@@ -51,7 +51,7 @@ class MarkerBar: NSScroller {
         }
     }
 
-    var markerHeight: CGFloat = 0.5
+    var markerHeight: CGFloat = 1.0
 
     override class var isCompatibleWithOverlayScrollers: Bool {
         return true
--- a/Sources/XiEditor/MarkerBar.swift
+++ b/Sources/XiEditor/MarkerBar.swift
@@ -15,7 +15,6 @@
 import Cocoa
 
 class Marker {
-    // let description: String
     let line: Int
     let color: NSColor
 
@@ -39,6 +38,8 @@ class MarkerBar: NSScroller {
 
     override func viewWillDraw() {
         super.viewWillDraw()
+
+        // last child layer will be used for drawing the overlay scroller
         overlayScrollerLayer = self.layer?.sublayers?.last
         overlayScrollerLayer?.addObserver(self, forKeyPath: "opacity", options: .new, context: nil)
         self.layer?.addSublayer(markerLayer)
--- a/Sources/XiEditor/MarkerBar.swift
+++ b/Sources/XiEditor/MarkerBar.swift
@@ -33,16 +33,21 @@ class MarkerBar: NSScroller {
     var markers: [Marker] = []
     var backgroundColor: NSColor = NSColor.red
     let markerLayer = CAShapeLayer()
-    var overlayScrollerLayer: CALayer!
+    var overlayScrollerLayer: CALayer?
     weak var markerDelegate: EditViewController?
 
     override func viewWillDraw() {
         super.viewWillDraw()
 
-        // last child layer will be used for drawing the overlay scroller
-        overlayScrollerLayer = self.layer!.sublayers!.last
-        overlayScrollerLayer.addObserver(self, forKeyPath: "opacity", options: .new, context: nil)
-        self.layer!.addSublayer(markerLayer)
+        if (overlayScrollerLayer == nil) {
+            overlayScrollerLayer = self.layer!.sublayers!.last
+            overlayScrollerLayer?.addObserver(self, forKeyPath: "opacity", options: .new, context: nil)
+            self.layer!.addSublayer(markerLayer)
+        }
+    }
+
+    deinit {
+        overlayScrollerLayer?.removeObserver(self, forKeyPath: "opacity")
     }
 
     override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {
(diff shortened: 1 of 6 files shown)


jansol:
> Only show when scrollbar is shown (XCode behaviour)? Or show always (IntelliJ behaviour)? Or have a configuration for that?

I'd say whether showing the markers when there is no scrollbar makes sense boils down to whether users remember (roughly) where in the file they are at all times. The main use I see for showing markers without a scrollbar is to provide a hint as to how far one needs to scroll and thus how fast of a movement on the touchpad is needed.

In long files I certainly do not usually have a particularly accurate sense of that so the xcode way is fine. If we had a minimap of the document that was always visible, things would be different. (although in that case I'd argue that the markers should be shown on the minimap anyway)

XiEditor/MarkerBar.swift
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @objc private func hideMarkerBar() {
+        NSAnimationContext.runAnimationGroup({_ in
+            NSAnimationContext.current.duration = 0.5
+            markerBar.animator().alphaValue = 0.0
+        }, completionHandler: {})
+    }
+
+    override func scrollWheel(with event: NSEvent) {
```
scholtzan: This is probably not the perfect way to do it, but I tried subclassing `NSScroller` which resulted in very weird rendering of the scroll bars.

scholtzan:
I was poking around a bit more in this and tried to make the marker bar disappear when there is no scrollbar. For this, I was trying to sub-class `NSScroller` which resulted in weird rendering issues (the scrollbars didn't disappear anymore and didn't look right). I ended up doing some workaround based on scroll events which I am not really happy about. 
I'd appreciate if anyone has any hints/ideas of how to solve this or about better ways to implement the marker bar.

nangtrongvuon:
I played with this for a while, and couldn't really come up with anything.. Even with the current implementation, I couldn't get the markers to hide. 

I think XCode's implementation subclasses `NSScroller`, but I did run into the same weird rendering issues you did trying to subclass `NSScroller` myself. 

I think some ideas worth trying could be posting a notification for `hasVerticalScroller` and hiding the marker bar view when it changes, or maybe trying different approaches to overriding the NSScroller to overcome the rendering issues - I'll try to do this when I have some more time.

cmyr:
I think @nangtrongvuon is on the right track. If you look at `NSScrollView`, it has a `verticalScroller` property which you can set. This can be any subclass of `NSScroller`.

So I think we want to create a custom class of `NSScroller`, and then set this to be the scroller used by our scrollview. You can do this in Xcode, like so:

custom_scroller

The tricky part here is figuring out how to sync the find state with the relevant scroller. My first thought would be to have some `FindResultsDisplayDelegate` protocol that is implemented by the scroller (or the scrollview, which could forward things to the scroller?) and then have a `weak var displayDelegate: FindResultsDisplayDelegate?` on the find view controller. This doesn't feel great though, so maybe there's another approach that would be good, using notifications?

scholtzan:
Right, I tried subclassing `NSScroller`:
```
class CustomScroller: NSScroller {}
```

but then the scroll bars are rendered incorrectly and don't fade out anymore:

screenshot 2018-11-09 at 13 31 50

It feels buggy, but there might be some weird/clever workaround.

Notifications feel like a good idea. I'll try to see how it works out using those.

cmyr:
Ah interesting. this could be related to layer backing or something similar, I'll take a peek at that

scholtzan:
Another interesting thing in that context: currently the marker bar is in a separate view that is rendered over the scrollbar. If the width of that view is >= 15.0, then the same rendering issues as in the screenshot happen.

XiEditor/MarkerBar.swift
```diff
@@ -0,0 +1,128 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+class MarkerBar: NSView, CALayerDelegate {
+    var markers: [Marker] = []
+    var backgroundColor: NSColor = NSColor.clear
+    var parent: EditViewController!
+
+    var markerBarWidth: CGFloat = 16.0 {
```
scholtzan: If this is 15.0 or smaller then some weird rendering issues with the scroll bars happen

cmyr:
Looks like a custom `NSScroller` subclass will need the following:

```swift
    override class var isCompatibleWithOverlayScrollers: Bool {
        return true
    }
```

If I do this I get the expected drawing behaviour.

scholtzan:
Great, thank you! It works :)

scholtzan:
Interesting... If `isCompatibleWithOverlayScrollers = true` then `draw(_ dirtyRect: NSRect)` is not called. It is possible to override `drawKnob()` but then indicators can only be drawn on top of the knob and not on the other parts of the scrollbar.

scholtzan:
Overriding `updateLayer` doesn't change anything. It seems to get called only once when the scroller is initialized.

jansol:
Seems like [`drawKnobSlot`](https://developer.apple.com/documentation/appkit/nsscroller/1523622-drawknobslot) might be the right thing to override? Not sure how this behaves wrt updates though.

If it's not possible to use the scroller background for the marker bar I would also be fine with a separate view that is pinned to the right (all the way under the scrollbar when that exists).

scholtzan:
I haven't had much luck with overriding `drawKnobSlot`. Doing that only stuff that is drawn right on top of the knob is visible.
I figured out some workaround: I can observe the opacity value of the layer that draws the knob and then update the opacity value of the layer drawing the markers accordingly.

cmyr:
@rakaramos @mmatoszko I'd be curious if either of you have any thoughts on this?

Sources/XiEditor/MarkerBar.swift
```diff
@@ -0,0 +1,89 @@
...
+        self.layer?.addSublayer(markerLayer)
+    }
+
+    override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {
+        if keyPath == "opacity" {
+            markerLayer.opacity = change![NSKeyValueChangeKey.newKey] as! Float
+        }
+    }
+
+    var markerHeight: CGFloat = 1.0
```
cmyr: again this can be a `let`; and if it isn't variable between instances of this class we might make it a class constant, e.g. `static let markerHeight: CGFloat = 1.0`; this would be referenced in an instance as `MarkerBar.markerHeight`.

Sources/XiEditor/MarkerBar.swift
```diff
@@ -0,0 +1,89 @@
...
+    func setMarker(_ items: [Marker]) {
+        markers = items
+        drawMarkers()
+    }
+
+    func drawMarkers() {
+        let path = CGMutablePath()
+        let totalLines = CGFloat(parent.lines.height)
+        let markerBarHeight = self.layer?.bounds.height
+        let maxScrollerWidth = self.layer?.bounds.width
```
cmyr: I see a few different approaches with these optional values.

The first question to ask with an `Option` type (here or in rust) is, "can this value ever be (`None`/`nil`) in normal use?" If the answer is 'definitely not', then the best thing to do is to unwrap immediately, and then the rest of the function doesn't have to deal with them.

If the answer is "maybe", or "I don't know", or "yes", then the second question is, "Is this function still meaningful if this value is (`None`/`nil`)?" if yes, then you want to provide a default value in the `nil` case, and if not you want to exit early, maybe logging that something unexpected happened.

More concretely, with this function it is _unlikely_ that the layer doesn't exist when this gets called, but I'm not sure it's impossible. It's worth at least being defensive. As this function _is not_  meaningful in that case, we should exit early. In swift, the pattern for this is a guard statement:

```swift
    func drawMarkers() {
        guard let markerBarHeight = layer?.bounds.height,
            let maxScrollerWidth = layer?.bounds.width,
            let width = overlayScrollerLayer?.bounds.width else { return; }
    // no more unwraps required
```

Sources/XiEditor/Search.swift
```diff
@@ -421,7 +424,19 @@ extension EditViewController {
             if let resultCount = statusQuery["matches"] as? Int {
                 (query?.searchField as? FindSearchField)?.resultCount = resultCount
             }
+
+            if status.first?["lines"] != nil && !(status.first?["lines"] is NSNull) {
+                query?.lines = statusQuery["lines"] as! [Int]
```
cmyr: just an observation, related to my earlier comment about optionals: ideally by this point in the function we have validated the input and can just access members safely.

scholtzan:
I see what you mean. I am not sure if there is a good approach. For comparison, if I open a large file in Xcode and set some breakpoints (which will be visible in the scrollbar) then it can also happen that the knob overlaps with indicators that are not visible in the text area.

jansol:
I don't think that is a *huge* issue. As I see it they are mainly a rough guide to where things are happening in the file. (Also because this is in line with other editors.)

Sources/XiEditor/MarkerBar.swift
```diff
@@ -32,17 +32,17 @@ class Marker {
...
 
     override func viewWillDraw() {
         super.viewWillDraw()
 
         // last child layer will be used for drawing the overlay scroller
-        overlayScrollerLayer = self.layer?.sublayers?.last
-        overlayScrollerLayer?.addObserver(self, forKeyPath: "opacity", options: .new, context: nil)
-        self.layer?.addSublayer(markerLayer)
+        overlayScrollerLayer = self.layer!.sublayers!.last
+        overlayScrollerLayer.addObserver(self, forKeyPath: "opacity", options: .new, context: nil)
```
cmyr: do we need to set this observer each time? Does the layer change? In any case it's important when adding an observer that you also remove it before the object gets deallocated; see [the docs](https://developer.apple.com/documentation/objectivec/nsobject/1412787-addobserver#) for more detail.

it also isn't obvious that we need to hold on to a reference to this layer. Maybe I'm missing something?

Sources/XiEditor/MarkerBar.swift
```diff
@@ -32,17 +32,17 @@ class Marker {
 class MarkerBar: NSScroller {
     var markers: [Marker] = []
     var backgroundColor: NSColor = NSColor.red
-    var parent: EditViewController!
-    var markerLayer = CAShapeLayer()
-    var overlayScrollerLayer: CALayer?
+    let markerLayer = CAShapeLayer()
+    var overlayScrollerLayer: CALayer!
```
cmyr: implicitly unwrapped optionals are dangerous, and should be used with caution. Basically: you should only use an implicitly unwrapped optional for things that are guaranteed to be present during normal use,  but which can't be present when the object is init'd; a good example of this is some subview that is always created in `viewDidLoad`, and exists for the rest of the object's lifecycle, but can't exist in `init`. Just making this an optional makes the intent clearer.

Sources/XiEditor/MarkerBar.swift
```diff
@@ -69,18 +69,21 @@ class MarkerBar: NSScroller {
 
     func drawMarkers() {
         let path = CGMutablePath()
-        let totalLines = CGFloat(parent.lines.height)
-        let markerBarHeight = self.layer?.bounds.height
-        let maxScrollerWidth = self.layer?.bounds.width
+        let totalLines = CGFloat(markerDelegate!.lines.height)
```
cmyr: this looks like something else that can go in the `guard` statement.

Sources/XiEditor/MarkerBar.swift
```diff
@@ -69,18 +69,21 @@ class MarkerBar: NSScroller {
...
+
+        print(totalLines)
+
+        guard let markerBarHeight = layer?.bounds.height,
```
cmyr: It's idiomatic to put the guard statement at the very top of the function; that way when reading the function the first thing you do is scan the `guard`, which tells you about the preconditions the function expects to hold.

Sources/XiEditor/MarkerBar.swift
scholtzan: We need the layer later again to determine the width of the scrollbar (it is wider when mouse cursor is on top) when drawing the markers. It would be possible to use `self.layer!.sublayers!.last` though but I thought `overlayScrollerLayer` would be more descriptive.

Sources/XiEditor/MarkerBar.swift
```diff
@@ -0,0 +1,95 @@
...
+    var markers: [Marker] = []
+    var backgroundColor: NSColor = NSColor.red
+    let markerLayer = CAShapeLayer()
+    var overlayScrollerLayer: CALayer?
+    weak var markerDelegate: EditViewController?
+
+    override func viewWillDraw() {
+        super.viewWillDraw()
+
+        if (overlayScrollerLayer == nil) {
```
scholtzan: This should make sure that we only add one observer, but I am not certain if this is the best way of doing this.

cmyr: Why viewWillDraw? It seems like we just want to add the observer once, when the view is created; viewWillDraw is going to get called _a lot_.

scholtzan: I tried to set the observer in `init` but at that point `self.layer!.sublayers!.last` doesn't exist yet.

Sources/XiEditor/MarkerBar.swift
```diff
@@ -0,0 +1,95 @@
...
+    let line: Int
+    let color: NSColor
+
+    init(_ line: Int, color: NSColor) {
+        self.line = line
+        self.color = color
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
```
mmatoszko: Hmm do we need this init? This class does not inherit from anything, my guess would be that it's not needed.

Sources/XiEditor/MarkerBar.swift
mmatoszko: Doesn't it get called when the user resizes the view?

Sources/XiEditor/MarkerBar.swift
scholtzan: True, will remove.

scholtzan:
@mmatoszko `viewWillDraw` doesn't seem to get called when it is resized. I just added a `print` to the method to see how often it is executed and it only gets called once.

Sources/XiEditor/MarkerBar.swift
scholtzan: @mmatoszko `viewWillDraw` doesn't seem to get called when it is resized. I just added a `print` to the method to see how often it is executed and it only gets called once.

Sources/XiEditor/MarkerBar.swift
```diff
@@ -0,0 +1,90 @@
...
+        self.color = color
+    }
+}
+
+class MarkerBar: NSScroller {
+    var markers: [Marker] = []
+    var backgroundColor: NSColor = NSColor.red
+    let markerLayer = CAShapeLayer()
+    var overlayScrollerLayer: CALayer?
+    weak var markerDelegate: EditViewController?
```
cmyr: this should be a `MarkerDelegate` type?

scholtzan: yup

PR #325, opened 2018-11-01
Remove sync callback from Dispatcher

mmatoszko:
Output in the `Dispatcher.swift` is not used anymore. Let's remove it in order to clear the road a little for the future refactoring.

This might slightly influence #275. I think that before introducing bigger changes of `CoreConnection.swift` we should first have some tests for it.

### DIFF
--- a/XiEditor/Dispatcher.swift
+++ b/XiEditor/Dispatcher.swift
@@ -21,19 +21,19 @@ class Dispatcher {
         self.coreConnection = coreConnection
     }
 
-    func dispatchSync<E: Event, O>(_ event: E) -> O {
+    func dispatchSync<E: Event>(_ event: E) {
         let rpc = event.rpcRepresentation
-        return coreConnection.sendRpc(rpc.method, params: rpc.params) as! O
+        let _ = coreConnection.sendRpc(rpc.method, params: rpc.params)
     }
 
-    func dispatchAsync<E: Event, O>(_ event: E) -> O {
+    func dispatchAsync<E: Event>(_ event: E) {
         let rpc = event.rpcRepresentation
-        return coreConnection.sendRpcAsync(rpc.method, params: rpc.params) as! O
+        coreConnection.sendRpcAsync(rpc.method, params: rpc.params)
     }
 
     func dispatchWithCallback<E: Event>(_ event: E, callback: @escaping (RpcResult) -> ()) {
         let rpc = event.rpcRepresentation
-        return coreConnection.sendRpcAsync(rpc.method, params: rpc.params) { result in
+        coreConnection.sendRpcAsync(rpc.method, params: rpc.params) { result in
             callback(result)
         }
     }
--- a/XiEditor/Dispatcher.swift
+++ b/XiEditor/Dispatcher.swift
@@ -47,7 +47,7 @@ enum EventDispatchMethod {
 }
 
 protocol Event {
-    //NOTE: output is now unused; this file in general should be considered deprecated.
+    //NOTE: this file in general should be considered deprecated.
     // In the future we would like to move to having a 'XiCore protocol', and then implementing that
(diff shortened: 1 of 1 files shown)


raphlinus:
I'm not sure I understand this. If we do a sync RPC through dispatcher, how do we get the return value? If we're not doing any sync RPC's through dispatcher (I haven't checked), we should think about what to do. I think the plan for vi key bindings involves a synchronous RPC, so that it's possible to tell in the `EditViewController.keyDown` method whether the key is to be interpreted as an editing command or directed to the IME. The main issue for that seems to be https://github.com/xi-editor/xi-editor/issues/93, but I'm not sure exactly where the discussion of sync RPC lives.

mmatoszko:
I assumed that since `Dispatcher.swift` is not consistently used across the project it should be removed in the future. This PR was my attempt to gradually remove it.

Maybe it will make more sense to first go with the interface for `CoreConnection` and then decide which abstractions would compose nicely into the architecture.

Thanks a lot for the background regarding why the results of the sync calls are needed. I didn't really knew that 😄

mmatoszko:
I removed the sync dispatch method from the Dispatcher.

`EventDispatchMethod.sync` enum type is still there, as it is used for `dispatchWithCallback` [here](https://github.com/xi-editor/xi-mac/blob/master/XiEditor/XiDocumentController.swift#L126) and [here](https://github.com/xi-editor/xi-mac/blob/master/XiEditor/XiDocumentController.swift#L178).

cmyr:
@mmatoszko so in general I appreciate the piecemeal approach, but I _wonder_ if in this case it might make sense to rip off the bandaid and just delete Dispatcher.swift, and then chase down and correct all the compiler warnings?

Alternatively, we could start actually adding a replacement mechanism, migrate towards that, and then remove Dispatcher? Or barring that, I'd be curious to hear if you have a sense of what your next steps would be after this PR, to carry on this bit of work?

mmatoszko:
Ripping off the bandaid is a funny approach, but this would create a big PR which would take a long time to be reviewed 😅

My current understanding is that removing `Dispatcher.swift` would also mean removing [Events enum](https://github.com/xi-editor/xi-mac/blob/master/XiEditor/Dispatcher.swift#L89) which is widely used in the project. I think that it is super important to agree on an appropriate abstraction first.
When the abstraction is established we could replace the `Events enum`. Then we can start thinking about removing `Dispatcher`.

Opinions? I don't feel super comfortable with the codebase yet, so it's very likely that I overlooked something important 😄

cmyr:
Yea, this would mean getting rid of `Events`.

My thinking is that we just want to have some `XiCore` protocol that exposes all the various RPC methods as... methods. It might make sense to also have a `CoreDocument` protocol that represents the various edit methods?

mmatoszko:
Nice and simple, I like it!

If it is fine for you I'll try approaching this in a separate PR's 👍

cmyr:
Okay, so looking at this I _think_ the best plan of attack would be to add the replacement interface, and _then_ remove this?

In this world, it might also be worth declaring a separate interface for `CoreConnection`; at some point in the future we will want to explore using something other than stdio to provide the connection to core.

mmatoszko:
Yes, exactly :)

When it comes to the `CoreConnection`, We can go with adding simple protocol:
```swift
protocol Connection {
    func sendRpcAsync(_ method: String, params: Any, callback: RpcCallback?)
}
```

Having that, we can just make the `CoreConnection` conform to that protocol, and add more changes incrementally.

What should we now do with this PR?
I'm not sure this is a good place for discussion - maybe we should move it into an issue or something 😄

mmatoszko:
Works for me 👌
I think I have a brief idea how to proceed with `XiCore` protocol. I will prepare a little POC and let you know on how I see it ☺️

Issue #339, opened 2018-11-09
error: value of type '[NSDocument]' has no member 'compactMap'


SCdF:
Apologies if this is my fault, but it doesn't look like master compiles right now and I think I've done everything right

```
$ git clone --recurse-submodules https://github.com/xi-editor/xi-mac

...

$ xcodebuild

...

/Users/scdf/Code/Other/xi-mac/Sources/XiEditor/AppDelegate.swift:456:37: error: value of type '[NSDocument]' has no member 'compactMap'
        return NSApplication.shared.orderedDocuments
               ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~

** BUILD FAILED **


The following build commands failed:
        CompileSwift normal x86_64
        CompileSwiftSources normal x86_64 com.apple.xcode.tools.swift.compiler
(2 failures)
(no-nvm) scdf at SCdF in ~/Code/Other/xi-mac on master
$ rustup update stable
info: syncing channel updates for 'stable-x86_64-apple-darwin'

  stable-x86_64-apple-darwin unchanged - rustc 1.30.1 (1433507eb 2018-11-07)

(no-nvm) scdf at SCdF in ~/Code/Other/xi-mac on master
$ xcodebuild -version
Xcode 9.2
Build version 9C40b
```

cmyr:
@SCdF I suspect this is a swift version difference? Can you try updating your Xcode to at least 9.4?

SCdF:
Ahh, damn. I'm guessing if there is no compatible upgrade on the App Store I'm on the latest right?

I'm still on Sierra, and don't think I can upgrade any time soon. (high sierra sounded really unstable when it came out and I can't find an iso for it anymore, and Mojave drops acceptable font rendering for external non retina monitors, and I tried it for like 20 minutes before my eyes started bleeding).

cmyr:
As a hack you could try just patching locally, replacing the line in question with `.map { $0 as! Document }` or something? Our support target is 'the last two macOS releases' but I don't see any reason to break over something this minor so if that fixes your situation we'd definitely accept a PR.

SCdF:
@cmyr changing that caused the compilation to succeed. Thanks for the help! I created a PR linked above

mmatoszko:
Can we close it?

SCdF:
Yep, the linked PR was merged and solved the compilation problems in XCode 9.2.

Issue #260, opened 2018-09-17
set up travis / CI for this repo


cmyr:
We don't have much in the way of testing on the mac side, but it would be nice if we had a simple travis script set up that would at least compile the editor and run what we have.

AlanFoster:
I'll see what I can do 🤞

cmyr:
@AlanFoster It's a bit more complicated then that.

The problem is that there's currently no way for us to compile xi-mac without compiling xi-core, because xi-core's compilation is handled by a build script in xi-mac.

I think a possible solution would be to make xi-core its own target, and then have a scheme that we use for CI that only builds xi-mac? But I'm not sure if this is the best bet, it would take some exploring.

cmyr:
yep, thanks :)

PR #340, opened 2018-11-09
Use map instead of compactMap for XCode 9.2 compat

SCdF:
See #339

### DIFF
--- a/Sources/XiEditor/AppDelegate.swift
+++ b/Sources/XiEditor/AppDelegate.swift
@@ -454,7 +454,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     // Provice a convenience helper for ordered documents
     private var orderedDocuments: [Document] {
         return NSApplication.shared.orderedDocuments
-            .compactMap { $0 as? Document }
+            .map { $0 as! Document }
     }
 
     func handleFontChange(fontName: String?, fontSize: CGFloat?) {
--- a/Sources/XiEditor/AppDelegate.swift
+++ b/Sources/XiEditor/AppDelegate.swift
@@ -451,8 +451,9 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
         }
     }
 
-    // Provice a convenience helper for ordered documents
+    // Provide a convenience helper for ordered documents
     private var orderedDocuments: [Document] {
+        // Force unwrapping is intentional: if this cast is wrong we want to crash!
         return NSApplication.shared.orderedDocuments
             .map { $0 as! Document }
     }


Sources/XiEditor/AppDelegate.swift
```diff
@@ -454,7 +454,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     // Provice a convenience helper for ordered documents
     private var orderedDocuments: [Document] {
         return NSApplication.shared.orderedDocuments
-            .compactMap { $0 as? Document }
+            .map { $0 as! Document }
```
raphlinus: I think this is ok, it seems to me the documents should always be `Document`. Since @mmatoszko wrote this code, maybe he has an opinion?

cmyr: My feeling is that if this _isn't_ a `Document`, crashing is good.

mmatoszko: I'm not sure if document can crash here. Maybe going with `flatMap` would be a better option?

cmyr: @mmatoszko The only time this should crash is if we're somehow instantiating an `NSDocument` subclass that is not our own `Document`. To me that indicates a serious problem, and crashing would be good. Do you think there are any normal circumstances where this is possible?

mmatoszko: Ok, I get your point now, thanks 👍 
@SCdF could you drop a comment line, that the force unwrapping is intentional here?
We should reduce the amount of the force unwrappings in the future, and it would be nice to have an explicit information that we should omit this one.

raphlinus: Also tiny nit, s/Provice/Provide/

Happy to approve after those minor changes.

cmyr: @mmatoszko I do also agree that we unwrap to much in general fwiw. 😄

mmatoszko: @liamrosenfeld great idea! 👍 
Going with `// swiftlint:disable force_unwrapping` would be super nice for this particular case..
I'm just not sure if this won't be a lot of work in the beginning 😄
I think we should first go with #260 in order to have a working CI and automate as much work as possible ^^

Opinions?

cmyr: Also: my opinion on force unwrapping is that it is not bad _per se_, but that it should be used judiciously to enforce invariants. So if an RPC has a defined schema, force unwrapping when extracting items from the RPC's params is fine; if it crashes it means either core is sending bad data or the client is misunderstanding the schema, and crashing is good in both those cases.

I think there's a common problem with things like,

```swift
if let thing = optionalThingThatShouldNeverBeNil {
    thing.doStuff()
}
// ignore the nil case
```

In that the nil case isn't handled at all, it just silently falls through, which can lead to some pretty hard to debug behaviour.

SCdF: OK I've added the comment and fixed that existing spelling mistake as noted.

I'm going to leave the linting to someone else, because I don't actually _know_ Swift (or, while we're talking about it, Rust really).

rakaramos:
sure thing!

Issue #306, opened 2018-10-16
Add a context menu


jansol:
This would provide an alternative way to access things like
* Toggle comments
* Basically everything from Edit->Transform
* Find (seeded with selection or clicked word?) and Replace
* Cut, Copy & Paste

A context (right click) menu feels like a particularly natural way to issue commands that operate on selections, particularly when using the mouse to select text. This should also show the same key equivalents as the corresponding Edit menu entry (where applicable).

AlanFoster:
I'll try taking a look at this :+1: I'm not an ios/swift developer, so this may take a bit of time for me

cmyr:
@AlanFoster cool, feel free to ask Qs here if you get stuck.

AlanFoster:
@MartinJrP Go ahead :+1:

cmyr:
@MartinJrP welcome, look forward to a PR! If you have any questions about process or anything else feel free to drop by http://xi.zulipchat.com. ✌️

PR #343, opened 2018-11-11
Add context menu with cut, copy and paste

MartinJrP:
Addresses issue #306

Implemented by overriding EditView.defaultMenu. This allows for the system to handle right click events regardless of how the user has chosen to access context menus (tap with two fingers vs bottom right corner tap).

The NSMenu is implemented in EditViewController so that it can share the existing implementation.

NSMenuItem.keyEquivalent was left blank in keeping with the HIG. This also aligns with what Xcode does.

Future work should include a method to validate the menuItems actions. For example, “Paste” should be disabled when no items exist on the pasteboard. This might be a bit more involved than I initially thought since this should apply to items in the Menu Bar as well.

I went with cut, copy and paste and sensible defaults. Find might be another great initial menu item but I did have questions about its behaviour - namely beginning a search from selected text. This might probably then be implemented from the existing .find() action. 

I'm sure there's a discussion to be had for what items will exist in the context menu as time goes on. Thanks!

### DIFF
--- a/Tests/XiEditorTests/FindStatusTests.swift
+++ b/Tests/XiEditorTests/FindStatusTests.swift
@@ -37,7 +37,7 @@ class FindStatusTests: XCTestCase {
         XCTAssertNil(status?.isRegex)
         XCTAssertNil(status?.wholeWords)
         XCTAssertEqual(64, status?.matches)
-        XCTAssertEqual([10, 79, 85, 123, 134, 165], status?.lines)
+        XCTAssertEqual([10, 79, 85, 123, 134, 165], status!.lines)
     }
 
     func testParsingFullDict() {
@@ -60,7 +60,7 @@ class FindStatusTests: XCTestCase {
         XCTAssertEqual(true, status?.isRegex)
         XCTAssertEqual(true, status?.wholeWords)
         XCTAssertEqual(64, status?.matches)
-        XCTAssertEqual([10, 79, 85, 123, 134, 165], status?.lines)
+        XCTAssertEqual([10, 79, 85, 123, 134, 165], status!.lines)
     }
 
     func testParsingDictWithoutId() {


ryanbloom:
Hi @MartinJrP, thanks for the pull request!

This looks like the right approach,  but I'm having trouble activating the menu with a control+click gesture. Not quite sure why that is, since macOS is supposed to show menus automatically.

You might find a better way to solve this, but one option is manually calling [`NSMenu.popUp(positioning:at:in:)`](https://developer.apple.com/documentation/appkit/nsmenu/1518212-popupmenupositioningitem) when the `EditView` receives a control+click.

jansol:
The NSTextInputClient thing is definitely still a mystery to most of us, which is largely why IME and press-and-hold accents don't work properly yet. Wouldn't be surprised if it had other side effects as well.

MartinJrP:
It's definitely got me beat so far. Don't have too much experience with the Text subsystem either. I'll add another commit that does add support for control+click as well as word_selection on accessing the menu

jansol:
Uhm, what's going on here?

MartinJrP:
Wasn't able to make too much headway on NSTextInputClient... A few ideas but nothing solid to report yet unfortunately. The last commit does add support for accessing the context menu through a right click as well as text selection on a secondary click. Maybe I didn't highlight it correctly but hopefully the next step is a review and some more feedback?

cmyr:
Okay, so played around with this a little bit, and it generally feels good. There are some problems though: if I have a selected region, and I ctrl+click on a word in that region, it deselects the region before showing the context menu.

I think the best way to address this, for now, is to avoid modifying the selection. When the user right/control clicks, we should just display the context menu at the click location, without any other logic. If the user chooses cut/copy/paste we send that RPC; core uses the current selection to determine how to handle the RPC.

In the future, I agree, we may want to query core for what commands should be available for a given location. For now, though, I think a simple/dumb implementation is good, and I think the simpler the better?

cmyr:
@MartinJrP how're finals going? 😉

PR #345, opened 2018-11-12
Update protocol semantic information

scholtzan:
This is part of https://github.com/xi-editor/xi-editor/pull/986

### DIFF
--- a/Sources/XiEditor/Annotation.swift
+++ b/Sources/XiEditor/Annotation.swift
@@ -83,7 +83,7 @@ struct AnnotationStore {
     }
 
     /// Returns for each line in the provided range the annotations in that line.
-    func annotationsForLines(lines: [Line<LineAssoc>?], lineRange: Range<Int>) -> [Int: [AnnotationType: Array<AnnotationSpan>.Iterator]] {
+    func annotationsForLines(lines: [Line<LineAssoc>?], lineRange: CountableRange<Int>) -> [Int: [AnnotationType: Array<AnnotationSpan>.Iterator]] {
         var annotationIx: [AnnotationType : Int] = [:]
         for (annotationType, _) in annotations {
             annotationIx[annotationType] = 0
@@ -91,7 +91,7 @@ struct AnnotationStore {
 
         var annotationsForLines: [Int: [AnnotationType: Array<AnnotationSpan>.Iterator]] = [:]
 
-        for lineIx in lineRange.makeIterator() {
+        for lineIx in lineRange {
             annotationsForLines[lineIx] = [:]
             let relLineIx = lineIx - lineRange.first!
             guard let line = lines[relLineIx] else {
--- a/Sources/XiEditor/Annotation.swift
+++ b/Sources/XiEditor/Annotation.swift
@@ -91,7 +91,7 @@ struct AnnotationStore {
 
         var annotationsForLines: [Int: [AnnotationType: Array<AnnotationSpan>.Iterator]] = [:]
 
-        for lineIx in lineRange {
+        for lineIx in lineRange.makeIterator() {
             annotationsForLines[lineIx] = [:]
             let relLineIx = lineIx - lineRange.first!
             guard let line = lines[relLineIx] else {
--- a/Sources/XiEditor/Annotation.swift
+++ b/Sources/XiEditor/Annotation.swift
@@ -1,0 +1,125 @@
+// Copyright 2018 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// Supported annotation types.
+enum AnnotationType: String, CaseIterable {
+    case Selection = "selection"
+    case Find = "find"
+}
+
+/// Represents an annotation (eg. selection, find highlight).
+struct Annotation {
+    let startLine: Int
+    let startColumn: Int
+    let endLine: Int
+    let endColumn: Int
+    let payload: AnyObject?
+    let type: AnnotationType
+
+    init?(range: [Int], data: AnyObject?, annotationType: AnnotationType) {
+        let position = range
+
+        if position.count != 4 { return nil }
+
+        type = annotationType
+        startLine = position[0]
+        startColumn = position[1]
+        endLine = position[2]
+        endColumn = position[3]
+        payload = data
+    }
+}
+
+/// Represents an annotation to be represented in a specific line.
+typealias AnnotationSpan = (startIx: Int, endIx: Int, annotation: Annotation)
+
+/// Stores all annotations that were received from core.
+struct AnnotationStore {
+    var annotations: [AnnotationType: [Annotation]]
+
+    init(from json: [[String: AnyObject]]) {
+        annotations = [:]
+
+        for annotationType in AnnotationType.allCases {
+            let annotationsOfType = json.filter({$0["type"] as! String == annotationType.rawValue})
+            annotations[annotationType] = []
+
+            if !annotationsOfType.isEmpty {
+                for annotation in annotationsOfType {
+                    let ranges = annotation["ranges"] as! [[Int]]
+                    var payloads: [AnyObject] = []
+
+                    if !(annotation["payloads"] is NSNull) && annotation["payloads"] != nil {
+                        payloads = annotation["payloads"] as! [AnyObject]
+                    }
+
+                    for (i, range) in ranges.enumerated() {
+                        let payload = payloads.count > 0 ? payloads[i] : nil
+                        annotations[annotationType]?.append(Annotation(range: range, data: payload, annotationType: annotationType)!)
+                    }
+                }
+            }
+        }
+
+        // sort annotations by lines where they end
+        annotations = annotations.mapValues {a in a.sorted(by: { $0.endLine < $1.endLine })}
+    }
+
+    /// Returns for each line in the provided range the annotations in that line.
+    func annotationsForLines(lines: [Line<LineAssoc>?], lineRange: Range<Int>) -> [Int: [AnnotationType: Array<AnnotationSpan>.Iterator]] {
+        var annotationIx: [AnnotationType : Int] = [:]
+        for (annotationType, _) in annotations {
+            annotationIx[annotationType] = 0
+        }
+
+        var annotationsForLines: [Int: [AnnotationType: Array<AnnotationSpan>.Iterator]] = [:]
+
+        for lineIx in lineRange {
+            annotationsForLines[lineIx] = [:]
+            let relLineIx = lineIx - lineRange.first!
+            guard let line = lines[relLineIx] else {
+                continue
+            }
+
+            for (annotationType, annotationsOfType) in annotations {
+                var annotationsInLine: [AnnotationSpan] = []
+                var ix = annotationIx[annotationType]!
+
+                // get all annotations that are part of this line
+                while ix < annotationsOfType.count && annotationsOfType[ix].startLine <= lineIx {
+                    if (annotationsOfType[ix].endLine >= lineIx) {
+                        let annotation = annotationsOfType[ix]
+                        let start = annotation.startLine == lineIx ? annotation.startColumn : 0
+                        let end = annotation.endLine == lineIx ? annotation.endColumn : line.text.count
+                        let startIx = utf8_offset_to_utf16(line.text, start)
+                        let endIx = utf8_offset_to_utf16(line.text, end)
+                        annotationsInLine.append(AnnotationSpan(startIx, endIx, annotation))
+
+                        if annotation.endLine == lineIx {
+                            // processing annotation is done, move on to next
+                            annotationIx[annotationType] = annotationIx[annotationType]! + 1
+                        }
+                    }
+
+                    ix += 1
+                }
+                annotationsForLines[lineIx]![annotationType] = annotationsInLine.makeIterator()
+            }
+        }
+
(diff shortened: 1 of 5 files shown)


Sources/XiEditor/EditView.swift
```diff
@@ -481,9 +483,36 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
             } else {
                 let builder = TextLineBuilder(line.text, font: font)
                 builder.setFgColor(argb: foregroundArgb)
-                styleMap.applyStyles(builder: builder, styles: line.styles, selColor: selArgb, highlightColors: highlightsArgb)
-                textLine = builder.build(fontCache: renderer.fontCache)
+                styleMap.applyStyles(builder: builder, styles: line.styles)
+
+                // draw annotations
+                // todo: make this better
```
scholtzan: This part is really ugly. I am not sure what the best way is to approach this.

scholtzan:
One thing that is very awkward in the current implementation is rendering the selection and highlight spans. I am wondering what the best way is to deal with this. Maybe it would make sense to do `def_style` for annotations?

cmyr:
So my thinking here is that we define certain 'types' of annotations, and we describe how a client might handle them, but then actually doing that is up to the client.

So the client has access to the theme information; we can just say, "for selection annotations we expect the client to adjust the background color of the annotated region by using the `selectionColor` of the current theme. The client may also chose to render selection annotations with the `inactiveSelectionColor` when the view in question does not have user focus."

in essence the rendering of selections is defined as part of an informal protocol; final behaviour is left up to the client.

Sources/XiEditor/EditView.swift
scholtzan: Mh, or maybe it's alright(ish)...

scholtzan:
Okay, that is the current behaviour. So I guess it is ready for review.

Sources/XiEditor/LineCache.swift
```diff
@@ -110,13 +140,30 @@ fileprivate class LineCacheState<T>: UnfairLock {
...
+        let annotationData = (update["annotations"] as! [[String: AnyObject]])
+
+        for annotationType in AnnotationType.all {
+            let annotationsOfType = annotationData.filter({$0["type"] as! String == annotationType.rawValue})
+
+            if annotationsOfType.isEmpty {
+                annotations[annotationType] = []
+            } else {
+                let annotationTypeData = (annotationsOfType.first!)["data"] as! [[String: AnyObject]]
+                annotations[annotationType] = annotationTypeData.map({(d: [String: AnyObject]) -> Annotation in
```
mmatoszko: There's a super clever way to simplify this one ;)
`annotations[annotationType] = annotationTypeData.map(Annotation.init)`

Sources/XiEditor/LineCache.swift
```diff
@@ -60,6 +54,41 @@ struct Line<T> {
...
+
+/// Represents an annotation (eg. selection, find highlight)
+struct Annotation {
+    var startLine: Int
+    var startColumn: Int
+    var endLine: Int
+    var endColumn: Int
+    var metadata: [String: AnyObject]?
+
+    init(fromJson json: [String: AnyObject]) {
```
mmatoszko: If a specific requirements need to be met in order to create `Annotation` we could make use of the optional initializer here 😄

```swift
    init?(fromJson json: [String: AnyObject]) {
         guard let position = json["pos"] as? [Int], position.count == 4 else { return nil }

         startLine = position[0]
         startColumn = position[1]
         endLine = position[2]
         endColumn = position[3]
         metadata = json["metadata"] as? [String: AnyObject]
     }
```

Sources/XiEditor/EditView.swift
```diff
@@ -481,9 +483,35 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
             } else {
                 let builder = TextLineBuilder(line.text, font: font)
                 builder.setFgColor(argb: foregroundArgb)
-                styleMap.applyStyles(builder: builder, styles: line.styles, selColor: selArgb, highlightColors: highlightsArgb)
-                textLine = builder.build(fontCache: renderer.fontCache)
+                styleMap.applyStyles(builder: builder, styles: line.styles)
+
+                // draw annotations
+                for annotationType in AnnotationType.all {
```
cmyr: I'm worried about performance, here; it looks like we're going to be iterating through every annotation for every line.

I'm not sure of the best approach, but my initial sense is that it makes the most sense to treat each annotation type independently; we keep an index into each set of annotations, and for each line we just check if that annotation occurs in this line, advancing the cursors as necessary.

The other thing is that not all annotations are going to be drawn in the same way at the same time. Find and selection are both background colors, but other annotations might need special handling; so I think it makes sense to treat each type uniquely (we could also get a bit fancier and group certain types together; i'm not sure if that would be a win or not).

Sources/XiEditor/EditView.swift
```diff
@@ -481,9 +483,35 @@ class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
...
+                    let annotationsForLine = annotations[annotationType]?.filter({
```
cmyr: can't this easily be nil?

Sources/XiEditor/LineCache.swift
```diff
@@ -60,6 +54,38 @@ struct Line<T> {
     }
 }
 
+enum AnnotationType: String {
+    case Selection = "selection"
+    case Find = "find"
+}
+
+extension AnnotationType {
+    static var all: [AnnotationType] {
```
cmyr: We should be able to get this from the stdlib; check out https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html#ID581

Sources/XiEditor/LineCache.swift
```diff
@@ -60,6 +54,38 @@ struct Line<T> {
...
+        return [AnnotationType.Selection, AnnotationType.Find]
+    }
+}
+
+/// Represents an annotation (eg. selection, find highlight)
+struct Annotation {
+    var startLine: Int
```
cmyr: these can all be `let` bindings, I think?

Sources/XiEditor/EditView.swift
scholtzan: Makes sense.

scholtzan: Okay, that is a bit trickier than I thought since annotations of the same type can overlap with each other. So I guess instead of just one index, we'd need to track multiple active indices.

cmyr: hm, not sure I follow. I mean: overlap is possible I guess in some future cases (lints etc) but not in selections or find highlights?

scholtzan: For find highlights it is possible (for example if we have the text "testar" and one query for _test_ and one for _star_). And might also be nice to have for future cases.

cmyr: we could then at least keep a single offset which of the place where, for each line, we should start scanning for possible annotations?

scholtzan: That is also possible, though maybe it's simpler to just have a set of active annotations that gets then updated for each line. I can just try it out.

scholtzan: Keeping track of an offset turned out to be a better solution.

scholtzan:
This should be ready for another look 👀

Sources/XiEditor/EditView.swift
```diff
@@ -483,9 +492,42 @@ final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
             } else {
                 let builder = TextLineBuilder(line.text, font: font)
                 builder.setFgColor(argb: foregroundArgb)
-                styleMap.applyStyles(builder: builder, styles: line.styles, selColor: selArgb, highlightColors: highlightsArgb)
-                textLine = builder.build(fontCache: renderer.fontCache)
+                styleMap.applyStyles(builder: builder, styles: line.styles)
+
+                for annotationType in annotationsToBeDrawn {
```
mmatoszko: This for loop seems super hard to understand. Can't we chop it into smaller methods? 😄

Sources/XiEditor/LineCache.swift
```diff
@@ -110,13 +131,38 @@ fileprivate class LineCacheState<T>: UnfairLock {
     /// Updates the state by applying a delta. The update format is detailed in the
     /// [xi-core docs](http://xi-editor.github.io/xi-editor/docs/frontend-protocol.html#view-update-protocol).
     func applyUpdate(update: [String: AnyObject]) -> InvalSet {
+        // parse annotations
+        let annotationData = (update["annotations"] as! [[String: AnyObject]])
+
+        for annotationType in AnnotationType.allCases {
```
mmatoszko: I didn't think it through, but it seems like iterating over `annotationData` could create a simpler solution here 🤔

Sources/XiEditor/LineCache.swift
```diff
@@ -60,6 +54,32 @@ struct Line<T> {
...
 }
 
+enum AnnotationType: String, CaseIterable {
+    case Selection = "selection"
+    case Find = "find"
+}
+
+/// Represents an annotation (eg. selection, find highlight)
+struct Annotation {
+    let startLine: Int
```
cmyr: we do have a `BufferPosition` type defined in `EditView.swift`, that might be useful here? Although also fine as-is.

Sources/XiEditor/LineCache.swift
cmyr: I also think we can totally move this to a helper function, `updateAnnotations`. `applyUpdate` is already over a hundred lines long, and this new code is pretty independent.

Sources/XiEditor/EditView.swift
cmyr: I agree that there's some room for improvement here. Certainly the color picking could easily be a helper function. I haven't totally wrapped my head around the actual logic yet, but something that occurs to me is that we could have some little struct like `AnnotationHelper` that has a method `annotationsForLine`, that handles a bunch of the logic and just gives us an iterator or something to work with?

Anyway it seems like the logic here is pretty straight forward, and that's really encouraging.

scholtzan: Totally agree about putting this into helper functions. 
Not entirely sure about having a `annotationsForLine` helper function. I don't think it would be very performant since the function would for each line have to go through the annotations (and worst case all of them) from the beginning to find the ones shown in the current line. Oh unless the function returns an iterator for each line.

cmyr: yea, that was my thinking; maybe you init the helper with a line range, and then it tracks the state for you and for each line returns an iterator.

Sources/XiEditor/Annotation.swift
```diff
@@ -0,0 +1,125 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// Supported annotation types.
+enum AnnotationType: String, CaseIterable {
```
scholtzan: Travis CI doesn't seem to like this :(

mmatoszko: `CaseIterable` is Xcode 10 feature. We're currently at 9.2 👼

cmyr: Okay, if there's a compelling reason to update our swift version that's okay with me.

scholtzan: I could also do some workaround instead of using `CaseIterable` if it's too much of a hassle.

Sources/XiEditor/Annotation.swift
```diff
@@ -0,0 +1,125 @@
...
+    case Selection = "selection"
+    case Find = "find"
```
rakaramos: suggestion:

if you do `case selection`, it will be same as `case selection = "selection"`.

(and camelCase is more swifty)

what do you think?

Sources/XiEditor/Annotation.swift
```diff
@@ -0,0 +1,129 @@
...
+        endColumn = position[3]
+        payload = data
+    }
+}
+
+/// Represents an annotation to be represented in a specific line.
+typealias AnnotationSpan = (startIx: Int, endIx: Int, annotation: Annotation)
+
+/// Stores all annotations that were received from core.
+struct AnnotationStore {
```
cmyr: Yea seeing this here I think it makes a lot of sense.

Sources/XiEditor/EditView.swift
```diff
@@ -414,6 +414,24 @@ final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         }
     }
 
+    func annotationColor(for annotation: Annotation) -> UInt32 {
+        let selectionColor = self.isFrontmostView ? dataSource.theme.selection : dataSource.theme.inactiveSelection ?? dataSource.theme.selection
+        let selArgb = colorToArgb(selectionColor)
+        let highlightColors = dataSource.theme.findHighlights
+        let highlightsArgb = (highlightColors ?? [])!.map({
```
cmyr: this looks weird, trying to get my head around it. If `highlightColors` can be `nil`, I think it makes sense to do the `?? []` when we declare it above here? But this is confusing, because if it can be nil and this can be empty then below we will be indexing into an empty array, which should crash?

scholtzan: Right, I think it makes sense to use the selection color as default then if `highlightColors` is `nil`.

cmyr: this also might explain a bug I saw testing here where with the github theme the selection color was bright red?

scholtzan: The selection color? Currently, the find highlight color for the first query is red but that shouldn't influence the selection color.

cmyr:
@scholtzan the CI failure needs to be resolved, but then I think we're good?

scholtzan:
Done.

scholtzan:
I felt that it was easy to use. The only challenge (and maybe not so pretty part still?) is that since annotations can span multiple lines and overlap with eachother, for each line the visible annotations need to be determined (so what `annotationsForLine` does). But I am not sure if it would make sense to tweak the protocol in some way to make this easier. I still think it is quite good as it currently is.

PR #346, opened 2018-11-12
Fixed status bar flash in new windows

cobbal:
Previously, the status bar would be visible for a second on new window creation, even though it was empty.


cmyr:
@cobbal Cool, thanks for this. If you could rebase on master and then force push CI should be back working, and someone will get you a review shortly. :)

cmyr:
@nangtrongvuon do you still have a plugin somewhere that uses the statusbar? Want to double check that this still works as expected?

nangtrongvuon:
@cmyr I'll get to it soon!

Issue #320, opened 2018-10-29
A crash at EditView _blinkInsertionPoint


lht:
This happened when I was quickly closing tabs, but I cannot reproduce it.  The stacktrace is different with #261.

```
Process:               XiEditor [9636]
Path:                  /Users/USER/*/XiEditor.app/Contents/MacOS/XiEditor
Identifier:            com.levien.XiEditor
Version:               1.0 (1)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           XiEditor [9636]
User ID:               502

Date/Time:             2018-10-30 08:07:34.320 +1100
OS Version:            Mac OS X 10.13.6 (17G65)
Report Version:        12
Bridge OS Version:     3.0 (14Y664)
Anonymous UUID:        51ACA5E2-68C1-3D80-AAD6-08D809F517FF

Sleep/Wake UUID:       CD42EF1A-C9FA-425F-B8D6-C7109EFFA557

Time Awake Since Boot: 380000 seconds
Time Since Wake:       2000 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [0]

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.levien.XiEditor           	0x000000010bd48879 $S8XiEditor8EditViewC20_blinkInsertionPointyyF + 201 (EditView.swift:359)
1   com.levien.XiEditor           	0x000000010bd48897 $S8XiEditor8EditViewC20_blinkInsertionPointyyFTo + 23
2   com.apple.Foundation          	0x00007fff514b3739 __NSFireTimer + 83
3   com.apple.CoreFoundation      	0x00007fff4f349014 __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__ + 20
4   com.apple.CoreFoundation      	0x00007fff4f348c87 __CFRunLoopDoTimer + 1095
5   com.apple.CoreFoundation      	0x00007fff4f34878a __CFRunLoopDoTimers + 346
[...]

Issue #315, opened 2018-10-26
Benchmark Mojave's text painting performance


raphlinus:
OS 10.14 Mojave is making substantial changes to text painting. In particular, RGB subpixel painting is now [off by default](https://www.reddit.com/r/apple/comments/8wpk18/macos_mojave_nukes_subpixel_antialiasing_making/), though it is possible to turn it back on.

I haven't experimented with 10.14 yet, but have dug pretty deep into 10.12 and 10.13 (no significant differences between the two). Both rely on software rendering, and, though it uses SIMD, it comes nowhere near what a GPU can do. Part of the reason the code is slow is that it does gamma-corrected blending (see http://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/ though I don't endorse everything said there). Note that there are multiple code paths depending on whether the bg and fg colors are pure black/white (as is often the case in macOS) or arbitrary colors (common with styles/themes), and this should be measured in a benchmark.

Without RGB subpixel rendering, it's easier to use a GPU; painting text should just be an A8 texture draw. Gamma questions remain, but GPU's are certainly able to do gamma corrected blending, either in a shader or by setting the colorspace of the buffer.

If Mojave has fast text painting, then we probably want to revert our OpenGL renderer. There are a fairly large number of papercut issues, including #266, imperfect bidi (see #263), IME positioning, and I'm sure a few more. In addition, OpenGL is deprecated, so it's not really a good long term solution.

In addition to just painting, it's worth digging into whether the macOS performance bug underlying #20 has been fixed. Again, if not, we probably want to start doing our own logic.

If Mojave has slow text painting, then I think a long term solution is to write a text surface in Rust, backed by a suitable portable GPU-accelerated API (WebRender would work well, and porting the logic in the existing OpenGL renderer on top of gfx-rs would also make lots of sense). Such a surface could be used by multiple front-ends. In addition, moving the line cache logic into Rust as well could avoid much of the difficulty with slow JSON in Swift (see #102). But this is a lot of work and very much a long term prospect.

Having benchmarks would be super useful. Marking as "help wanted" because it doesn't require deep knowledge of the xi-mac codebase, can be done as a standalone project; verifying #20 is pretty much a case of running the code in the linked radar.

jansol:
I have no idea how to even start benchmarking the actual text painting though.

jansol:
[...]

PR #352, opened 2018-11-17
Fix a crash at EditView _blinkInsertionPoint

rakaramos:
Addresses: #320 

As per @cmyr's [comment](https://github.com/xi-editor/xi-mac/issues/320?fbclid=IwAR3mv8CzVDgOB6Wsm4F__TBoNL0TpHKun64dm7ykAzDOevd-btJXBOqvko4#issuecomment-434113029), this should fix the issue.

### Pitch

Something caught my attention on this class, the `cursor blink` state is managed via `NSTimer` class (doing UI work), every 1 second.

IMHO, that should be done via [CVDisplayLink](https://developer.apple.com/documentation/corevideo/cvdisplaylink-k0k) because it's machinery allows us to run code on every frame refresh.

I would love to create a PR with these changes, but first I would like to run it by you guys.
Please note that I'm not an macOS expert, I do iOS for a living and when we have a situation like that, we avoid `NSTimer` in favour of `CADisplayLink`. Just porting some knowledge that _maybe_ makes sense on the macOS side.

### DIFF
--- a/Sources/XiEditor/EditView.swift
+++ b/Sources/XiEditor/EditView.swift
@@ -360,6 +360,8 @@ final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
 
     /// timer callback to toggle the blink state
     @objc func _blinkInsertionPoint() {
+        guard dataSource != nil else { return }
+        
         _cursorStateOn = !_cursorStateOn
         partialInvalidate(invalid: dataSource.lines.cursorInval)
     }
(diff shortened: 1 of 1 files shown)


jansol:
My (limited) knowledge about macOS APIs agrees with using CADisplayLink. Maybe this should even have its own layer if it doesn't already.

rakaramos:
@cmyr fair enough, that's correct! Also #315 is the right place for it, and with opengl dropped in favour of metal, complicates it.
thanks!

trishume:
@cmyr turned out to only be on multi-display when you do it a certain way that it fakes it with a timer. Most of the time it works. See the update to http://thume.ca/2017/12/09/cvdisplaylink-doesnt-link-to-your-display/

I still think an NSTimer is probably what you want here since the frequency is low. If you were doing a Sublime-style smooth fade you'd want display link.

PR #353, opened 2018-11-17
Improve core log handling

### LINKED ISSUE #298: pass logging dir to core at launch
as of https://github.com/xi-editor/xi-editor/pull/888, we can specify the directory that core uses for logging. This should bet set at startup; we already find the system logging location in `AppDelegate.swift`, with the `errorLogDirectory` var.

cmyr:
This adds a few things:

- it passes our logging directory to core as a launch argument (closes #298)
- it keeps the last 100 lines of core's stderr around, and dumps this to a file if core exits unexpectedly

It still doesn't feel like we handle the core connection very elegantly. It would be nice, for instance, if we showed an alert if core had crashed, maybe with an option to open the log directory. It would also be nice if we were better about how we managed the subprocess's shutdown, in general? We crash pretty hard currently.

### DIFF
--- a/Sources/XiEditor/LoggingUtilities.swift
+++ b/Sources/XiEditor/LoggingUtilities.swift
@@ -16,10 +16,10 @@ import Foundation
 
 
 /// Error tolerant wrapper for append-writing to a file.
-struct FileWriter {
+final class FileWriter {
     let path: URL
     let handle: FileHandle
-    
+
     init?(path: String) {
         let path = NSString(string: path).expandingTildeInPath
         if FileManager.default.fileExists(atPath: path) {
@@ -49,9 +49,10 @@ struct CircleBuffer<T> {
     public let capacity: Int
     /// The index of the most recently added item
     private var head: Int = 0
-    var storage = [T]()
-    
+    private var storage = [T]()
+
     init(capacity: Int) {
+        assert(capacity > 0, "CircleBuffer capacity should be >= 1")
         self.capacity = capacity
     }
 
--- a/Sources/XiEditor/LoggingUtilities.swift
+++ b/Sources/XiEditor/LoggingUtilities.swift
@@ -1,0 +1,79 @@
+// Copyright 2018 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+
+/// Error tolerant wrapper for append-writing to a file.
+struct FileWriter {
+    let path: URL
+    let handle: FileHandle
+    
+    init?(path: String) {
+        let path = NSString(string: path).expandingTildeInPath
+        if FileManager.default.fileExists(atPath: path) {
+            print("file exists at \(path), will not overwrite")
+            return nil
+        }
+        self.path = URL(fileURLWithPath: path)
+        FileManager.default.createFile(atPath: self.path.path, contents: nil, attributes: nil)
+        
+        do {
+            try self.handle = FileHandle(forWritingTo: self.path)
+        } catch let err as NSError {
+            print("error opening log file \(err)")
+            return nil
+        }
+    }
+    
+    func write(bytes: Data) {
+        handle.write(bytes)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


cmyr:
@mmatoszko @rakaramos @liamrosenfeld maybe one of you would like to give this a look?

Sources/XiEditor/LoggingUtilities.swift
```diff
@@ -0,0 +1,79 @@
...
+    func write(bytes: Data) {
+        handle.write(bytes)
+    }
+}
+
+/// Fixed-size storage, discarding oldest items
+struct CircleBuffer<T> {
+    /// The maximum possible size for this buffer. The number of items
+    /// in the buffer may be less than this; space is allocated lazily.
+    public let capacity: Int
```
mmatoszko: Is there a chance this could go below 0?
If no, we could go with `UInt` 👍

Sources/XiEditor/LoggingUtilities.swift
```diff
@@ -0,0 +1,79 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+
+/// Error tolerant wrapper for append-writing to a file.
+struct FileWriter {
```
mmatoszko: `FileWriter` consists of `path: URL` which is a value type, and `handle: FileHandle` which is a reference type.
Is there a real reason why it is a `struct`?

Sources/XiEditor/LoggingUtilities.swift
```diff
@@ -0,0 +1,79 @@
...
+
+/// Fixed-size storage, discarding oldest items
+struct CircleBuffer<T> {
+    /// The maximum possible size for this buffer. The number of items
+    /// in the buffer may be less than this; space is allocated lazily.
+    public let capacity: Int
+    /// The index of the most recently added item
+    private var head: Int = 0
+    var storage = [T]()
+
```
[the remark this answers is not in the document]
cmyr: oops yea good catch; this was public because I was accessing it in a test in an earlier revision. Just out of curiosity, is there an idiomatic way to expose a private variable to a test runner?

Sources/XiEditor/LoggingUtilities.swift
cmyr: Good Q, I tend to use `Int` a lot in Swift (and very rarely in Rust) because (at least back when   I was writing a lot of Swift) it's the type used for indexing into most collections, and just idiomatically seems to be the most commonly used integer type. Will fix.

Sources/XiEditor/LoggingUtilities.swift
cmyr: No good reason, no. My default habit is to use a struct anywhere I won't ever want to pass a reference around. Is having a reference type in a struct not idiomatic?

Sources/XiEditor/LoggingUtilities.swift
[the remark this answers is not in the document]
rakaramos: by default props are `internal`.
in order to access internal props/methods on test, all you need to do it use `@testable` keyword.

```swift
@testable import MODULE_NAME
```

Sources/XiEditor/LoggingUtilities.swift
rakaramos: I think the main point is that `structs` have value semantics, which means copy-on-write. Copying a struct with underlaying reference type means that we have to handle the copy of the underlying reference type.

```swift
struct S {
    let ref: FileWrapper
    
    init() {
        ref = FileWrapper()
    }
}

let s = S()
print(s.ref) //"<NSFileWrapper: 0x6000011fd700>\n"

let copy = s

print(copy.ref) //"<NSFileWrapper: 0x6000011fd700>\n"
```

Notice that it's are sharing the reference, which breaks the expected behaviour of a `struct`.

Sources/XiEditor/LoggingUtilities.swift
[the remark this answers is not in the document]
cmyr: Is there a trick for `private` members? I presume not?

Sources/XiEditor/LoggingUtilities.swift
cmyr: Okay, thanks!

Sources/XiEditor/LoggingUtilities.swift
cmyr: Yea so I tried this as a `UInt` but then I couldn't index into the storage array. Added an assert.

Sources/XiEditor/LoggingUtilities.swift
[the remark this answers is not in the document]
mmatoszko: Hmm but why making all the variables private? Let's make private only those variables, which are not a part of the public interface :smile:

PR #370, opened 2018-12-07
Extract XiClient implementation from AppDelegate

mmatoszko:
## Summary
In the current implementation `AppDelegate` implements `XiClient`. This approach makes `AppDelegate` a huge class, which has many responsibilities. This PR is an initial approach to make the `AppDelegate` smaller and avoid making it a bag for storing things which are used all over the project 😄

I'm sure there are many more things which we could extract from `AppDelegate`. I think creating `ClientImplementation` is a quite nice first step to achieve that.

### DIFF
--- a/Sources/XiEditor/AppDelegate.swift
+++ b/Sources/XiEditor/AppDelegate.swift
@@ -77,17 +77,11 @@ class ScrollTester {
 }
 
 @NSApplicationMain
-class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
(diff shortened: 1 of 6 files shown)


cmyr:
Cool, this looks like a nice improvement, will take a proper look soon. 👍

mmatoszko:
This might look like a big change in the beginning, but the most of the stuff that I did there is just moving the protocol implementation to a separate class, and work hard to make it compile again :smile: 

I introduced a couple of protocols along the way in order to leave a room for decoupling things even more in the near future :sweat_smile:

PR #373, opened 2018-12-12
Added Toggle Tail config in Debug.

sjoshid:
PR for https://github.com/xi-editor/xi-editor/issues/922

### DIFF
--- a/Sources/XiEditor/ClientImplementation.swift
+++ b/Sources/XiEditor/ClientImplementation.swift
@@ -150,7 +150,7 @@ class ClientImplementation: XiClient, DocumentsProviding, ConfigCacheProviding, 
         let document = documentForViewIdentifier(viewIdentifier: viewIdentifier)
         DispatchQueue.main.async {
             let newStatusItem = StatusItem(source, key, value, alignment)
-            document?.editViewController?.statusBar.addStatusItem(newStatusItem)
+            document?.editViewController?.addStatusItem(newStatusItem)
         }
     }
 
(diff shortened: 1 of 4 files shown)


Sources/XiEditor/XiCore.swift
```diff
@@ -109,6 +111,10 @@ final class CoreConnection: XiCore {
         let params = ["destination": destination, "frontend_samples": frontendSamples] as AnyObject
         sendRpcAsync("save_trace", params: params)
     }
+    
+    func toggleTailConfig(enabled: Bool) {
```
mmatoszko: Could you add an unit test for that?
[`testTracingConfigEnabled`](https://github.com/xi-editor/xi-mac/blob/master/Tests/XiEditorTests/XiCoreTests.swift#L37) should be a clear example how to do it 😄

Sources/XiEditor/AppDelegate.swift
```diff
@@ -490,6 +499,10 @@ class AppDelegate: NSObject, NSApplicationDelegate, XiClient {
     func updateRpcTracingConfig(_ enabled: Bool) {
         xiCore?.tracingConfig(enabled: enabled)
     }
+    
+    func updateToggleTailConfig(_ enabled: Bool) {
```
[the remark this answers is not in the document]
sjoshid: Done.

Sources/XiEditor/XiCore.swift
sjoshid: Done.

jansol:
The function and RPC are called 'toggle' but what it does looks more like 'set' at a quick glance. This seems somewhat confusing to me. Maybe rename or explain the naming in a comment in the code?

cmyr:
This is tracking https://github.com/xi-editor/xi-editor/pull/1011; the final shape of this will be determined by decisions there, so holding review until then.

sjoshid:
@cmyr 
Made some progress here. I have followed existing language changed implementation as it is similar in functionality to tail feature.

cmyr:
@sjoshid This includes a bunch of files I don't think you intended?

A tip I find useful: after updating a PR (and before confirming to open a PR) click on the "files changed" tab and read through the diffs. I very often find mistakes I've made this way!

sjoshid:
Oops. Sorry about that. Not sure how they sneaked in. I think I did a bunch of useless commits but the final version looks good.

PR #374, opened 2018-12-12
Add RUST_BACKTRACE=1 to env when launching core

### LINKED ISSUE #371: explicitly set RUST_BACKTRACE=1 for xi-core process
We probably _always_ want our logs to include backtraces; it makes sense that we hard-code setting this environment variable before starting `xi-core`.

This should be straight forward. in `RPCSending.swift`, before launching the process, we'll want to set `task.environment["RUST_BACKTRACE"] = "1"`.

cmyr:
- closes #371

## Review Checklist

- [x] I have tested the code


jansol:
Wouldn't an `if let` have an empty body and do everything in the else clause? Unless you meant using the `??` operator in which case I agree that would look nicer.

Issue #378, opened 2018-12-13
Improve json parsing for findStatus


scholtzan:
The find status parsing in `findStatus` could use some cleanup.

mmatoszko:
Seems pretty straightforward, I would love to work on that one 😃

PR #379, opened 2018-12-15
Improve json parsing for findStatus

mmatoszko:
Related to #378 

## Summary
This might be just a tip of the iceberg, but it seems like a good start.

There's one thing which is super unclear for me right now.

```swift
if firstStatus.caseSensitive != nil {
    query?.ignoreCase = !(statusQuery.caseSensitive != nil)
}
```

We first check the first item of the array if it has a value for `"case_sensitive"`, and then we get an other value under the same key from the actual index of the array.
Do we want to check the `status.first` for the presence of the value?

## Review Checklist
- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/Search.swift
+++ b/Sources/XiEditor/Search.swift
@@ -388,15 +388,6 @@ extension EditViewController {
 
     func findStatus(status: [[String: AnyObject]]) {
         var findMarker: [Marker] = []
-
-        // status has the following expected format:
-        // [{
-        //      id: Int,
-        //      chars: String,
-        //      case_sensitive: Boolean,
-        //      whole_words: Boolean,
-        //      matches: Int
-        // }]
         let statusStructs = status.flatMap(FindStatus.init)
         for statusQuery in statusStructs {
             guard let firstStatus = statusStructs.first else { continue }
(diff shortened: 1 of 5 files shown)


cmyr:
Sort of a higher level comment, but my thinking here is that we should actually have a `FindQuery` struct and a `fromJson` failable initializer, and we should parse these into a real type as they come in, doing validation at that point.

mmatoszko:
Ok, I created `FindStatus` struct and changed the `findStatus(status: [[String: AnyObject]])` accordingly. I decided to not use `FindQuery`, because it's also used for `multi_find`, and the protocol definitions differ.

Feel free to take a look 😄

The question in [Summary](https://github.com/xi-editor/xi-mac/pull/379#issue-238937061) remains the same @scholtzan what do you think?

Sources/XiEditor/Search.swift
```diff
@@ -388,59 +388,37 @@ extension EditViewController {
 
     func findStatus(status: [[String: AnyObject]]) {
         var findMarker: [Marker] = []
+        let statusStructs = status.flatMap(FindStatus.init)
```
cmyr: I think it would be good to know, somehow, if there is malformed data in here; that would indicate a bug in core.

mmatoszko: Do you mean an _integration_ test? 😄

cmyr: now that's an idea 🤔

PR #384, opened 2018-12-18
Use logical line numbers in the gutter

cmyr:
This is frontend support for https://github.com/xi-editor/xi-editor/pull/1056; I'll update the submodule when that is merged.

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/LineCache.swift
+++ b/Sources/XiEditor/LineCache.swift
@@ -22,6 +22,8 @@ struct Line<T> {
     var text: String
     var cursor: [Int]
     var styles: [StyleSpan]
+    /// This line's logical number, if it is the start of a logical line
+    var number: UInt?
     /// Associated data, to be managed by client
     var assoc: T?
 
@@ -40,6 +42,7 @@ struct Line<T> {
         // this could be a more clear exception
         text = json["text"] as! String
         cursor = json["cursor"] as? [Int] ?? []
+        number = json["ln"] as? UInt
         if let styles = json["styles"] as? [Int] {
             self.styles = StyleSpan.styles(fromRaw: styles, text: self.text)
         } else {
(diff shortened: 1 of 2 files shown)


Sources/XiEditor/LineCache.swift
```diff
@@ -170,7 +174,20 @@ fileprivate class LineCacheState<T>: UnfairLock {
                     }
                     let startIx = oldIx - nInvalidBefore
                     if op_type == "copy" {
-                        newLines.append(contentsOf: lines[startIx ..< startIx + nCopy])
+                        var lineNumber = op["ln"] as! UInt
+                        let toCopy = lines[startIx ..< startIx + nCopy]
+                        // the line number in the update is the logical line number of the first VISUAL
+                        // line to copy. If this line is not itself logical, increment lineNumber for the next line.
+                        if toCopy.first??.number == nil {
```
jansol: Am I missing something about Swift syntax or should this be `first?.number`?

jansol: Hmm, or is this actually `toCopy.first ?? (self).number`? In that case, please reformat.

jansol:
Hm, upon further inspection it seems that the current line number in the gutter is only highlighted if the cursor is on the first visual part of that logical line. Not sure how easy/hard that would be to fix.

Sources/XiEditor/LineCache.swift
cmyr: no, this is a double optional; the items in this list are themselves option types, and the `first` accessor returns an option (the list may be empty). I did a double take as well.

cmyr:
That's a good point about the highlighting the line number; let's open an issue for that when this gets in.

Sources/XiEditor/LineCache.swift
jansol: Care to add a comment :D I'm sure I'll be confused again next week if I get back to this code.

PR #385, opened 2018-12-18
Add reindent option

scholtzan:
Adds an option to reindent selected text. Depends on https://github.com/xi-editor/xi-editor/pull/1057


mmatoszko:
I think the build doesn't pass because of the default `--depth=50` when Travis CI clones the submodules. We need to either increase the depth, or update the submodule revision

jansol:
Where do we have that many submodules O.o

mmatoszko:
We don't 😄 `--depth` parameter is about the number of commits we fetch from the submodule

PR #390, opened 2018-12-19
Provide an infrastructure for search integration tests

mmatoszko:
## Summary
This is an infrastructure for performing search integration tests. A concrete set of tests and a test file will be provided as a separate PR.

## Review Checklist

- [ ] I have tested the code


cmyr:
Cool! Maybe @scholtzan can take a first pass and see if she has any thoughts for what particular code paths we might to make sure this is covering?

scholtzan:
This looks like a good start. I guess we should add more test cases for all the combinations of options for regex/lowercase/whole words/...

mmatoszko:
The API of the `SearchTester` is ready for [passing in a `FindQuery`](https://github.com/xi-editor/xi-mac/pull/390/files#diff-8691470f9d0e4faa0bd8af8dec4066edR40) :)

cmyr:
Haven't gotten a chance to give this a close look yet but should have time tomorrow.

jansol:
IMHO we should bundle required test files instead of downloading stuff at build time.

cmyr:
This is a bit of a debate. My inclination might be to have a 'xi-test' module that just includes files, and can be used as a submodule in various places?

jansol:
That would be fine. As long as it does not make outside connections at build time. (because those are pain when you want reproducible builds)

Tests/IntegrationTests/SearchTester.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    func search(filePath: String, for terms: [String]) {
+        let queries = terms.map {
+            FindQuery(id: nil, term: $0, caseSensitive: false, regex: false, wholeWords: false)
+        }
+        search(filePath: filePath, for: queries)
+    }
+
+    func search(filePath: String, for queries: [FindQuery]) {
```
cmyr: is this intended to be public?

Tests/IntegrationTests/SearchTester.swift
```diff
@@ -0,0 +1,82 @@
...
+        sendEditAsync("multi_find", viewIdentifier: viewIdentifier, params: params)
+    }
+
+    private func sendEditAsync(_ method: String, viewIdentifier: ViewIdentifier, params: Any, callback: RpcCallback? = nil) {
+        let inner = ["method": method, "params": params, "view_id": viewIdentifier] as [String : Any]
+        xiCore.sendRpcAsync("edit", params: inner, callback: callback)
+    }
+}
+
+private var errorLogDirectory: URL? = {
```
cmyr: Unless you have a particular reason to want logs, I think we can just not pass a log directory for these tests.

cmyr:
Okay, here's a new repo for test files: https://github.com/xi-editor/test-data

Tests/IntegrationTests/SearchTester.swift
mmatoszko: I needed to do a little trick in order to make it possible:
https://github.com/xi-editor/xi-mac/pull/390/commits/d6d5fd75fa5dc29dcdf1d66055438edded922f80

mmatoszko:
@cmyr should be fine now, please take a look 😄

Sources/XiEditor/RPCSending.swift
```diff
@@ -76,7 +76,9 @@ class StdoutRPCSender: RPCSending {
         let errLogPath = errorLogDirectory?.path
         let errLogArgs = errLogPath.map { ["--log-dir", $0] }
         task.launchPath = path
-        task.arguments = errLogArgs
+        if let errLogArgs = errLogArgs {
```
cmyr: This is allowed to be nil, so this check isn't strictly necessary, although it doesn't do any harm.

mmatoszko: Ok, I didn't describe it enough 😄

Without this check we get into the following during assigning `task.arguments` a `nil` value:
> caught "NSInvalidArgumentException", "must provide array of arguments"
I added this check in order to avoid tests failure. I'm pretty sure it won't do us any harm 😄

PR #392, opened 2018-12-21
Unify & simplify some view geometry calculations

cmyr:
This unifies the 'offset to line' calculations, and removes
the separate storage of a scroll offset in the EditView.

This tries to have  the y-offset to line number math in one place.
Additionally we don't track the scroll offset independently, and
assume that the view's bounds are up to date when we need them.

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/EditView.swift
+++ b/Sources/XiEditor/EditView.swift
@@ -32,13 +32,17 @@ extension NSFont {
 /// A store of properties used to determine the layout of text.
 struct TextDrawingMetrics {
     let font: NSFont
-    var attributes: [NSAttributedStringKey: AnyObject] = [:]
-    var ascent: CGFloat
-    var descent: CGFloat
-    var leading: CGFloat
-    var baseline: CGFloat
-    var linespace: CGFloat
-    var fontWidth: CGFloat
+    let attributes: [NSAttributedStringKey: AnyObject]
+    let ascent: CGFloat
+    let descent: CGFloat
+    let leading: CGFloat
+    let baseline: CGFloat
+    let linespace: CGFloat
+    let fontWidth: CGFloat
+
+    var topPadding: CGFloat {
+        return descent + leading
+    }
 
     init(font: NSFont, textColor: NSColor) {
         self.font = font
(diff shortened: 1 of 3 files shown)


Sources/XiEditor/EditViewController.swift
```diff
@@ -298,18 +303,18 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
...
         shadowView.showLeftShadow = newOrigin.x > 0
         shadowView.showRightShadow = (editViewWidth.constant - (newOrigin.x + self.view.bounds.width)) > rightTextPadding
-        // TODO: this calculation doesn't take into account toppad; do in EditView in DRY fashion
-        let first = Int(floor(newOrigin.y / textMetrics.linespace))
-        let height = Int(ceil((scrollView.contentView.bounds.size.height) / textMetrics.linespace)) + 1
-        let last = first + height
 
+        let first = editView.yOffsetToLine(newOrigin.y)
+        let maxY = newOrigin.y + scrollView.contentView.bounds.size.height
+        let last = editView.yOffsetToLine(maxY) + 1
```
mmatoszko: why are we adding 1 here?

Sources/XiEditor/EditView.swift
```diff
@@ -434,16 +425,14 @@ final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
...
-        let lastVisible = max(0, Int(ceil((dirtyRect.origin.y + dirtyRect.size.height - topPad + scrollOrigin.y) / linespace)))
+        let xOff = dataSource.gutterWidth + x0 - dataSource.scrollOrigin.x
+        let yOff = topPad - dataSource.scrollOrigin.y
 
         // Note: this locks the line cache for the duration of the render
         let lineCache = dataSource.lines.locked()
         let totalLines = lineCache.height
-        let first = min(totalLines, firstVisible)
-        let last = min(totalLines, lastVisible)
+        let first = min(yOffsetToLine(dirtyRect.origin.y + dataSource.scrollOrigin.y), totalLines)
```
mmatoszko: Don't we need to use `topPad` value for `first` and `last` calculation?

cmyr:
testing these is possible; but i'm not sure how to write those tests in such a way that you aren't just writing the same equation in both places. 😒

Sources/XiEditor/EditView.swift
cmyr: now I'm wondering if topPad should be part of the actual `yOffsetToLine` fn. I think yes.

cmyr: Okay, so after digging a bit deeper: `topPad` was handled in `yOffsetToLine`, sort of, by the use of `textMetrics.descent`. Assuming a font has no leading, descent == linespace - ascent. This was definitely confusing, though; I've added a new commit that just adds a `topPadding` calculated property to `TextDrawingMetrics`, so that this will be the same everywhere.

cmyr:
I played around to see if a simple unit test was easy to do but instantiating the edit view is tricky, so no dice.

For manual testing: select a line. Move the cursor so that the center hashmark of the selection indicator (the little dash in the center of the 'I' cursor) is just above the selection. Click, and make sure that the cursor is placed on the line above the selection. Repeat, but this time click just _below_ the selection; the cursor should be placed on the selected line.

cmyr:
thanks @mmatoszko!

PR #393, opened 2018-12-21
Add context menu on EditContainerView

### LINKED ISSUE #306: Add a context menu

MartinJrP:
## Prelude

Wasn't aware moving to a branch for this work would close the original PR. For reference, it is PR #343 

## Summary
Implement a default context menu (for any event) on EditContainerView. The menu itself is generated in EditViewController so that it shares the same actions as other menu items and can be easily re-configured.

In addition, overriding menu(for:) in EditContainerView allows macOS to determine when/where to present the menu regardless of the user’s right click preferences.

Finally, having the menu on EditContaierView ensure it captures all possible events that could trigger a context menu because it is the top most view of the editor’s hierarchy.

### Implementation Notes
Right now, EditContainerView handles presenting the menu but in the future, when implementing a "word_select" on pop up for example, we might want to present the menu ourselves in EditViewController.

Also of note, we can look at querying core for figuring out what items should be available at a point. Eg copy should be disabled if there isn't a selection. 

Core can also decide when to select text when opening the menu (eg. "word_select" when nothing is selected, no selection if there is a selection). One concern is that I think this is a Mac specific behaviour so maybe querying core for something like "selection_exists" might work better.

## Related Issues
closes #306 

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/EditViewController.swift
+++ b/Sources/XiEditor/EditViewController.swift
@@ -183,6 +183,17 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         }
     }
 
+    // TODO: There should be a mechanism to validate the availability of a menu item's action.
+    var contextMenu: NSMenu = {
+        let menu = NSMenu()
+        
+        menu.addItem(withTitle: "Cut", action: #selector(cut(_:)), keyEquivalent: "")
+        menu.addItem(withTitle: "Copy", action: #selector(copy(_:)), keyEquivalent: "")
+        menu.addItem(withTitle: "Paste", action: #selector(paste(_:)), keyEquivalent: "")
+        
+        return menu
+    }()
+    
     var unifiedTitlebar = false {
         didSet {
             // Dont check if value is same as previous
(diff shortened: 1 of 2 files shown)


Sources/XiEditor/EditViewController.swift
```diff
@@ -183,6 +183,17 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         }
     }
 
+    // TODO: There should be a mechanism to validate the avaialbilty of a menu item's action.
```
cmyr: nit: availability

MartinJrP: Got it :)

PR #394, opened 2018-12-21
Fix scroll jitter during rewrap

cmyr:
This is client support for https://github.com/xi-editor/xi-editor/pull/1066. It also includes the commit from #392; if that gets merged first I'll rebase.

## Review Checklist

- [ ] I have tested the code

### DIFF
--- a/Sources/XiEditor/EditView.swift
+++ b/Sources/XiEditor/EditView.swift
@@ -123,13 +123,6 @@ func colorToArgb(_ color: NSColor) -> UInt32 {
 }
 
 final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
-    var scrollOrigin: NSPoint {
-        didSet {
-            if scrollOrigin != oldValue {
-                needsDisplay = true
-            }
-        }
-    }
     var lastRevisionRendered = 0
     var gutterXPad: CGFloat = 8
     var gutterCache: GutterCache?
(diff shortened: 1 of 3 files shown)


jansol:
1) Open a long file
2) Scroll almost all the way down
3) Observe empty view for a few seconds
4) Click on it
5) A wild selection appears! It's super effective! Scroll state spazzed out!

cmyr:
is this with sqlite.c?

jansol:
With my IRC logs, but I guess sqlite3.c would do the same given a sufficiently narrow window. (albeit with slightly shorter delays before it's done rewrapping)

cmyr:
hmm i'm not totally sure what your'e describing, but I wonder if it's related to not having https://github.com/xi-editor/xi-editor/pull/1065?

there is a bunch of stuff in flight, should probably rationalize. that, I don't even know what combination of commits I'm running atm 😕

cmyr:
hmm okay I think I see your issue.

jansol:
I did pull xi-editor/xi-editor#1066 and rebase it on top of xi-editor/xi-editor#1065 so I don't think that is the problem.

cmyr:
So @jansol's bug has revealed a fundamental problem here, which is that the frontend's line state and core's line state can get out of sync, and this can't be reasonably addressed without either limiting asynchrony or sending revision IDs back and forth with movement commands; in any case this needs more thought.

PR #396, opened 2018-12-23
Remove optionality from sendRpc RpcResult in Document.swift

mmatoszko:
## Summary
The signature of the result does not have to be an optional

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/Document.swift
+++ b/Sources/XiEditor/Document.swift
@@ -128,7 +128,7 @@ class Document: NSDocument {
 
     /// Note: this is a blocking call, and will also fail if the tab name hasn't been set yet.
     /// We should try to migrate users to either fully async or callback based approaches.
-    func sendRpc(_ method: String, params: Any) -> RpcResult? {
+    func sendRpc(_ method: String, params: Any) -> RpcResult {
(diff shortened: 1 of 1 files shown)


Sources/XiEditor/Document.swift
```diff
@@ -128,7 +128,7 @@ class Document: NSDocument {
 
     /// Note: this is a blocking call, and will also fail if the tab name hasn't been set yet.
     /// We should try to migrate users to either fully async or callback based approaches.
-    func sendRpc(_ method: String, params: Any) -> RpcResult? {
```
cmyr: hmm, I'm not sure. This come down to the whole situation with requests vs notifications; some RPCs return results, and some don't, and this code path is theoretically shared. The calls that use this that return results are `copy` and `cut`; both of these need core to return what text should be put in the pasteboard.

Does the inner `sendRpc` method on `XiCore` return an option?

mmatoszko: screenshot 2018-12-23 at 20 53 49

`Document.swift` is the only case when the return type is optional. My understanding is that we're on the safe side as soon as the project compiles 👯

PR #397, opened 2018-12-24
[Refactor] RPC Commands to `enum`

rakaramos:
## Summary
- This is a tiny and simple PR where I just wrapped all the RPC request/notification commands into a enum, backed by a `String`.
- By using an `enum`, we avoid typos (also wave-away stringly typed interface) and leverage compiler checks (like when we add a new case, depending on the implementation, the compiler will error out, showing where we should handle this new case).
- Another change is to use `assertionFailure`, instead of `print-return`. `assertionFailure` runs only on debug mode and crashes the program (which, IMHO is the correct thing to do). So now, when something goes wrong, instead of checking Xcode's console, we'll have Xcode breaking it and giving us the full stracktrace.

## Review Checklist
- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/RPCSending.swift
+++ b/Sources/XiEditor/RPCSending.swift
@@ -287,7 +287,9 @@ class StdoutRPCSender: RPCSending {
                 return
         }
 
-        let viewIdentifier = params["view_id"] as? ViewIdentifier ?? ""
+        guard let viewIdentifier = params["view_id"] as? ViewIdentifier else {
+            fatalError("view_id missing")
+        }
         
         switch method {
         case .update:
--- a/Sources/XiEditor/RPCSending.swift
+++ b/Sources/XiEditor/RPCSending.swift
@@ -287,30 +287,28 @@ class StdoutRPCSender: RPCSending {
                 return
         }
 
-        guard let viewIdentifier = params["view_id"] as? ViewIdentifier else {
-            fatalError("view_id missing")
-        }
+        let viewIdentifier = params["view_id"] as? ViewIdentifier
         
         switch method {
         case .update:
             let update = params["update"] as! [String: AnyObject]
-            self.client?.update(viewIdentifier: viewIdentifier, update: update, rev: nil)
+            self.client?.update(viewIdentifier: viewIdentifier!, update: update, rev: nil)
 
         case .scrollTo:
             let line = params["line"] as! Int
             let col = params["col"] as! Int
-            self.client?.scroll(viewIdentifier: viewIdentifier, line: line, column: col)
+            self.client?.scroll(viewIdentifier: viewIdentifier!, line: line, column: col)
 
         case .defStyle:
             client?.defineStyle(style: params as! [String: AnyObject])
 
         case .pluginStarted:
             let plugin = params["plugin"] as! String
-            client?.pluginStarted(viewIdentifier: viewIdentifier, pluginName: plugin)
+            client?.pluginStarted(viewIdentifier: viewIdentifier!, pluginName: plugin)
 
         case .pluginStopped:
             let plugin = params["plugin"] as! String
-            client?.pluginStopped(viewIdentifier: viewIdentifier, pluginName: plugin)
+            client?.pluginStopped(viewIdentifier: viewIdentifier!, pluginName: plugin)
 
         case .availableThemes:
             let themes = params["themes"] as! [String]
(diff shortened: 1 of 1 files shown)


Sources/XiEditor/EditView.swift
```diff
@@ -360,6 +360,8 @@ final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
 
     /// timer callback to toggle the blink state
     @objc func _blinkInsertionPoint() {
+        guard dataSource != nil else { return }
```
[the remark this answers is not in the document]
cmyr: This is a tricky one. On the one hand, I think we are too cavalier about unwrapping in general. On the other hand, this should get set immediately after the view is initialized, and should never be unset, so if it's nil we probably have a bug and we might want to crash to find out?

Sources/XiEditor/RPCSending.swift
```diff
@@ -222,126 +257,126 @@ class StdoutRPCSender: RPCSending {
...
+              let params = json["params"],
+              let method = RPCNotificationMethod(rawValue: jsonMethod)
             else {
-                print("unknown json from core: \(json)")
+                assertionFailure("unknown json from core: \(json)")
                 return
         }
-        let viewIdentifier = params["view_id"] as? ViewIdentifier
 
+        let viewIdentifier = params["view_id"] as? ViewIdentifier ?? ""
```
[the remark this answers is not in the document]
cmyr: so this really comes down to the implicit contracts that exist between core and client. We're responsible for both of these components; this isn't untrusted input. If this param is missing, that represents core breaking the API contract, and if that's the case there's really not a lot of value in continuing to run. It's probably even _bad_ to continue to run, because we're hiding a fundamental problem; crashing in an informative way is going to probably help us more.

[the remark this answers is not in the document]
rakaramos: yup, makes total sense! undid it already! 👍

Sources/XiEditor/RPCSending.swift
```diff
@@ -40,6 +40,41 @@ enum RpcResult {
...
 }
 
+enum RPCRequestMethod: String {
+    case measureWidth = "measure_width"
+}
+
+enum RPCNotificationMethod: String {
+    case alert
+    
+    case update
```
cmyr: is there a reason these two don't have raw values?

Sources/XiEditor/RPCSending.swift
```diff
@@ -222,78 +257,85 @@ class StdoutRPCSender: RPCSending {
...
-                self.sendResult(id: id, result: result)
+        case .measureWidth:
+            guard let args = params as? [[String: AnyObject]],
+                  let result = client?.measureWidth(args: args) else {
+                    assertionFailure("unexpected data from core: \(params)")
+                    return
             }
-        default:
-            print("unknown request \(method)")
+
```
cmyr: hard to tell from the diff but this looks like unintended whitespace

Sources/XiEditor/RPCSending.swift
rakaramos: it will use the case label directly....it would be redundant to have like `case alert = "alert"`

Sources/XiEditor/RPCSending.swift
rakaramos: I think I've added this line, after the `guard` statement.

Sources/XiEditor/RPCSending.swift
cmyr: Ah cool I didn't know that. 👍

Sources/XiEditor/RPCSending.swift
cmyr: I was looking at the whitespace on the line above that? Obviously no big deal but I like turning on 'trim trailing whitespace' + 'including blank lines' in Xcode.

Sources/XiEditor/RPCSending.swift
```diff
@@ -222,126 +257,126 @@ class StdoutRPCSender: RPCSending {
...
-        guard let method = json["method"] as? String, let params = json["params"], let id = json["id"]
+        guard let jsonMethod = json["method"] as? String,
+              let params = json["params"],
+              let id = json["id"],
+              let method = RPCRequestMethod(rawValue: jsonMethod),
+              let args = params as? [[String: AnyObject]],
+              let result = client?.measureWidth(args: args)
             else {
-                print("unknown json from core: \(json)")
+                assertionFailure("unknown json from core: \(json)")
```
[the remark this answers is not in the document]
mmatoszko: `assertionFailure` is not supposed to fail in release builds. This looks like a sweet spot for me 😄

PR #398, opened 2018-12-24
XiDocument outline

mmatoszko:
## Summary
This is an outline for the XiDocument protocol work. I think there might be some interesting challenges from this place, writing a nice test for the `resize` might be one of them 😄

## Review Checklist

- [ ] I have tested the code

### DIFF
--- a/Sources/XiEditor/XiDocument.swift
+++ b/Sources/XiEditor/XiDocument.swift
@@ -16,6 +16,7 @@ import Foundation
 
 
 protocol XiDocument: class {
+    func resize(size: CGSize)
 }
 
 class XiDocumentConnection: XiDocument {
@@ -29,6 +30,10 @@ class XiDocumentConnection: XiDocument {
     init(asyncRpc: @escaping AsyncRpc, syncRpc: @escaping SyncRpc) {
         self.asyncRpc = asyncRpc
         self.syncRpc = syncRpc
+    }
+
+    func resize(size: CGSize) {
+        sendRpcAsync("resize", params: ["width": size.width, "height": size.height])
     }
 
     private func sendRpcAsync(_ method: String, params: Any, callback: RpcCallback? = nil) {
--- a/Sources/XiEditor/XiDocument.swift
+++ b/Sources/XiEditor/XiDocument.swift
@@ -20,17 +20,22 @@ protocol XiDocument: class {
 
 class XiDocumentConnection: XiDocument {
 
-    private let xiCore: XiCore
+    typealias AsyncRpc = (_ method: String, _ params: Any, _ callback: RpcCallback?) -> Void
+    typealias SyncRpc = (_ method: String, _ params: Any) -> RpcResult
 
-    init(xiCore: XiCore) {
-        self.xiCore = xiCore
+    private let asyncRpc: AsyncRpc
+    private let syncRpc: SyncRpc
+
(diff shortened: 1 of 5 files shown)


Sources/XiEditor/XiDocument.swift
```diff
@@ -0,0 +1,46 @@
...
+
+class XiDocumentConnection: XiDocument {
+
+    typealias AsyncRpc = (_ method: String, _ params: Any, _ callback: RpcCallback?) -> Void
+    typealias SyncRpc = (_ method: String, _ params: Any) -> RpcResult
+
+    private let asyncRpc: AsyncRpc
+    private let syncRpc: SyncRpc
+
+    init(asyncRpc: @escaping AsyncRpc, syncRpc: @escaping SyncRpc) {
```
cmyr: the idea with passing these in as closures is to make this easier to mock? Or this is just kind of the 'swifty' way? I guess historically i would expect some delegate protocol or something but maybe we're moving away from that?

Sources/XiEditor/XiDocument.swift
```diff
@@ -0,0 +1,46 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+
+protocol XiDocument: class {
```
cmyr: I wonder if this might more accurately be thought of as a `XiView`? resize, for instance, isn't really a document concern. In any case this is a subtle distinction, in our current architecture.

Sources/XiEditor/XiDocument.swift
mmatoszko: This is just the initial idea of incorporating the `XiDocumentConnection` into the existing code. With this approach we'll be able to pass in [`sendRpcAsync`](https://github.com/xi-editor/xi-mac/blob/master/Sources/XiEditor/Document.swift#L118) and [`sendRpc`](https://github.com/xi-editor/xi-mac/blob/master/Sources/XiEditor/Document.swift#L131) from existing [`Document`](https://github.com/xi-editor/xi-mac/blob/master/Sources/XiEditor/Document.swift#L23) class and have existing `viewIdentifier`, tracing and queueing existing notifications for free 👯. Once we have that we'll be able to iterate on the code and decouple it 😄

Sources/XiEditor/XiDocument.swift
mmatoszko: That's an interesting point. Do you think all of the rpc calls connected to document could be also a part of `XiDocument`?

cmyr: basically: I think that any RPC that takes as a parameter a `ViewIdentifier` should be part of this; those are the commands which are addressed to a specific document/view.

Sources/XiEditor/XiViewProxy.swift
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+import Foundation
+
+
+protocol XiViewProxy: class {
+    func resize(size: CGSize)
+}
+
+class XiViewConnection: XiViewProxy {
```
rakaramos: could you make it `final` too?

PR #404, opened 2018-12-30
Add Unsaved/Edited Status Indicator

### LINKED ISSUE #365: Indicator for unsaved changes
It would be useful to have some kind of indicator to show that files have unsaved changes. This would especially be useful when working on several files at the same time. Maybe there could be some kind of indicator in each file tab.

rlziii:
## Summary
This feature adds an unsaved/edited status indicator to appropriate tabs in the editor.  Previously, the only indication that a document was unsaved was with the macOS window close button (the solid red circle would appear with a black dot inside) and (if previously saved) the document icon would be dimmed.  However, these indicators were only true if the document in question was the current active tab; and even then they were not always apparent in their subtleness.  With this change, any tab with unsaved changes will appear with a bullet character (`•`) in front of the document name.

## Related Issues
Related to #365 

closes #365 

## Details
Previous behavior:
screen shot 2018-12-30 at 10 59 37

New behavior:
screen shot 2018-12-30 at 10 59 12

## Review Checklist
- [x] I have tested the code


rlziii:
@cmyr This change was trivial indeed.  Implemented in a new commit.

Sources/XiEditor/EditViewController.swift
```diff
@@ -370,10 +370,9 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         DispatchQueue.main.async { [weak self] in
             self?.document.updateChangeCount(hasNoUnsavedChanges ? .changeCleared : .changeDone)
             
-            // If the document's edited status has changed then display as such in the window and tab titles.
-            if let window = self?.document.windowControllers[0] as? XiWindowController,
-                window.editedStatus != !hasNoUnsavedChanges {
-                window.editedStatus = !hasNoUnsavedChanges
+            // Display the document's edited status in the window and tab titles.
```
cmyr: is this still accurate? my understanding is we don't show the status in the window title.

rlziii: The tab and window title mimic one another.  So upon switching to a new tab, the window title changes as well.  So the edited indicator will also be displayed (or not displayed) in both places.

cmyr: ah okay I understand, we always set this here and then basically filter in the `didSet`? 

My only nit then would maybe be the name `showEditedIndicator`.

Thinking about this more, can't determine whether there are unsaved changes from `document.hasUnautosavedChanges`? This should always be up to date, so we can maybe just check the documents count and this value in the `windowTitle` override?

rlziii: That last suggestion does work well!  When I was originally starting this I couldn't seem to get `document.hasUnautosavedChanges` to be the expected result; but in hindsight it was likely an error on my part while testing.  A new commit removes the `showEditedIndicator` state variable in favor of the logic being moved into the `windowTitle(forDocumentDisplayName:)` method.  I believe this is much more simplified and doesn't rely on a state variable (which I don't generally like using in these circumstances).

Sources/XiEditor/XiWindowController.swift
```diff
@@ -16,10 +16,18 @@ import AppKit.NSWindowController
 
 final class XiWindowController: NSWindowController {
     override func windowTitle(forDocumentDisplayName displayName: String) -> String {
+        var editedIndicator = ""
+        
+        // If there is more than one tab AND the document in question has unsaved changes then show an edited indicator.
+        // When there is only one tab the edited indicator is already built into the macOS close (red circle) button.
+        if NSDocumentController.shared.documents.count > 1, document?.hasUnautosavedChanges ?? false {
+            editedIndicator = "• "
```
jansol: Having the space be part of the indicator seems somewhat hacky to me. Personally I'd make the sapce part of the format string, or drop it entirely. Not going to block this PR on that though, especially considering the recurring thought of ditching the whole NSDocument infra...

cmyr:
Thanks @rlziii!

PR #405, opened 2018-12-30
Single CLI Button

liamrosenfeld:
## Summary
Unifies the two CLI Buttons into a single button that toggles between the two.

## Checklist
- [x] Toggles Name
- [x] Toggles Functionality

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/AppDelegate.swift
+++ b/Sources/XiEditor/AppDelegate.swift
@@ -186,6 +186,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         xiCore.setTheme(themeName: preferredTheme)
         Trace.shared.trace("appWillLaunch", .main, .end)
         documentController = XiDocumentController()
+        
+        // Set Cli Menu Title
+        renameCliToggle()
     }
     
     // Clean up temporary Xi stderr log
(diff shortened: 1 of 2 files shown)


Sources/XiEditor/AppDelegate.swift
```diff
@@ -227,16 +239,12 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+        informationalAlert(title: message, message: info)
+        renameCliToggle()
     }
     
-    @IBAction func removeCli(_ sender: Any) {
-        let cliPath = URL(string: "file:///usr/local/bin/xi")!
+    func removeCli() {
+        let cliPath = URL(fileURLWithPath: "/usr/local/bin/xi")
         var message = "CLI Removed!"
         var info = "The file /usr/local/bin/xi has been deleted."
```
mmatoszko: I think using a path from the `cliPath` would be a nice little touch here 😄

liamrosenfeld: Amended!

liamrosenfeld:
@mmatoszko, the install/remove enum's raw values could be set to a localization variable.

cmyr:
@mmatoszko My gut feeling is that while we're still in the experimentation phase ignoring localization is okay; ultimately a lot will be changing between now and whenever we do a public release. Once we're at a stage where we're actually publishing an app package we'll want to go through and clean up all of our strings.

Sources/XiEditor/AppDelegate.swift
```diff
@@ -250,13 +258,32 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
         alert.alertStyle = .informational
-        alert.messageText = message
-        alert.informativeText = info
+        alert.messageText = title
+        alert.informativeText = message
         alert.runModal()
     }
+    
+    enum CliButtonState: String {
+        case install = "Install CLI"
```
cmyr: I'm thinking about these names. I don't love including 'CLI' in the UI, but I'm not sure what I actually prefer.

'Install Command Line Tool'?
'Install `xi` Shell Command'?
'Install Command Line `xi`'?

(This is assuming we can use an `NSAttributedString` to use Monaco or some other system monospace font)

mmatoszko: My 👍 goes for:
> Install xi Shell Command

liamrosenfeld: Enums don't seem to play nice with `NSAttributedString` so I renamed it to `Install 'xi' Shell Command`.

cmyr: Okay, and it also probably doesn't play well with localization?

One way I would have also considered addressing this is to not use `rawValue` as the title; we could have a 'menuItemTitle: NSAttributedString { get }` calculated property on the enum, and we could identify the variant by having an `Int` raw value that we set as the `tag` of the `NSMenuItem`. But I think this is good, just mentioning how I would approach that if it were the goal.

liamrosenfeld:
Thanks!

PR #415, opened 2019-01-10
README.md: remove '-scheme "XiEditor"'

srenatus:
## Summary

As discussed in Zulip, it seems that with that parameter, XiEditor.app
not actually ends up in the described place.


cmyr:
For the record this came out of a zulip discussion; specifying the scheme causes the build directory to be somewhere in ~/Library instead of relative to the working directory.

PR #417, opened 2019-01-14
Add support for opening multiple files with the CLI

### LINKED ISSUE #381: CLI should support opening multiple files at once
When passing the cli multiple files as arguments, only one gets actually opened.

kuglee:
Fixes #381. The wait functionality works with multiple files too.

### DIFF
--- a/Sources/XiCliCore/CommandLineTool.swift
+++ b/Sources/XiCliCore/CommandLineTool.swift
@@ -116,7 +116,7 @@ public final class CommandLineTool {
     func help() {
         let message = """
                 The Xi CLI Help:
-                xi <file> [--wait | -w] [--help | -h]
+                xi <file>... [--wait | -w] [--help | -h]
 
                 file
                     the path to the file (relative or absolute)
--- a/Sources/XiCliCore/CommandLineTool.swift
+++ b/Sources/XiCliCore/CommandLineTool.swift
@@ -27,7 +27,7 @@ public final class CommandLineTool {
             return
         }
         
-        if args.fileInputs.isEmpty {
+        guard !args.fileInputs.isEmpty else {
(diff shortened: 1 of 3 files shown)


cmyr:
Cool, thanks! @liamrosenfeld you might be interested in taking a first look at this?

liamrosenfeld:
Looks great to me! When I tried testing it, --wait wasn't working at first, but it started working again after I manually deleted the symlink from bin. The only thing I noticed is that the help screen needs updating.

Sources/XiCliCore/CommandLineTool.swift
```diff
@@ -110,12 +116,12 @@ public final class CommandLineTool {
     func help() {
         let message = """
                 The Xi CLI Help:
-                xi <file> [--wait | -W] [--help | -h]
+                xi <file> [--wait | -w] [--help | -h]
```
liamrosenfeld:
```suggestion
                xi <file>... [--wait | -w] [--help | -h]
```
This just brings it in line with [the docs](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-R2-and-2012/cc771080(v=ws.11)) used for the rest of the help

Sources/XiCliCore/CommandLineTool.swift
```diff
@@ -27,19 +27,24 @@ public final class CommandLineTool {
             return
         }
         
-        guard let fileInput = args.fileInput else {
+        if args.fileInputs.isEmpty {
```
liamrosenfeld:
```suggestion
        guard !args.fileInputs.isEmpty else {
```
This is not necessary, but it would be a nice addition to convey that it should always return.

Issue #364, opened 2018-11-29
Open file when drag and dropped into editor window


scholtzan:
If a file is drag & dropped from a file manager into the editor window, then the file should be opened. It currently only works if a file is drag & dropped on the xi-editor icon.

PR #420, opened 2019-01-25
Open files dragged into window

scholtzan:
This PR adds the capability to open files that are drag and dropped into an editor window.

Closes https://github.com/xi-editor/xi-mac/issues/364

### DIFF
--- a/Sources/XiEditor/EditView.swift
+++ b/Sources/XiEditor/EditView.swift
@@ -191,6 +191,7 @@ final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         let glLayer = TextPlaneLayer()
         glLayer.textDelegate = self
         layer = glLayer
+        registerForDraggedTypes([kUTTypeFileURL as NSPasteboard.PasteboardType])
     }
 
     let x0: CGFloat = 2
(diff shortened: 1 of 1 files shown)


Sources/XiEditor/EditView.swift
```diff
@@ -228,6 +229,21 @@ final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
         return true
     }
 
+    override func draggingEnded(_ sender: NSDraggingInfo) {
+        guard let pasteboard = sender.draggingPasteboard().propertyList(forType: NSPasteboard.PasteboardType(rawValue: "NSFilenamesPboardType")) as? NSArray,
+            let path = pasteboard[0] as? String
```
mmatoszko: going for `pasteboard[0]` assumes that the array is not empty. It would be safer to go for `pasteboard.first` 😄

Sources/XiEditor/EditView.swift
```diff
@@ -228,6 +229,21 @@ final class EditView: NSView, NSTextInputClient, TextPlaneDelegate {
...
+        else { return }
+
+        NSDocumentController.shared.openDocument(
+            withContentsOf: NSURL(fileURLWithPath: path) as URL,
```
mmatoszko: Do we need to go through `NSURL` here? Can't we use `URL.init`?

scholtzan:
Thanks for the review! I did the suggested changes.

PR #422, opened 2019-01-27
Uses compactMap instead of map.filter.map

FrankKair:
Instead of using:

```swift
let cmds = cmdsJson.map { Command(jsonObject: $0) }
                .filter { $0 != nil }
                .map { $0! }
```

We could use `compactMap` to simplify these operations. It's both more legible and more efficient (one loop against 3 loops).

```swift
let cmds = cmdsJson.compactMap { Command(jsonObject: $0) }
```

More info:
https://developer.apple.com/documentation/swift/sequence/2950916-compactmap

## Review Checklist

- [ ] I have tested the code


FrankKair:
I just realized that Xcode 9.2 is being used, so I should use `flatMap` instead.

jansol:
While I appreciate the added simplicity, IIRC this line came up earlier and the consensus was to leave it be until Swift 5.0 is released, at which point we'll be interested in bumping the required Xcode version.

Issue #434, opened 2019-02-10
Selections sometimes not rendered at the end of the line


scholtzan:
There are some issues with rendering selections. When selecting a paragraphs sometimes selections do not go until the end of some lines although they should be.

Here a screenshot of what is a single large selection with missing pieces in some lines (indicated in red):
<img width="739" alt="screenshot 2019-02-10 at 3 17 01 pm" src="https://user-images.githubusercontent.com/1239705/52541203-3a90b200-2d47-11e9-9cce-21f2571cb1f0.png">

cmyr:
First question is whether these selections are being sent by core?

scholtzan:
The selections sent by core look correct. In the screenshot case, core sent one selection annotation covering the whole range. So I guess it's more of a frontend rendering issue.

ryanbloom:
Looks like it might have to do with calculating the width of non-ASCII characters? I notice all the problematic lines have a left or right quotation mark...

cmyr:
Interesting observation @ryanbloom, doesn't really make any sense to me but sure looks suspicious.

ryanbloom:
Okay, I did a little more research. Quotation marks and apostrophes are 3 bytes in UTF-8, while the rest of the characters in this example are 1 byte. So if the code mixes up characters with bytes somehow, it would be off by 2 character widths for every quote/apostrophe in the line it's processing. That seems consistent with @scholtzan's screenshot.

Most likely we're passing a character offset to a measurement function that expects a byte offset. Like... [here](https://github.com/xi-editor/xi-mac/blob/4082c2752bf7ae1c31dec9a63e52a5c980e4fdef/Sources/XiEditor/XiTextPlane/TextLine.swift#L273) maybe? Apple's documentation isn't really clear on what the parameters are supposed to mean; this is just a guess.

scholtzan:
I found a way to fix it: https://github.com/xi-editor/xi-mac/pull/436 I am not entirely sure why it works that way though.

Issue #433, opened 2019-02-10
Fatal error: Can't form Range with upperBound < lowerBound


scholtzan:
I selected some text in a large document (https://raw.githubusercontent.com/xi-editor/test-data/master/the-golden-boys.txt), scrolled down a little and tried to select some other text (starting at the end of the selection moving the cursor up):

```
Thread 1: Fatal error: Can't form Range with upperBound < lowerBound
```
in `convertRange` in `StyleMap`.

scholtzan:
Sometimes it also already happens during the first selection attempt

cmyr:
So is this in Swift land?

scholtzan:
I don't know if it's related, but I also get this warning a lot: `stylemap can't resolve 8`

scholtzan:
Or maybe the document is weird? I can't reproduce it anywhere else.

cmyr:
no more info in the error? We use ranges to describe the visible lines, I'm not sure what else we use them for.

scholtzan:
<img width="964" alt="screenshot 2019-02-10 at 4 04 45 pm" src="https://user-images.githubusercontent.com/1239705/52541755-9fe7a180-2d4d-11e9-931b-57f304dc4ad2.png">


Stack traces:

```
(lldb) po Thread.callStackSymbols
▿ 31 elements
  - 0 : "0   ???                                 0x00000001073ce329 0x0 + 4416398121"
  - 1 : "1   XiEditor                            0x00000001000cfd60 main + 0"
  - 2 : "2   libswiftCore.dylib                  0x00000001006bb14a $Ss12StaticStringV14withUTF8BufferyxxSRys5UInt8VGXElFyt_Tg507$SSRys5F18VGIgy_ACytIegyr_TRAFIgy_Tf1ncn_n + 58"
  - 3 : "3   libswiftCore.dylib                  0x00000001009418fe $Ss18_fatalErrorMessage__4file4line5flagss5NeverOs12StaticStringV_A2HSus6UInt32VtFySRys5UInt8VGXEfU0_TA + 110"
  - 4 : "4   libswiftCore.dylib                  0x00000001006bb14a $Ss12StaticStringV14withUTF8BufferyxxSRys5UInt8VGXElFyt_Tg507$SSRys5F18VGIgy_ACytIegyr_TRAFIgy_Tf1ncn_n + 58"
[...]

PR #436, opened 2019-02-13
Fix selection rendering and prevent crashes

scholtzan:
This fixes https://github.com/xi-editor/xi-mac/issues/434 and https://github.com/xi-editor/xi-mac/issues/433. I am not entirely sure why...

### DIFF
--- a/Sources/XiEditor/Annotation.swift
+++ b/Sources/XiEditor/Annotation.swift
@@ -106,10 +106,9 @@ struct AnnotationStore {
                 while ix < annotationsOfType.count && annotationsOfType[ix].startLine <= lineIx {
                     if (annotationsOfType[ix].endLine >= lineIx) {
                         let annotation = annotationsOfType[ix]
-                        let start = annotation.startLine == lineIx ? annotation.startColumn : 0
-                        let end = annotation.endLine == lineIx ? annotation.endColumn : line.text.count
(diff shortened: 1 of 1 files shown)


mmatoszko:
Shouldn't the `startColumn` and `endColumn` of the `Annotation` be already translated into UTF-16 while we create the `Annotation` struct? I think this could simplify the math here 😄
Or maybe there is a specific reason to use UTF-8 all over the Mac app?

scholtzan:
I looked a bit more into this. I examined the values on the following line which produced the issues:

“Let’s have it quick, Jack,” Bob called, as he stuck his head in at the

The `line.text.count` for the line is 72. If I move the cursor to the end of the line then the utf16 value of the position is 71, the utf8 one is 77. So `line.text.count` seems to already be in utf16. `0` in utf8 is the same in utf16. 

Similar results for other lines. So I believe that confirms what @cmyr mentioned.

mmatoszko:
Fun fact: this might resurface around [Swift 5](https://forums.swift.org/t/string-s-abi-and-utf-8/17676)

cmyr:
Good note @mmatoszko, we'll have to remove various utf-16 conversion code when we do that migration.

And to satisfy my own curiosity, I can confirm that `“ (u+201c)` and `’ (u+2019)` are each 3 code units in utf-8 but only 1 in utf-16.

PR #439, opened 2019-02-18
Fix xi-mac command line being able to open folders

### LINKED ISSUE #424: Can't open files after accidentally trying to open a directory
What it says on the tin. Trying to open a directory using the command line causes xi-mac to activate but seemingly do nothing. Additionally further attempts to open actual files via the command line or the GUI seem to do nothing at all.

nangtrongvuon:
## Summary

Small fix to make xi-mac CLI not open folders.

Should close #424.

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiCliCore/CommandLineTool.swift
+++ b/Sources/XiCliCore/CommandLineTool.swift
@@ -52,6 +52,16 @@ public final class CommandLineTool {
     func resolvePath(from input: String) throws -> String {
         let fileManager = FileManager.default
         var filePath: URL!
+
+        // Small helper function used to determine if path is not a folder.
+        func pathIsDirectory(_ path: String) -> Bool {
+            var isDirectory = ObjCBool(false)
+            if fileManager.fileExists(atPath: path, isDirectory: &isDirectory) {
+                return isDirectory.boolValue
+            } else {
+                return false
+            }
+        }
         
         if input.hasPrefix("/") {
             filePath = URL(fileURLWithPath: input)
@@ -70,6 +80,10 @@ public final class CommandLineTool {
         }
         
         let pathString = filePath.path
+
+        guard !pathIsDirectory(pathString) else {
+            throw CliError.pathIsDirectory
+        }
         
         if !fileManager.fileExists(atPath: pathString) {
             let createSuccess = fileManager.createFile(atPath: pathString, contents: nil, attributes: nil)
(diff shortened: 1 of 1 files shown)


mmatoszko:
It seems that the reason it crashes on CI is referencing [xi-editor](https://travis-ci.com/xi-editor/xi-mac/builds/101312245) submodule commit which is not available yet. Do we need to update `xi-editor` in this PR?

nangtrongvuon:
Oh, I think it's actually pointing to my quick open branch on xi-editor. I'll amend my commit with the proper master.

Sources/XiCliCore/CommandLineTool.swift
```diff
@@ -70,6 +80,10 @@ public final class CommandLineTool {
         }
         
         let pathString = filePath.path
+
+        guard !pathIsDirectory(pathString) else {
```
mmatoszko: nit: Wouldn't it be more readable if we didn't inverse that logic?
What do you think of something like:
```
if pathIsDirectory(pathString) else {
    throw CliError.pathIsDirectory
}
```
? 😄

nangtrongvuon: Doesn't this mean throwing if the path is not a directory? Shouldn't this be 
`if pathIsDirectory(pathString) {
    throw CliError.pathIsDirectory
}`
?

mmatoszko: You’re right, my code won’t even compile 😅

nangtrongvuon: :) Though I think your comment makes sense, as not inverting `guard` logic seems to make it easier to read. I'll change this up a bit.

Sources/XiCliCore/CommandLineTool.swift
```diff
@@ -52,6 +52,16 @@ public final class CommandLineTool {
...
         let fileManager = FileManager.default
         var filePath: URL!
+
+        // Small helper function used to determine if path is not a folder.
+        func pathIsNotDirectory(_ path: String) -> Bool {
+            var isDirectory = ObjCBool(false)
+            if fileManager.fileExists(atPath: path, isDirectory: &isDirectory) {
+                return !isDirectory.boolValue
+            } else {
+                return false
```
mmatoszko: I think we can now return `true` here :)

nangtrongvuon: You're right, going to fix that :)

mmatoszko:
:rocket:

PR #445, opened 2019-03-10
Fix case sensitivity selection for find and rename ignorecase

scholtzan:
There was a small bug where the `case_sensitive` parameter in `find_status` was not parsed correctly.
Fixing this also made me rename `ignoreCase` to `caseSensitive` since it became quite confusing.

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/Search.swift
+++ b/Sources/XiEditor/Search.swift
@@ -190,7 +190,7 @@ class SuplementaryFindViewController: NSViewController, NSSearchFieldDelegate, N
     let resultCountLabel = Label(title: "")
 
     // assigned in IB
-    let ignoreCaseMenuTag = 101
+    let caseSensitiveMenuTag = 101
     let wrapAroundMenuTag = 102
     let regexMenuTag = 103
     let wholeWordsMenuTag = 104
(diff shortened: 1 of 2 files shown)


Sources/XiEditor/Search.swift
```diff
@@ -230,7 +230,7 @@ class SuplementaryFindViewController: NSViewController, NSSearchFieldDelegate, N
     override func validateMenuItem(_ menuItem: NSMenuItem) -> Bool {
         switch menuItem.tag {
         case ignoreCaseMenuTag:
```
nangtrongvuon: I'd probably rename this too while we're doing this.

Sources/XiEditor/Search.swift
```diff
@@ -275,7 +275,7 @@ class SuplementaryFindViewController: NSViewController, NSSearchFieldDelegate, N
     }
 
     @IBAction func selectIgnoreCaseMenuAction(_ sender: NSMenuItem) {
```
nangtrongvuon: Likewise for this. Perhaps `selectCaseSensitiveMenuAction`?

Sources/XiEditor/Search.swift
```diff
@@ -395,7 +395,7 @@ extension EditViewController {
             }
 
             if firstStatus.caseSensitive != nil {
-                query?.ignoreCase = !(statusQuery.caseSensitive != nil)
+                query?.caseSensitive = (statusQuery.caseSensitive != nil && statusQuery.caseSensitive!)
```
mmatoszko: I don't fully understand why is the `statusQuery.caseSensitive` negated here. Wasn't it just the case of the previous setting?

scholtzan: Yeah, the wording `ignoreCase` and `caseSensitive` was source for confusion before. Case sensitive is basically the opposite of ignoring case (if casing is ignored then it's _not_ sensitive to casing), that's why it had to be negated before.

Issue #446, opened 2019-03-12
Unable to clone repo


favetelinguis:
Trying to run the suggested command:

`git clone --recurse-submodules https://github.com/xi-editor/xi-mac
`

Results in the following error:

```
Submodule path 'test-data': checked out '018b32541e4126daa0cd674d2189e6f965a71ece'
error: Server does not allow request for unadvertised object ce5f7ec43f24a54c16a6e20a8d0e9ea4fd90f5f5
Fetched in submodule path 'xi-editor', but it did not contain ce5f7ec43f24a54c16a6e20a8d0e9ea4fd90f5f5. Direct fetching of that commit failed.
```

jansol:
Huh, did someone do a force push on the xi-editor repo recently? Looks like it's failing here too. To fix that the submodule needs to be updated to point to a valid commit again.

As a workaround you can clone without the `--recurse-submodules` bit and do a separate clone of the xi-editor repo into the corresponding directory. You'll want to wipe your copy and re-clone once the submodule is fixed though...

nangtrongvuon:
@scholtzan `ce5f7ec43f24a54c16a6e20a8d0e9ea4fd90f5f5` points to one of your old commits. Perhaps you didn't add the xi-editor submodule when merging the case sensitive PR?

scholtzan:
Oops. Yeah seems like I didn't add it. Should be fixed in https://github.com/xi-editor/xi-mac/pull/447

PR #447, opened 2019-03-13
Fix submodule problem

scholtzan:
Should fix https://github.com/xi-editor/xi-mac/issues/446


cmyr:
oops sorry for letting this sit!

PR #448, opened 2019-03-21
Enhanced logging

jeremywiebe:
## Summary
This PR introduces a small change to the RPC logging so that both send and received messages are logged. It also prefixes them so it's easy to tell which direction the message was going.

For example:
```
SEND:{"method":"new_view","id":0,"params":{}}
RECV:{"id":0,"result":"view-id-2"}
```

## Related Issues
~none~

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/RPCSending.swift
+++ b/Sources/XiEditor/RPCSending.swift
@@ -19,6 +19,10 @@ import AppKit
 /// These logs can be used for profiling & debugging.
 let XI_RPC_LOG = "XI_CLIENT_RPC_LOG"
 
+let NEW_LINE = [0x0a as UInt8]
+let SEND_BYTES = "SEND:".data(using: .utf8)!
+let RECV_BYTES = "RECV:".data(using: .utf8)!
+
 /// An error returned from core
 struct RemoteError {
     let code: Int
@@ -202,13 +206,9 @@ class StdoutRPCSender: RPCSending {
 
     private func sendJson(_ json: Any) {
         do {
-            let data = try JSONSerialization.data(withJSONObject: json, options: [])
-            let mutdata = NSMutableData()
-            mutdata.append(data)
-            let nl = [0x0a as UInt8]
-            mutdata.append(nl, length: 1)
-            rpcLogWriter?.write(bytes: mutdata as Data)
-            inHandle.write(mutdata as Data)
+            let data = try JSONSerialization.data(withJSONObject: json, options: []) + NEW_LINE
+			self.rpcLogWriter?.write(bytes: SEND_BYTES + data)
+            inHandle.write(data)
         } catch _ {
             print("error serializing to json")
         }
@@ -220,7 +220,9 @@ class StdoutRPCSender: RPCSending {
     }
 
     private func handleRaw(_ data: Data) {
-        Trace.shared.trace("handleRaw", .rpc, .begin)
+		self.rpcLogWriter?.write(bytes: RECV_BYTES + data + NEW_LINE)
+
+		Trace.shared.trace("handleRaw", .rpc, .begin)
         do {
             let json = try JSONSerialization.jsonObject(with: data, options: .allowFragments)
             handleRpc(json)
--- a/Sources/XiEditor/RPCSending.swift
+++ b/Sources/XiEditor/RPCSending.swift
@@ -20,8 +20,8 @@ import AppKit
 let XI_RPC_LOG = "XI_CLIENT_RPC_LOG"
 
 let NEW_LINE = [0x0a as UInt8]
-let SEND_LOG_PREFIX = "SEND:".data(using: .utf8)!
-let RECV_LOG_PREFIX = "RECV:".data(using: .utf8)!
+let SEND_LOG_PREFIX = "[CLIENT] ".data(using: .utf8)!
+let RECV_LOG_PREFIX = "[CORE]   ".data(using: .utf8)!
 
 /// An error returned from core
 struct RemoteError {
--- a/Sources/XiEditor/RPCSending.swift
+++ b/Sources/XiEditor/RPCSending.swift
@@ -206,9 +206,15 @@ class StdoutRPCSender: RPCSending {
 
     private func sendJson(_ json: Any) {
         do {
-            let data = try JSONSerialization.data(withJSONObject: json, options: []) + NEW_LINE
-            self.rpcLogWriter?.write(bytes: SEND_LOG_PREFIX + data)
-            inHandle.write(data)
+            var data = try JSONSerialization.data(withJSONObject: json, options: [])
+			data.append(NEW_LINE, count: 1)
+
+			if let writer = self.rpcLogWriter {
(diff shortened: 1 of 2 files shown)


Sources/XiEditor/RPCSending.swift
```diff
@@ -19,6 +19,10 @@ import AppKit
 /// These logs can be used for profiling & debugging.
 let XI_RPC_LOG = "XI_CLIENT_RPC_LOG"
 
+let NEW_LINE = [0x0a as UInt8]
+let SEND_LOG_PREFIX = "SEND:".data(using: .utf8)!
+let RECV_LOG_PREFIX = "RECV:".data(using: .utf8)!
```
nangtrongvuon: Perhaps use `[SEND]` and `[RECV]`?

jeremywiebe: I have been debating formatting.  If I do, then each message now takes more than one line per message, which I think might make it a bit harder to scan and find the message you're after.

For example:

```
SEND: {
  "params":{
    "enabled":true
  }, 
  "method":"tracing_config"
}
SEND:{
  "params":{
    "client_extras_dir":"...",
    "config_dir":"..."
  },
  "method":"client_started"
}
RECV:{
  "method":"available_languages",
  "params":{
    "languages":[
      "ASP","ActionScript","AppleScript","Batch File","BibTeX","Bourne Again Shell (bash)","C","C#","C++","CSS","Clojure","D","Diff","Erlang","Git Attributes","Git Commit","Git Config","Git Ignore","Git Link","Git Log","Git Rebase Todo","Go","Graphviz (DOT)","Groovy","HTML","HTML (ASP)","HTML (Erlang)","HTML (Rails)","HTML (Tcl)","Haskell","JSON","Java","Java Properties","Java Server Page (JSP)","JavaScript","JavaScript (Rails)","LaTeX","Lisp","Literate Haskell","Lua","MATLAB","Makefile","Markdown","NAnt Build File","OCaml","OCamllex","OCamlyacc","Objective-C","Objective-C++","PHP","Pascal","Perl","Plain Text","Python","R","Rd (R Documentation)","Regular Expression","Ruby","Ruby Haml","Ruby on Rails","Rust","SQL","SQL (Rails)","Scala","TOML","Tcl","TeX","Textile","XML","YAML","reStructuredText"
    ]
  }
}
```

I could be convinced to apply formatting if folks think it'd be useful. :) 

I do like you're idea of naming though.. I could easily use `CLIENT` and `CORE` and still retain the "column" layout. I'll update the PR with that change tonight.

```
[CLIENT] {"method":"new_view","id":0,"params":{}}
[CORE]   {"id":0,"result":"view-id-2"}
```

Sources/XiEditor/RPCSending.swift
```diff
@@ -19,6 +19,10 @@ import AppKit
 /// These logs can be used for profiling & debugging.
 let XI_RPC_LOG = "XI_CLIENT_RPC_LOG"
 
+let NEW_LINE = [0x0a as UInt8]
+let SEND_LOG_PREFIX = "[CLIENT] ".data(using: .utf8)!
+let RECV_LOG_PREFIX = "[CORE]   ".data(using: .utf8)!
```
nangtrongvuon: Perhaps change these to `CORE` and `CLIENT` too? Then I think this is good to go.

Sources/XiEditor/RPCSending.swift
```diff
@@ -202,13 +206,9 @@ class StdoutRPCSender: RPCSending {
...
         do {
-            let data = try JSONSerialization.data(withJSONObject: json, options: [])
-            let mutdata = NSMutableData()
-            mutdata.append(data)
-            let nl = [0x0a as UInt8]
-            mutdata.append(nl, length: 1)
-            rpcLogWriter?.write(bytes: mutdata as Data)
-            inHandle.write(mutdata as Data)
+            let data = try JSONSerialization.data(withJSONObject: json, options: []) + NEW_LINE
+            self.rpcLogWriter?.write(bytes: SEND_LOG_PREFIX + data)
```
cmyr: Do we know that `+` on two `Data` doesn't reallocate?

jeremywiebe: Good call @cmyr.  I hadn't thought of that!  I've pushed an update....

XiEditor.xcodeproj/project.pbxproj
```diff
@@ -319,6 +319,7 @@
 				AE499DCD1C82BB2B002D68AF /* Products */,
 			);
 			sourceTree = "<group>";
+			usesTabs = 0;
```
nangtrongvuon: I think it was agreed some time ago that we'd use tabs in xi-mac, so I don't think this is necessary.

Sources/XiEditor/RPCSending.swift
```diff
@@ -202,13 +206,15 @@ class StdoutRPCSender: RPCSending {
...
-            let nl = [0x0a as UInt8]
-            mutdata.append(nl, length: 1)
-            rpcLogWriter?.write(bytes: mutdata as Data)
-            inHandle.write(mutdata as Data)
+            var data = try JSONSerialization.data(withJSONObject: json, options: [])
+			data.append(NEW_LINE, count: 1)
+
+			if let writer = self.rpcLogWriter {
+                writer.write(bytes: SEND_LOG_PREFIX)
+                writer.write(bytes: data)
```
nangtrongvuon: A little bit of reindent, and then this is mergeable to me.

XiEditor.xcodeproj/project.pbxproj
jeremywiebe: Interesting.  It seems like all the code I've encountered uses spaces. The indentation issues below are because my editor keeps wanting to use tabs! :) 

I'll fix up the indentation to match the file and revert this change.

nangtrongvuon: Woops, after further checking I realized the default (and what xi-mac actually use) are spaces. I'm not sure why this showed up as a change, maybe newer XCodes demand that you have it explicit now or something.

jeremywiebe: I've been using Xcode 10.2 Beta 4.  I suspect that it now makes that setting explicit and, when missing, defaults to tabs. We might want to consider turning that on explicitly as other folks start to use Xcode 10.2.

I've removed this line for now and will post this Q into Zulip for further discussion. :)

Sources/XiEditor/RPCSending.swift
jeremywiebe: FYI, I did some profiling and it looks like the `+` does re-alloc. Does this look better?

cmyr: Yep!

cmyr:
Looks like there's still some whitespace stuff to clean up, otherwise looks gtg.

jeremywiebe:
@cmyr Thanks for reviewing. I don't see any more tab issues when looking at the *Files changed* tab. Can you point to an example where you think there's still issues?

PR #458, opened 2019-03-28
Fix CLI waiting for files involving symlinks

didroe:
## Summary

When a file had a symlink in it's path, the path returned from `resolvePath` didn't match the path returned from the observer. This change introduces a common method for canonicalizing file paths in
both places.

Fixes https://github.com/xi-editor/xi-editor/issues/1143

## Questions

This is the first time I've done any Mac/Swift/Xi development. As such, I'm not sure my approach is ideal and I have the following questions about my changes:

* I have simplified the resolving code. I think this still handles the existing cases correctly but am I missing something?
* I wasn't sure whether to limit my changes to the observer code. Canonicalising both the input and observer paths seems fine to me, but perhaps doing this on the inputs could cause issues elsewhere?

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Tests/XiCliCoreTests/XiCliCoreTests.swift
+++ b/Tests/XiCliCoreTests/XiCliCoreTests.swift
@@ -23,11 +23,14 @@ class XiCliCoreTests: XCTestCase {
         super.setUp()
         let testArguments = Arguments(arguments: ["test.txt", "--wait"])
         commandLineTool = CommandLineTool(args: testArguments)
-        FileManager.default.createFile(atPath: "test.txt", contents: "This is a tester file".data(using: .utf8), attributes: nil)
+        let fileManager = FileManager.default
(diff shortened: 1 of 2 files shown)


Tests/XiCliCoreTests/XiCliCoreTests.swift
```diff
@@ -90,4 +98,15 @@ class XiCliCoreTests: XCTestCase {
         expectation.fulfill()
     }
 
+    func fileInCurrentDir(_ filename: String) -> String {
+        var url = URL(string: FileManager.default.currentDirectoryPath)!
```
cmyr: There is a convenience initializer that might be helpful here, [`Url.init(fileURLWithPath: String, relativeTo: URL?)`](https://developer.apple.com/documentation/foundation/url/3126803-ini)

liamrosenfeld:
LGTM! 👍

PR #463, opened 2019-05-02
Upgrade to Swift 4.2

jeremywiebe:
## Summary
This PR moves the codebase to Swift 4.2. It turns out to be a pretty small.

## Related Issues
--None--

Related to #461 

## Checklist
Example:

- [x] Entire codebase compiles without error (including unit tests)
- [x] All unit tests pass
- [x] Xi doesn't crash in basic smoke-test

## Review Checklist

- [ ] I have responded to reviews and made changes where appropriate.
- [x] I have tested the code
- [x] I have updated comments / documentation related to the changes I made.
- [x] I have rebased my PR branch onto xi-mac/master.

PR #464, opened 2019-05-04
Upgrade to Swift 5!

### LINKED ISSUE #461: update to swift 5
Swift 5 is now out, and we should look into migrating.

This should probably be painless, but there are a few places where we should be careful; in particular the places where we currently convert utf8 offsets to utf16 offsets; I haven't looked too closely at the new String api in swift, but there should be some way to use utf8 offsets directly.

jeremywiebe:
## Summary
This PR upgrades the codebase to Swift 5. It will require compiling with Xcode 10.2.

The most significant change is the raw byte handling in `RPCSending.swift` so please review that closely... I think I've gotten it right with hopefully similar performance!

## Related Issues
Related to #463

Example: Closes #461

## Checklist
- [x] All code compiles without warnings
- [x] All unit tests pass

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/Search.swift
+++ b/Sources/XiEditor/Search.swift
@@ -85,8 +85,8 @@ class FindViewController: NSViewController, NSSearchFieldDelegate {
             newSearchFieldController.parentFindView = self
 
             if searchField != nil {
-                searchQueries.insert(newSearchFieldController, at: searchQueries.index(of: searchField!)! + 1)
-                searchFieldsStackView.insertView(newSearchFieldController.view, at: searchQueries.index(of: searchField!)! + 1, in: .center)
+                searchQueries.insert(newSearchFieldController, at: searchQueries.firstIndex(of: searchField!)! + 1)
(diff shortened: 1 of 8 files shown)


Sources/XiEditor/Search.swift
```diff
@@ -543,6 +543,9 @@ extension EditViewController {
 
         case .hideReplaceInterface:
             Swift.print("hideReplaceInterface not implemented")
+
+        @unknown default:
```
cmyr: what's this syntax? Can you link me a reference?

jeremywiebe: I just learned about it while upgrading also. It allows us to have an exhaustive enum with a `default` case to protect against future additions to the enum.  If we don't put the `@unknown` on the `default`, the compiler will complain that we've covered all the cases already.

cmyr: cool, thanks!

PR #466, opened 2019-05-22
Fix Emoji drawing for dark themes

### LINKED ISSUE #266: Emoji colors are inverted for dark themes
With a dark-on-light theme:

Emoji with light theme

With a light-on-dark theme:

Emoji with dark theme

There are also slight variations depending on the exact shade of the theme's text color.

jeremywiebe:
**NOTE: This is incomplete. I've gotten a bit lost in the vertex shader as that's completely new territory for me. Any pointers would be appreciated.**

## Summary
This PR fixes drawing emoji when using dark themes. It detects an Emoji in a CTRun and provides a flag to the text vertex shader program so it can avoid inverting the emoji colours.

Closes #266 

## Checklist
Example:

- [ ] make xi perfect

## Review Checklist

- [ ] I have tested the code

### DIFF
--- a/Sources/XiEditor/XiTextPlane/Renderer.swift
+++ b/Sources/XiEditor/XiTextPlane/Renderer.swift
@@ -42,7 +42,7 @@ class Renderer {
     var textProgram: ShaderProgram
     var text_u_scale: GLuint = 0
     let maxTextInstances = 65536
-    let textInstanceSize = 12 // in floats
+    let textInstanceSize = 13 // in floats
     var textInstances: [GLfloat]
     var textInstanceIx = 0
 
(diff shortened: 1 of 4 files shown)


raphlinus:
Unfortunately I think you might require a different pipeline to draw emoji than to draw RGB subpixeled text. The existing shaders blend the foreground color using the RGB pixel values as 3 independent alpha masks (this requires dual source blending, which is set in the Swift code that does the draw call).

So to do it really correctly, you'll need another draw for the color emojis. These won't take an fg color (although it's possible they could take an alpha value), don't do dual source blending. The fragment shader should be really simple, basically just return `texture(mask, uv)` (note, you'll need all 4 channels, the emoji has an alpha channel).

Sorry this isn't simpler - it would be nice to use the existing shaders, just make them do emoji properly.

nangtrongvuon:
> Unfortunately I think you might require a different pipeline to draw emoji than to draw RGB subpixeled text. The existing shaders blend the foreground color using the RGB pixel values as 3 independent alpha masks (this requires dual source blending, which is set in the Swift code that does the draw call).

Just wondering, why is it bad to draw emoji as RGB subpixeled text? Is it because then they don't appear properly?

raphlinus:
@nangtrongvuon Very simply, there are different rules for drawing it. Text is only an alpha value (though there are three separate channels for RGB) and takes the foreground color. Emoji is basically an RGBA image. I'm not sure if there's good documentation easily accessible, but if you look at the text drawing part of, say, Android, you'll find different paths for drawing them.

jeremywiebe:
@raphlinus thank-you for the direction! I was kinda lost once I hit the shaders. I'll begin getting more familiar with this part of xi-mac (and OpenGL) and update this PR!

PR #467, opened 2019-06-09
Make app appears in the "Open as..." menu

balthild:
image

### DIFF
--- a/Sources/XiEditor/Info.plist
+++ b/Sources/XiEditor/Info.plist
@@ -16,6 +16,20 @@
 	<array>
 		<dict>
 			<key>CFBundleTypeExtensions</key>
+			<array/>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>LSItemContentTypes</key>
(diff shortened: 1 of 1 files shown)


Sources/XiEditor/Info.plist
```diff
@@ -53,7 +67,5 @@
 	<string>Main</string>
 	<key>NSPrincipalClass</key>
 	<string>NSApplication</string>
-	<key>NSSupportsAutomaticGraphicsSwitching</key>
```
nangtrongvuon: Not sure if you want to remove this, AFAIK this makes xi-mac use the discrete GPU whenever possible and that eats a bunch of unnecessary battery.

balthild: Xcode removed it as duplication. There's another `NSSupportsAutomaticGraphicsSwitching` in the file.

nangtrongvuon: Gotcha. Seems like an all around good change then.

PR #469, opened 2019-06-18
Fix janky scrolling on external mice

### LINKED ISSUE #411: Scrolling with an external mouse is very janky
When scrolling with the wheel of an external mouse, instead of smooth scrolling like when swiping on the touchpad, I get a couple of "ticks", then the view stops for a second, and then jumps to the final location.

It's a cheap wireless mouse and I use Scroll Reverser with it, but other applications don't have this problem so I doubt it's related to that.

nangtrongvuon:
## Summary
This fixes an issue where the `scroll(to: NSPoint)` usage in xi-mac's case _only_ supports smooth scrolling
and thus doesn't properly take into account notched mouse wheel scroll (which usually means external mice).

Smooth scrolling (like the MacBook trackpad or Apple Magic Mouse) sends scrollEvents that are chunked, continuous and cumulative, and thus the scroll view's `clipView`'s bounds is set properly for each of these small chunks of scrolling. 

Scrolling with notched mice scrolls in discrete units, takes into account acceleration but does not redraw the view when the view is continuously redrawn during scrolling - this is because the bounds origin is only set after the scrolling has stopped completely, causing the janky scroll that is described in #411. This is likely a bug/oversight on Apple's part.

This PR simply sets the `clipView`'s bound origin properly on all scrolling.

closes #411.

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/XiClipView.swift
+++ b/Sources/XiEditor/XiClipView.swift
@@ -21,8 +21,13 @@ protocol ScrollInterested: class {
 class XiClipView: NSClipView {
     weak var delegate: ScrollInterested?
 
+    // Smooth scrolling (like the MacBook trackpad or Apple Magic Mouse) sends scroll events that are chunked, continuous and cumulative, 
+    // and thus the scroll view's clipView's bounds is set properly (in small increments) for each of these small chunks of scrolling. 
+    // Scrolling with notched mice scrolls in discrete units, takes into account acceleration but does not redraw the view when the view is continuously redrawn (like in xi-mac) during scrolling. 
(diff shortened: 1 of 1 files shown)


cmyr:
Very cool, will review this later today!

cmyr:
So CI is failing because of an issue with a search test? I'm not sure what's up with that, but it might be due to some change in `xi-editor`. Maybe change this PR to *only* include the scroll change, and not update the submodule, and then we can look at the other problem separately?

cmyr:
In any case if you do that change feel free to merge on green!

PR #471, opened 2019-07-02
Copies default config file when none found

### LINKED ISSUE #375: Populate new config file with defaults
Currently when no config file is present and the user hits `cmd+,` xi-mac appears to do nothing and silently creates an empty config file. Instead, it should create a config file with the current config and open that.

nangtrongvuon:
Previously xi-mac only copied the base config file on startup - this commit makes it also copies the base config when the user opens the preference folder and no config is found.

closes #375

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/AppDelegate.swift
+++ b/Sources/XiEditor/AppDelegate.swift
@@ -98,29 +98,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     }
 
     lazy var defaultConfigDirectory: URL = {
-        let applicationDirectory = FileManager.default.urls(
+        return FileManager.default.urls(
             for: .applicationSupportDirectory,
             in: .userDomainMask)
             .first!
             .appendingPathComponent("XiEditor")
-
-        // create application support directory and copy preferences file on first run
-        if !FileManager.default.fileExists(atPath: applicationDirectory.path) {
-            do {
-
-                try FileManager.default.createDirectory(at: applicationDirectory,
-                                                        withIntermediateDirectories: true,
-                                                        attributes: nil)
-                let preferencesPath = applicationDirectory.appendingPathComponent(PREFERENCES_FILE_NAME)
-                let defaultConfigPath = Bundle.main.url(forResource: "client_example", withExtension: "toml")
-                try FileManager.default.copyItem(at: defaultConfigPath!, to: preferencesPath)
-
-
-            } catch let err  {
-                fatalError("Failed to create application support directory \(applicationDirectory.path). \(err)")
-            }
-        }
-        return applicationDirectory
     }()
 
     lazy var errorLogDirectory: URL? = {
@@ -168,7 +150,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         let xiCore = CoreConnection(rpcSender: rpcSender)
         self.xiCore = xiCore
         updateRpcTracingConfig(collectSamplesOnBoot)
-
+        setupConfigDirectory()
         xiCore.clientStarted(configDir: getUserConfigDirectory(), clientExtrasDir: bundledPluginPath)
 
         // fallback values used by NSUserDefaults
@@ -274,6 +256,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     //MARK: - top-level interactions
     @IBAction func openPreferences(_ sender: NSMenuItem) {
         let preferencesPath = defaultConfigDirectory.appendingPathComponent(PREFERENCES_FILE_NAME)
+        setupConfigDirectory()
         NSDocumentController.shared.openDocument(
             withContentsOf: preferencesPath,
             display: true,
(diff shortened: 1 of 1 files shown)


cmyr:
hmm, this is confusing to me. What's the scenario where we open the preference folder and no config is found? Why wasn't it copied on startup? Is this like if the user manually deletes it?

nangtrongvuon:
@cmyr Yeah, pretty much. We only create `preferences.xiconfig` on first run at the moment, when the Application Support folder for Xi hasn't existed yet. If the user deletes that config file (or it disappears for some reason), it causes what is described in #375.

Sources/XiEditor/AppDelegate.swift
```diff
@@ -288,14 +288,26 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+        
         NSDocumentController.shared.openDocument(
             withContentsOf: preferencesPath,
             display: true,
             completionHandler: { (document, alreadyOpen, error) in
                 if let error = error {
                     print("error opening preferences \(error)")
                 }
-        })
+        })
```
cmyr: whitespace!

nangtrongvuon: oops, will get to that.

cmyr: ping?

nangtrongvuon:
@cmyr Sorry, I've been a bit occupied with some stuff along with not feeling too well in general, gonna get to this sometime this week!

nangtrongvuon:
@cmyr Got to updating this PR today with your suggestions, does it look okay now?

Sources/XiEditor/AppDelegate.swift
```diff
@@ -285,6 +264,38 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+                                                        attributes: nil)
+                let preferencesPath = applicationDirectory.appendingPathComponent(PREFERENCES_FILE_NAME)
+                let defaultConfigPath = Bundle.main.url(forResource: "client_example", withExtension: "toml")
+                try FileManager.default.copyItem(at: defaultConfigPath!, to: preferencesPath)
+            } catch let err  {
+                fatalError("Failed to create application support directory \(applicationDirectory.path). \(err)")
+            }
+        } else if !FileManager.default.fileExists(atPath: preferencesDirectory.path) {
+            do {
+                let defaultConfigPath = Bundle.main.url(forResource: "client_example", withExtension: "toml")
```
cmyr: This block duplicates code above. Instead of having this be two separate cases, can't we just have to separate `if` statements, first checking for the directory (creating if needed) and then checking for the file, also creating if needed?

Sources/XiEditor/AppDelegate.swift
```diff
@@ -285,6 +264,38 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+        if !FileManager.default.fileExists(atPath: applicationDirectory.path) {
+            do {
+                try FileManager.default.createDirectory(at: applicationDirectory,
+                                                        withIntermediateDirectories: true,
+                                                        attributes: nil)
+                let preferencesPath = applicationDirectory.appendingPathComponent(PREFERENCES_FILE_NAME)
+                let defaultConfigPath = Bundle.main.url(forResource: "client_example", withExtension: "toml")
+                try FileManager.default.copyItem(at: defaultConfigPath!, to: preferencesPath)
+            } catch let err  {
+                fatalError("Failed to create application support directory \(applicationDirectory.path). \(err)")
```
cmyr: I'm not sure why this error should be fatal and the other one shouldn't. I don't know if there are some circumstances (sandboxing?) where we ever expect to fail here, but if we do fail showing a message to the user is better than hard crashing.

Sources/XiEditor/AppDelegate.swift
```diff
@@ -97,31 +97,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
-                try FileManager.default.copyItem(at: defaultConfigPath!, to: preferencesPath)
-
-
-            } catch let err  {
-                fatalError("Failed to create application support directory \(applicationDirectory.path). \(err)")
-            }
-        }
-        return applicationDirectory
-    }()
+    var defaultConfigDirectory: URL!
```
cmyr: This smells bad. This directory never changes, and we _do_ know it at compile time, even if it doesn't exist; so maybe this should just be,

```swift
lazy var defaultConfigDirectory: URL = {
    return FileManager.default.urls(		
        for: .applicationSupportDirectory,		
         in: .userDomainMask)		
         .first!		
         .appendingPathComponent("XiEditor")
}()
```
?

This means we can also use `self.defaultConfigDirectory` in `setupConfigDirectory`.


Edit:

to clarify, one of the reasons I don't like this right now is that it's hard to reason about; we have no idea when or where this gets set, and it turns out that it gets set as a side-effect of another function.

nangtrongvuon:
@cmyr Thanks for the review! Was a tiny little bit hasty and tired when I was writing this a few weeks earlier, but I fixed it up with your feedback. Does it look good now?

Sources/XiEditor/AppDelegate.swift
```diff
@@ -98,29 +98,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     }
 
     lazy var defaultConfigDirectory: URL = {
-        let applicationDirectory = FileManager.default.urls(
-            for: .applicationSupportDirectory,
-            in: .userDomainMask)
-            .first!
+        return FileManager.default.urls(
```
cmyr: can we clean up the trailing whitespace? there's a good setting in Xcode for this:
Screen Shot 2019-08-06 at 11 40 47 AM

Sources/XiEditor/AppDelegate.swift
```diff
@@ -169,6 +151,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         self.xiCore = xiCore
         updateRpcTracingConfig(collectSamplesOnBoot)
 
+        // Setups config folder and file
```
cmyr: Total 100% nit: I think this function is well enough named that the comment is redundant.

PR #481, opened 2019-10-19
#470 Change default find highlight color

### LINKED ISSUE #470: Change default find highlight color
Currently find highlights are red by default. We should default to `NSColor.selectedTextColor` or something? There may also be a color defined in the theme that we can use. (The motivation for not using the colors in the theme is that we support multiple 'find groups' with different colors, and want these to be distinct, but I think we should default to an appropriate color for the first group.

mohanio:
## Summary

Find highlight color defaults to red color instead it should use system's default find highlight color(NSColor.findHighlightColor)

Fix:
* Theme has new static variable representing default find highlight color
* Other find highlight colors are generated from the hue of default find highlight color
* Find highlight color specified by xi-editor on theme_changed event is ignored

## Related Issues
`closes #470 `

## Checklist
Example:

- [ ] Example check-list item
- [ ] make xi perfect

## Review Checklist

- [ ] I have tested the code


Sources/XiEditor/Theme.swift
```diff
@@ -82,7 +93,8 @@ extension Theme {
         let caret = NSColor(jsonRgbaColor: json["caret"] as? [String: Any] ?? [:])
         let line_highlight = NSColor(jsonRgbaColor: json["line_highlight"] as? [String: Any] ?? [:])
 
-        let find_highlight: NSColor? = NSColor(jsonRgbaColor: json["find_highlight"] as? [String: Any] ?? [:])
+        // Use default find highlight color instead of theme's find_highlight value
+        let find_highlight: NSColor? = Theme.defaultFindHighlightColor
```
nangtrongvuon: I think if the theme specifies a color for find highlights, then it's ok to use that color instead of forcing the default yellow.

mohanio: Sure, I'm also bit not clear on expected behavior. I will wait for clarification https://github.com/xi-editor/xi-mac/issues/470#issuecomment-544213362

mohanio:
> Thanks for this! Overall looks good. but you shouldn't have the issue link as the commit message. Instead, describe what you did, and then put the link in the commit message body instead.

I added both issue link and description about the fix in commit message.
image

Are you referring to some other place sorry bit new to github workflow.

mohanio:
> Thanks for this! Overall looks good. but you shouldn't have the issue link as the commit message. Instead, describe what you did, and then put the link in the commit message body instead.

Updated the commit subject and message.

PR #483, opened 2019-10-20
460 xi --wait does not return after application exit

### LINKED ISSUE #460: xi --wait does not return after application exit
- macOS 10.14.4
- the commit you're on #b55f1c0
- I run xi-mac with the following command: `xi --wait <file>`

When quitting xi-mac with the cmd+q keyboard shortcut, the cli does not terminate and keeps waiting.

mohanio:
Summary
Issue:
https://github.com/xi-editor/xi-mac/issues/460  xi --wait does not return after application exit

CLI xi --wait does not exit when the XiEditor app is killed using CMD+Q

XiEditor posts "io.xi-editor.XiEditor.FileClosed" notification on windowShouldClose(_ sender:) which is invoked whenever window is closed. But this delegate is not
invoked when the app is killed using CMD+Q

Fix:
NSDocument invokes close() on both app termination and window close so, instead of windowShouldClose(_ sender:) rely on NSDocument's close() method to post "io.xi-editor.XiEditor.FileClosed" notification

Now Xi CLI should stop waiting both on closing a window and on app exit(CMD+Q)

## Related Issues
closes #460 

## Checklist
Example:

- [ ] Example check-list item
- [ ] make xi perfect

## Review Checklist

- [ ] I have tested the code

### DIFF
--- a/Sources/XiEditor/Document.swift
+++ b/Sources/XiEditor/Document.swift
@@ -96,6 +96,7 @@ class Document: NSDocument {
             xiCore.closeView(identifier: identifier)
         }
         super.close()
+        editViewController?.postDocumentClosedNotification()
     }
 
     override var isEntireFileLoaded: Bool {
--- a/Sources/XiEditor/EditViewController.swift
+++ b/Sources/XiEditor/EditViewController.swift
@@ -966,6 +966,12 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
     @IBAction func duplicateLine(_ sender: NSMenuItem) {
         document.sendRpcAsync("duplicate_line", params: [])
     }
+    
+    func postDocumentClosedNotification() {
+        let path = self.document.fileURL?.path // To check if window contains file opened by cli
+        let notification = Notification.Name("io.xi-editor.XiEditor.FileClosed")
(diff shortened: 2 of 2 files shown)


Sources/XiEditor/Document.swift
```diff
@@ -96,6 +96,7 @@ class Document: NSDocument {
             xiCore.closeView(identifier: identifier)
         }
         super.close()
+        editViewController?.documentClosed()
```
nangtrongvuon: This looks alright to me but this can just simply call `postDocumentClosedNotification` right?

mohanio: Yes. Thought of giving more generic callback to EditViewController so, that it can be used for other purpose if required. But right now documentClosed() method doesn't do anything but postDocumentClosedNotification. What do you think, if it doesn't sound right I can directly invoke postDocumentClosedNotification.

nangtrongvuon: I think if that's necessary then we'll change it, but right now it shouldn't be two different method calls. Just `postDocumentClosedNotification` is good enough.

mohanio: sure

mohanio: @nangtrongvuon Could you please merge this if you are ok with the change.

PR #486, opened 2019-12-17
Tail feature

sjoshid:
Closes https://github.com/xi-editor/xi-editor/issues/922.
This PR adds a "Tail File" option under Debug menu. This option can be enabled/disabled per file. 

TODO: Gray out this option if the file doesnt exist. No point in tailing a file that doesnt exist. 

## Review Checklist

- [x] I have tested the code

### DIFF
--- a/Sources/XiEditor/Client.swift
+++ b/Sources/XiEditor/Client.swift
@@ -86,4 +86,7 @@ protocol XiClient: AnyObject {
 
     /// A notification containing the current replace status.
     func replaceStatus(viewIdentifier: String, status: ReplaceStatus)
+    
+    /// A notification telling toggle tail config was successfully changed.
+    func toggleTailConfigChanged(viewIdentifier: String, isTailEnabled: Bool)
 }
--- a/Sources/XiEditor/Core/CoreNotification.swift
+++ b/Sources/XiEditor/Core/CoreNotification.swift
@@ -71,6 +71,7 @@ enum CoreNotification {
 
     case findStatus(viewIdentifier: ViewIdentifier, status: [FindStatus])
     case replaceStatus(viewIdentifier: ViewIdentifier, status: ReplaceStatus)
+    case toggleTailChanged(viewIdentifier: ViewIdentifier, isTailEnabled: Bool)
 
     static func fromJson(_ json: [String: Any]) -> CoreNotification? {
         guard
@@ -227,6 +228,11 @@ enum CoreNotification {
                 let replaceStatus = ReplaceStatus(fromJson: status)
             {
                 return .replaceStatus(viewIdentifier: viewIdentifier!, status: replaceStatus)
+            }
+        case "toggle_tail_config_changed":
+            if let isTailEnabled = jsonParams["is_tail_enabled"] as? Bool
+            {
+                return .toggleTailChanged(viewIdentifier: viewIdentifier!, isTailEnabled: isTailEnabled)
             }
 
         default:
--- a/Sources/XiEditor/EditViewController.swift
+++ b/Sources/XiEditor/EditViewController.swift
@@ -152,6 +152,12 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
             updateLanguageMenu()
         }
     }
+    
+    var isTailEnabled: Bool = false {
+        didSet {
+            updateTailMenu()
+        }
+    }
 
     // used to calculate the gutter width. Initial -1 so that a new document
     // still triggers update of gutter width.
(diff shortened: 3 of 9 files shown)


Sources/XiEditor/EditViewController.swift
```diff
@@ -711,6 +717,10 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
         document.xiCore.setTheme(themeName: sender.title)
     }
     
+    @IBAction func debugToggleTail(_ sender: NSMenuItem) {
+        document.xiCore.toggleTailConfig(identifier: document.coreViewIdentifier!, enabled: !self.isTailEnabled)
```
nangtrongvuon: We could probably rewrite this in a way so we don't have to do `!self.isTailEnabled` here, since it feels counter intuitive.

Sources/XiEditor/Client.swift
```diff
@@ -86,4 +86,7 @@ protocol XiClient: AnyObject {
 
     /// A notification containing the current replace status.
     func replaceStatus(viewIdentifier: String, status: ReplaceStatus)
+    
+    /// A notification telling toggle tail config was successfully changed.
+    func toggleTailConfigChanged(viewIdentifier: String, isTailEnabled: Bool)
```
nangtrongvuon: Why is this `toggleTailConfigChanged`? Do you envision the tail config to have more things than whether if the user enables tailing? Even if so, I feel like this should just be something like `enableTailing`.

Sources/XiEditor/EditViewController.swift
sjoshid: Rewrite how?

Sources/XiEditor/Client.swift
sjoshid: Sure. I'll rename.

sjoshid: Ok. I have made this change.

sjoshid:
There is one small thing remaining. Gray out "Tail" option if the file doesnt exist. Im working on that. 
Also, we need to hold off on merging this till core's PR is merged.

sjoshid:
@nangtrongvuon 
Allow me to add more details to the naming. I sort of struggled with it. (and Im bad at naming)
There are two parts
a) When user clicks the "Tail" option, `EditViewController.debugToggleTail` is the one that toggles the state (tail flag) and sends it to core. Hence the word 'toggle' in method. Tail flag belongs to `EditViewController` and so does `debugToggleTail()`. Instance method modifying instance variable. So toggling here is ok I think. 
b) After sending to core, it replies with either `true` ie tail was toggled successfully or `false` otherwise. This reply from core is handled by `EditViewController.enableTailing`. It shouldnt be toggled because it represents the actual state of the flag. Toggling here would be incorrect. Besides, if core replies with `true` and I toggle in `EditViewController.enableTailing`, then my state would be `false`. Dont think this is correct either. 

I still think `EditViewController.enableTailing(**bool** )` is correct. Because it is not just `enableTailing()`. Parameter is important here. Without parameter, it means what you are saying: enable tailing. In my case it takes a bool. So I would read it as "Enable tailing. Yes or no?".

If you want me to separate enable and disable I would have to do something like
```
bool tailChangedInCore = <reply from core>
if tailChangedInCore
    EditViewController.enableTailing()
else
    EditViewController.disableTailing()
```

```
EditViewController.enableTailing() {
    self.isTailEnabled = true
}
```

```
EditViewController.disableTailing() {
    self.isTailEnabled = false
}
```

If this is what you want, I would be happy to make the change. If not, let me know what you think. Sorry about long comment. I hope Im not going off on a tangent.

Sources/XiEditor/Core/CoreNotification.swift
```diff
@@ -228,6 +229,11 @@ enum CoreNotification {
             {
                 return .replaceStatus(viewIdentifier: viewIdentifier!, status: replaceStatus)
             }
+        case "enable_tailing":
+            if let isTailEnabled = jsonParams["is_tail_enabled"] as? Bool
```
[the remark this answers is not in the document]
sjoshid: I havent push the change. I'll do it based on outcome of the naming discussion we are having.

Sources/XiEditor/EditViewController.swift
```diff
@@ -919,6 +951,11 @@ class EditViewController: NSViewController, EditViewDataSource, FindDelegate, Sc
             languagesMenu.addItem(item)
         }
     }
+    
+    public func enableTailing(_ isTailEnabled: Bool) {
+        self.isTailEnabled = isTailEnabled
```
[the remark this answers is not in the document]
sjoshid: See my long comment below.

nangtrongvuon:
My main problem with this:
`document.xiCore.toggleTailConfig(identifier: document.coreViewIdentifier!, enabled: !self.isTailEnabled)`

Is that you begin to mutate state in a method's argument which I think reads really weird. It doesn't produce the wrong results, but it's also something I'm not okay with seeing in a codebase, since it makes following this code hard and also makes it less maintainable. 

I'm also not a fan of using this frontend boolean to determine behavior of core. I believe this is a code smell, and core should be able to maintain which views are currently being tailed. The frontend doesn't really need to control this state, at most it should only communicate that state to the user (for xi-mac, this should be the `isTailEnabled` boolean), so they know that their file is being tailed and not in an editing state.

So in conclusion, I wouldn't split it into `enable` and `disable`, but I would just have a `toggleTailing` that will toggle tailing state for a view. If it worked and the state changed, cool, we should tell this to the user by changing `isTailEnabled`. If not, then all is fine, just keep the current frontend state.

Issue #487, opened 2019-12-20
Wrong top margin with statusbar -- first visible line is not rendered


jansol:
# Xi-Mac Issue Template
- [x] I have searched existing issues and could not find my issue.
- [x] This is an issue related to a WIP plugin

## Details
When using xiff, the top line of the view area does not get drawn at all (including line number in the gutter), likely caused by wrong margins when the statusbar is present.

PR #489, opened 2019-12-20
Fix mouse event not accounting for status bar frame

nangtrongvuon:
Part 2 of the #487 fix, this makes mouse events also compensate for the status bar's frame.


nangtrongvuon:
I took a quick look at those methods, and they don't seem to interact directly with the `editView`'s frame, so I think this should be fine. Thank you for both of your patience!

